Issue #1154, opened 2014-02-24
Improve layout of credits


supertuxkart-sourceforge-migration:
**Author: hikerstk**

The credit layout in the gui is somewhat strict, e.g.:
= name =

text1
text2

Will end up on two screens, instead of one screen. Atm you have to do:
= name =

some-other-text
- text1
- text2

Also a crash can be caused with incorrectly formatted lines, e.g. if you remove 'some-other-text' and only have the lines with -, it will crash with:
#0  0x0000000000767926 in std::vector<irr::core::string<wchar_t, irr::core::irrAllocator<wchar_t> >, std::allocator<irr::core::string<wchar_t, irr::core::irrAllocator<wchar_t> > > >::push_back (this=0xffffffffffffffe8, __x=...) at /usr/include/c++/4.5/bits/stl_vector.h:743
#1  0x000000000092fc4a in CreditsSection::addSubEntry (this=0x1666d760, subEntryString=...)

```
at /home/joerg/local/supertuxkart/supertuxkart/src/states_screens/credits.cpp:71
```
#2  0x000000000092e373 in CreditsScreen::loadedFromFile (this=0x16496b70) at /home/joerg/local/supertuxkart/supertuxkart/src/states_screens/credits.cpp:194
#3  0x00000000007f3bbf in GUIEngine::Screen::loadFromFile (this=0x16496b70) at /home/joerg/local/supertuxkart/supertuxkart/src/guiengine/screen.cpp:133
#4  0x00000000007d543d in GUIEngine::AbstractStateManager::pushScreen (this=0x14499d0, screen=0x16496b70)

```
at /home/joerg/local/supertuxkart/supertuxkart/src/guiengine/abstract_state_manager.cpp:130
```

Migrated-From: https://sourceforge.net/apps/trac/supertuxkart/ticket/1098

Alayan-stk-2:
The credits should support displaying pictures. Most other games do have some animations or pictures there. We could show concept art, gameplay scenes...

PR #1264, opened 2014-03-27
Issue 1154 fix; Display cancel download button.

artur-kink:
First commit fixes error in issue #1154 - https://github.com/supertuxkart/stk-code/issues/1154

Second commit adds a Cancel button during addon download.

### DIFF
--- a/src/states_screens/credits.cpp
+++ b/src/states_screens/credits.cpp
@@ -68,6 +68,12 @@ public:
     // ------------------------------------------------------------------------
     void addSubEntry(stringw& subEntryString)
     {
+        //Add a blank entry if no entry exists for this sub entry.
+        if(m_entries.size() == 0){
+            stringw blank_line(L"");
+            CreditsEntry entry(blank_line);
+            addEntry(entry);
+        }
         m_entries[m_entries.size()-1].m_subentries.push_back(subEntryString);
     }
 };   // CreditdsSection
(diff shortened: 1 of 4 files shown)


hiker:
Hmm - the assert seems indeed to be wrong, what I am not sure of is why the setManageMemory was commented out. We now have a (very very minor) memory leak there. I can't remember why I commented it out - I'll have a look at this.But since the rest of the pull request is actually independent of that - @auriamg or @deveee - can this be applied?

src/states_screens/credits.cpp
```diff
@@ -68,6 +68,12 @@ class CreditsSection
     // ------------------------------------------------------------------------
     void addSubEntry(stringw& subEntryString)
     {
+        //Add a blank entry if no entry exists for this sub entry.
+        if(m_entries.size() == 0){
```
leyyin: Can we have the '{' on a new line? And also a space after the 'if'.
See section about conditionals https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Conditionals.

src/states_screens/dialogs/addons_loading.cpp
```diff
@@ -248,6 +249,11 @@ GUIEngine::EventPropagation AddonsLoading::processEvent(const std::string& event
         voteClicked();
         return GUIEngine::EVENT_BLOCK;
     }
+    else if(event_source == "cancel")
```
leyyin: A space after the 'if' please.

src/states_screens/credits.cpp
```diff
@@ -68,6 +68,12 @@ class CreditsSection
...
+            stringw blank_line(L"");
```
leyyin: Does this need to be initialized with empty wide string?

hiker:
I've checked the asset, and fixed a minor memory leak (2f902f7). So cancel should now work as expected. The rest is up to @auriamg and/or @deveee :)

auriamg:
This is no longer mergeable, got conflicts in addition to the style issues mentionned above. Since escape should work, closing as outdated

PR #1609, opened 2014-10-09
Newachie

shantanusingh10:
Adding an achievement . Name - Sticky gum , earned after successfully trapping 10 karts in bubble gum .


hiker:
While I like the achievement, as long as it doesn't even compile, we won't have time to fix it. Please issue a pull request once this is up-to-date and working,
Thanks for your help!

Issue #1694, opened 2014-11-11
Make parachute duration depends only of time


Alayan-stk:
Currently, this item is infuriating and its effect is absurd.
Why ?
At high speed, you'll lose speed quickly enough with increased friction to have lost speed over the upper bound fraction which directly detach the parachute. You'll remain at a decent speed and won't lose much time.
At low speed however, the air friction increase is unable to make you lose speed but it forbid any acceleration. So even the lower bound of loss isn't reached if you keep trying to accelerate and you'll wait very long seconds that finally the parachute is detached.
A player knowing how that works will slow himself to quickly detach the parachute, but that's at best a workaround.
Note that a player/AI trying to keep its speed using nitro or something else will also lose time due to that mechanism. Using nitro while slowed down isn't the wiser use, but it shouldn't go to the point it slows you !

Reducing time duration would be a logical counterpart.

nado:
Oh. So if i understand correctly, you may lose the parachute faster by slowing down while having it ? Didnt know that and I’ve played stk for a few years… Maybe could we add this in the help ?

qwertychouskie:
I think it would be best if it was an amount of time, and that amount of time depends how far behind you are (distance, not places).  This might make it a bit easier to catch up with someone _very_ far ahead.

Alayan-stk-2:
Should be closed.

Has been fixed last year with #2768

PR #1696, opened 2014-11-12
Parachutes stay more often but shorter if the speed loss is too low

Alayan-stk:
See #1694 for discussion


hiker:
@Arthur-D is looking at this patch for 0.9a.

deveee:
I'm marking this as invalid since noone is interested in it.

PR #1735, opened 2014-11-22
Fix item placement problems in dynamic ribbons, and some minor corrections

marcoll:
There was an item placement problem with dynamic ribbons that showed up when there was more than one row. As far as I know, the only case where this could happen was the GP editor. For example: if there were 9 different GPs and a dynamic ribbon with 2 rows and 4 columns, the leftmost GPs showed screenshots of other GPs.

This pull request fixes that problem, and also adds some related minor corrections.

### DIFF
--- a/src/utils/time.cpp
+++ b/src/utils/time.cpp
@@ -46,6 +46,7 @@ std::string StkTime::toString(const TimeType &tt)
 {
     const struct tm *t = gmtime(&tt);
 
+    //I18N: Format for dates (%d = day, %m = month, %Y = year). See http://www.cplusplus.com/reference/ctime/strftime/ for more info about date formats.
     core::stringw w_date_format = translations->w_gettext(N_("%d/%m/%Y"));
     core::stringc c_date_format(w_date_format.c_str());
 


src/states_screens/tracks_screen.cpp
```diff
@@ -142,7 +142,7 @@ void TracksScreen::beforeAddingWidget()
     if (group_amount > 1)
     {
         //I18N: name of the tab that will show tracks from all groups
-        tabs->addTextChild( _("All"), ALL_TRACK_GROUPS_ID );
+        tabs->addTextChild( _(ALL_TRACK_GROUPS_ID), ALL_TRACK_GROUPS_ID );
```
auriamg: This may seem like nitpicking, but the label used to be "All" with a capital A, whereas ALL_TRACK_GROUPS_ID is "all" in lowercase. Furthermore, ALL_TRACK_GROUPS_ID is meant to be an ID, not a label; because the ID is similar to the label is IMO not a good reason to confuse IDs and labels, and I think we should always clearly separate ID and label. Translating an ID and then using it as label seems all wrong

deveee:
Just want to mention that the fix for dynamic ribbon widget is already applied. Everything other is rather style modifications and I set milestone to 0.8.3 because of this.

src/states_screens/tracks_screen.cpp
marcoll: The "all" track group is kind of special, since there is actually no such group. It's just a way of saying "any group". How would you do it?

auriamg:
The "all " track group being special doesn't change the fact that translating an ID is wrong :)

I would just leave it as it was, label then ID

```
tabs->addTextChild( _("All"), ALL_TRACK_GROUPS_ID );
```

I see no problem with how it was

hiker:
Not sure whom to assign - @auriamg , or @deveee  :) Auria volunteered today to get pull requests assigned, so she wins :)  But anyway, someone please decide ;)

src/states_screens/tracks_screen.cpp
```diff
@@ -152,6 +152,8 @@ void TracksScreen::beforeAddingWidget()
     //I18N: track group name
     FOR_GETTEXT_ONLY( _("standard") )
     //I18N: track group name
+    FOR_GETTEXT_ONLY( _("nextgen") )
+    //I18N: track group name
```
auriamg: Sorry for the delay before reviewing this... This string should not be translatable, it never appears in the final game

PR #1788, opened 2014-12-15
deleting some really old doc

konstin:
I don't think that anyone needs these files anymore. (And for the rare case that someone would need them, then theres still git)


auriamg:
Hi, sorry for the delay, we had more pressing issues. Regarding robots.txt, we should ask @hiker if it's outdated

PR #1850, opened 2014-12-26
Update irr_driver.cpp

NalinG:
Making the time less than zero should make the message appear just once and I think should solve the problem.
I am having some issues building stk (temporary), so I cannot check it and I am really sorry for it.

### DIFF
--- a/src/graphics/irr_driver.cpp
+++ b/src/graphics/irr_driver.cpp
@@ -2226,6 +2226,13 @@ void IrrDriver::update(float dt)
 
 void IrrDriver::requestScreenshot()
 {
+    RaceGUIBase* base = World::getWorld()
+                          ? World::getWorld()->getRaceGUI()
+                          : NULL;
+    if (base)
+    {
+        base->clearAllMessages();
+    }
     m_request_screenshot = true;
 }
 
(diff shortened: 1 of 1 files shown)


NalinG:
I think it should fix the bug now.

math07:
On lines 2081-2082, base is not initialized and a function is called on a NULL pointer, it'll crash.
RaceGUIBase* base;
base->clearAllMessages();

NalinG:
Thanks @math07  
I am really sorry for such silly mistake.

NalinG:
Hi, But is the issue already fixed, because now I can take the screenshots without the text.
If so, we should close the issue.

auriamg:
Hi, I just tested and unfortunately this does not work for me, there is still text in my screenshots. The issue is likely that you clear messages right before taking the screenshot, but the effect is not instantaneous, clearing messages will take effect upon next render only.

Fortunately the situation is quite simple, screenshots are already asynchronous, so you could clear message sin IrrDriver::requestScreenshot() instead

NalinG:
What if I just add clearmessage function in requestScreenshot method, will that do the trick.?
Because I cannot test it as it is giving just fine screenshots in my system (Ubuntu 14.04.1) even without any tweak.

auriamg:
Yes, clearing messages in requestScreenshot should work

NalinG:
Yup, I just did the same and sent the commit.
Thanks for all the help @auriamg .

auriamg:
Hi, thanks a lot, I have merged this pull request (I did it manually taking only the changes from last commit, to avoid including unecessary commits to the stk main lines)

NalinG:
Thank You for accepting my PR and helping me learn. :)

PR #1862, opened 2014-12-30
Make the lights adjustable in a race via the debug menu

konstin:
This makes it possible to adjust energy, radius and color of the lights while being in a race. For simplicity, only the light nearest to the player kart is edited. The current solution isn't the most elegant, but it should be possible to work with it.

@samuncle what should be the ranges for the values for energy, radius and the colors of a light? And how many steps should be between minimum and maximum? Currently, the ranges are just my try-and-error assumptions.

todos:
 - Better handling of the SpinnerWidget's
 - Make the values changable via mouse (clicking somewhere in the spinner) oder via keyboard
 - Ability to select a different light? Maybe with a GUI where the selected light changes the color.


src/utils/debug.cpp
```diff
@@ -144,8 +148,59 @@ void addAttachment(Attachment::AttachmentType type)
 
 }   // addAttachment
 
+// ----------------------------------------------------------------------------
+/** Computes the distance between two points represented by one vector3df each */
+float distance(core::vector3df one, core::vector3df two)
+{
+    return sqrt(pow(one.X - two.X, 2) + pow(one.Y - two.Y, 2) + pow(one.Z- two.Z, 2));
+}
+
```
auriamg: core::vector3df already contains a method named "getDistanceFrom" that does this, so this could could be slightly simplified by using this method

src/utils/debug.cpp
```diff
@@ -215,15 +270,16 @@ bool onEvent(const SEvent &event)
...
             mnu->addItem(L"Adjust values", DEBUG_VISUAL_VALUES);
 
-            mnu->addItem(L"Profiler",DEBUG_PROFILER);
+            mnu->addItem(L"Profiler", DEBUG_PROFILER);
             if (UserConfigParams::m_profiler_enabled)
                 mnu->addItem(L"Toggle capture profiler report",
                              DEBUG_PROFILER_GENERATE_REPORT);
             mnu->addItem(L"Do not limit FPS", DEBUG_THROTTLE_FPS);
-            mnu->addItem(L"FPS",DEBUG_FPS);
+            mnu->addItem(L"Show FPS", DEBUG_FPS);
```
auriamg: Not fully accurate since it also hides, if you want to change the label then "Toggle" would be better

Issue #1870, opened 2015-01-02
"You need more points" text overflow in story mode


siadyhr:
If you hit a challenge bubble in story mode you haven't unlocked yet, a text stating "YOU NEED MORE POINTS TO ENTER THIS CHALLENGE! CHECK THE MINIMAP FOR AVAILABLE CHALLENGES." If you hit it again, a new text appears below and so on, at last overflowing the screen. They do disappear after a while, though.
Maybe change so only one is displayed at a time?
![screenshot from 2015-01-02 17 36 00](https://cloud.githubusercontent.com/assets/8537104/5597707/5772e7ce-92a7-11e4-9b3e-47749d9d85a2.png)

auriamg:
The text should not appear more than once, and should perhaps use a smaller font

math07:
I'm gonna fix this.

PR #1878, opened 2015-01-04
Fix issue #1870

math07:
Hey,
This pull request fix issue #1870 by adding a method to check if a message related to a specific kart is already present in the message queue and calls it before putting the points message.
I'm sorry for the weird commit log, I had some conflicts rebasing this at first but now github says these branches can be automatically merged so it seems fine.

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -1899,22 +1899,21 @@ void Kart::crashed(const Material *m, const Vec3 &normal)
                                       "to enter this challenge!\n"
                                       "Check the minimap for\n"
                                       "available challenges.");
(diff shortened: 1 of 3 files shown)


deveee:
I just noticed that you removed center align (which is a bit ugly hack because every line is separated message). I cleared all messages instead of this. There is no place for other messages at the same time anyway.

math07:
Yeah, I didn't know you used to put lines  on separate messages as a hack to get center alignment. I thought it was an old piece of code left there maybe when multiline messages weren't supported so that's why I changed it. 

I thought other messages could have been important but if they are not, clearing everything like you did is easier.

Issue #1859, opened 2014-12-30
Layout problems in some screens


marcoll:
Some screens have a poor layout that breaks when using long translated strings. Specially the GP info and the track info screens:

![before1](https://cloud.githubusercontent.com/assets/7282822/5579097/853e0770-9035-11e4-983d-b12320763da5.png)
![before2](https://cloud.githubusercontent.com/assets/7282822/5579099/8cf8e890-9035-11e4-99e8-7ea0d3987390.png)

By doing some minor modifications to the corresponding .stkgui file I have fixed the problem and now they look like this:

![after1](https://cloud.githubusercontent.com/assets/7282822/5579109/b4ee1df2-9035-11e4-9014-12f510a00640.png)
![after2](https://cloud.githubusercontent.com/assets/7282822/5579117/c626256a-9035-11e4-83c8-5a794ca969c5.png)

If anyone thinks this is worth including in the upcoming release, I can send a pull request.

auriamg:
Fixed by applying tweaked version of pull request

PR #1881, opened 2015-01-04
Fix layout problems in some screens

marcoll:
This pull request includes a commit that fixes issue #1859.

It also includes another one that changes the text of a button in one of the GP editor's dialogs. "OK" instead of "Create", since the same dialog is also being used to rename GPs.


auriamg:
Thanks for the request, I've merged your changed with a few little tweaks making the XML cleaner (for instance no need to specify a height on an horizontal spacer, no need to specify a height for labels, etc.) I also right-aligned the text to make it closer to the associated control, this seems to look slightly better

so thanks!

PR #1882, opened 2015-01-06
Make note spinner values of addons search bar readable in low resolution

math07:
As the title says, here's how it currently looks like on low resolutions like 800x600. (1024x768 is also a problem). I changed the proportion and now it looks great on low resolutions and still looks great on high resolutions.

cantsee


auriamg:
Thanks, merged

PR #1896, opened 2015-01-09
Update checkboxes and textfield in the tabbed user screen too

Flakebi:
This fixes the issue that textfields and checkboxes were only updated if you clicked on the user. When the screen was opened in the tabbed version they weren't initialized because the focus was swapped in `selectUser`.

I changed it so the focus is only set in the non-tabbed-screen, but the rest gets initialized in the tabbed version too.


hiker:
I am currently working on the user screen, and will merge this once I am done. Thanks!

PR #1907, opened 2015-01-18
Miscellaneous fixes

Flakebi:
* leyyin found out that some stats were sent in `reportHardwareStats()` and in `getGLLimits(&json)`. Completing the getGLLimits method will introduce the same problem with e. g. GL_SHADING_LANGUAGE_VERSION
* I fixed some compiler warnings (and I have no idea why git changed the assert statement, I can't see the difference)
* The addon-server which is written into the addons_installed.xml now gets extracted from the UserConfig
* I removed a method which does nothing (vlj said it should be safe to remove)


src/addons/addons_manager.cpp
```diff
@@ -583,13 +583,19 @@ bool AddonsManager::uninstall(const Addon &addon)
...
-    //write the header of the xml file
+    // Write the header of the xml file
     xml_installed << "<?xml version=\"1.0\"?>" << std::endl;
-    xml_installed << "<addons  xmlns='http://stkaddons.net/'>"
+
+    // Get server address from config
+    std::string server = UserConfigParams::m_server_addons;
+    // Find the third slash (end of the domain)
+    std::string::size_type index = server.find('/');
+    index = server.find('/', index + 2) + 1; // Omit one slash
```
leyyin: This is not that portable. Optimally you should find the last slash. If the domain is: http://example.com/somefolder/.

auriamg:
Hi, please avoid mixing several unrelated changes in the same pull request, makes it a lot easier for us to review and merge

Flakebi:
Hi, sorry for mixing up changes again, each time I'm working on stk I think it's not worth creating a separate pull request for a simple change with only a few lines... I tried to split the changes into several commits so it should be easier to review the changes. As far as I know, the changes I made are more or less cosmetic changes. However, I'll try avoiding mixing fixes and other changes in the future (btw, is it ok to have more than one fix (with only a few lines) in one pull request?).

I commented out the unused code because the compiler generates a warning each time the file is compiled. To make clear that the function may be used in the future, I added comments above the code.

For now, the only warnings left, are in irrlicht and bullet. When doing a release build, there are more warnings because some variables are only used in `assert`, should I put the code directly into the asserts?

hiker:
It looks like at least some files are using tabs - though on second look perhaps that's in irrlicht only, in which case it is ok :)  (perhaps we should replace all tabs with spaces in irrlicht, doesn't make any sense to have different styles).

In case of graphical restriction a change of line ending style(?), that should be fixed first. At least in label_widget and CIrrDeviceLinux it appears code is commented out - just remove it (and simplify the if around that code). Any code related to SKID_OLD should actually be removed.
And pllleeeeease - avoid such mam

konstin:
> It looks like at least some files are using tabs - though on second look perhaps that's in irrlicht only, in which case it is ok :) (perhaps we should replace all tabs with spaces in irrlicht, doesn't make any sense to have different styles).

I've written `tootls/whitespaces.py` to automatically replaces tabs with whitespaces and trim trailing whitespaces. (Just in case anyone needs that)

Flakebi:
Thanks for reviewing this pr! And sorry for the huge amount of changes :/
I removed the code that was commented out and the SKID_OLD related code.
The changes in the graphical restrictions seem to be gone with the merge, you already added the `const` to fix the warnings :-)

lib/irrlicht/source/Irrlicht/CGUIContextMenu.cpp
```diff
@@ -129,8 +129,8 @@ void CGUIContextMenu::setSubMenu(u32 index, CGUIContextMenu* menu)
 	if (index >= Items.size())
 		return;
 
-    if (menu)
-        menu->grab();
+	if (menu)
+		menu->grab();
```
auriamg: You changed spaces into tabs, could you please configure your editor correctly to not do that and fix these? This uselessly clutters the pull request

PR #1919, opened 2015-01-22
Fix sending of 'client-quit' request on quit. Also fix abortable requests.

leyyin:
When executing a request there was no check to see if is abortable, only the request manager was consulted in this regard, ignoring the option to the individual requests.


konstin:
You should definitely try to rebase/squash thise commits: http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

PR #1956, opened 2015-02-04
Use camera position for LoD instead of the kart position, fixes #1742

### LINKED ISSUE #1742: LoD is based off of kart's position instead of camera's
Makes the new FPS camera slightly less useful, as it can't capture highly detailed models unless the kart is right beside it.

Flakebi:
As vlj uses separate cameras for the shadows, the LoD nodes can take the camera position now (fixes #1742).

### DIFF
--- a/src/graphics/lod_node.cpp
+++ b/src/graphics/lod_node.cpp
@@ -77,12 +77,7 @@ int LODNode::getLevel()
     Camera* camera = Camera::getActiveCamera();
     if (camera == NULL)
         return (int)m_detail.size() - 1;
-    AbstractKart* kart = camera->getKart();
-    // use kart position and not camera position when a kart is available,
-    // because for some effects the camera will be moved to various locations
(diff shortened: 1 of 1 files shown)


auriamg:
Assigning to 0.8.3 since it's a little late in the 0.8.2 release process IMO

hiker:
@auriamg Should this be merged now? There was a bit of a discussion in #1742, but this pull request appears to be after that, so I'd guess it takes the discussion into account(?)

Flakebi:
I think it's safe to merge this. We are not immediately before a release now and that should be one of my safest pull requests. :wink:
vlj searched for references of the used camera and I grepped the code for the camer usages twice and we both found nothing (bad). So this patch shouldn't have any negative influence (in the meaning of unexpected shadow rendering).

auriamg:
Seems fine, will merge, thanks for the contribution

PR #1958, opened 2015-02-04
set "Adjust Lights" base position to camera position

konstin:
Copied from #1956


konstin:
ping @hiker

auriamg:
It's still pretty hackish but seems a bit better than before ;)

PR #1960, opened 2015-02-05
Fix clang build, freeing memory too early

Flakebi:
Only the top half of the last lap message was displayed when SuperTuxKart was compiled with clang (version 3.5.1), the bottom half was an empty string. @hiker found out that clang deletes the string too fast, so caching it as a local variable fixes that problem.

### DIFF
--- a/src/modes/linear_world.cpp
+++ b/src/modes/linear_world.cpp
@@ -352,11 +352,11 @@ void LinearWorld::newLap(unsigned int kart_index)
 
         // Store the temporary string because clang would mess this up
         // (remove the stringw before the wchar_t* is used).
-        core::stringw css = kart->getName();
+        core::stringw kartName = kart->getName();
 
         //I18N: as in "fastest lap: 60 seconds by Wilber"
         irr::core::stringw m_fastest_lap_message =
-            _C("fastest_lap", "%s by %s", s.c_str(), css.c_str());
+            _C("fastest_lap", "%s by %s", s.c_str(), kartName);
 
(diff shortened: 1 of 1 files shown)


src/modes/linear_world.cpp
```diff
@@ -350,10 +350,13 @@ void LinearWorld::newLap(unsigned int kart_index)
...
         std::string s = StringUtils::timeToString(time_per_lap);
 
-        irr::core::stringw m_fastest_lap_message;
+        // Store the temporary string because clang would mess this up
+        // (remove the stringw before the wchar_t* is used).
+        core::stringw css = kart->getName();
+
         //I18N: as in "fastest lap: 60 seconds by Wilber"
-        m_fastest_lap_message += _C("fastest_lap", "%s by %s", s.c_str(),
-                                    core::stringw(kart->getName()));
```
hiker: I noticed a few things we should have tried first:
1) can you try removing the core::stringw, i.e. just leaving
kart->getName()
2) if this doesn't work, what about kart->getName().c_str()
?

Flakebi:
I improved it so it doesn't convert create a new stringw. Passing kart->getName() directly or even core::stringw(kart->getName().c_str()) doesn't help.

hiker:
I tried to remove the unnecessary string copy when I applied this patch - please let me know if this should not work on clang, then I'll go back to the original patch.

Thanks a lot!
   Joerg

Flakebi:
It works :-)

PR #1963, opened 2015-02-05
Sort resolutions by size

Flakebi:
With this change, the resolutions in the options screen are sorted by their size, so the greatest resolution is on the right and the lowest on the left.

### DIFF
--- a/src/guiengine/widgets/dynamic_ribbon_widget.cpp
+++ b/src/guiengine/widgets/dynamic_ribbon_widget.cpp
@@ -518,6 +518,7 @@ void DynamicRibbonWidget::clearItems()
     m_scroll_offset = 0;
     m_max_label_width = 0;
 }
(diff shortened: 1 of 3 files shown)


auriamg:
While this is a good idea, I think you should sort the original values as retrieved from irrlicht, it's not too clean that you use scanf to extract the values from the labels, especially since we had the int value at hand not long before when building the labels

Flakebi:
Ok, I had the scanf-part from the code that actually sets the resolution (in `eventCallback`).
I changed it to sort the resolutions and add them to the list afterwards.
I'm not sure if I should remove the sorting code ribbon widget, so I left it there for now.

auriamg:
Thanks a lot! I had forgotten that scanf was already used, but your new version is clearly far superior so thanks for the big code improvement

PR #1969, opened 2015-02-09
Fix music and sound when starting with -R and add brackets to a calculation

Flakebi:
* I must have forgotten to add brackets there. It shouldn't have much impact, but it still looks better with brackets
* The bigger part of this pull request starts sound and music when stk is started with '-R' and the ready-set-go-phase is skipped. I can do most of the initialization in the `reset()` function but starting the music there doesn't work so I moved it to the `MUSIC_PHASE` which also adds the music author at the bottom of the screen. The only functional change I made is setting `m_race_now` to false after that's done, so if another race is started afterwards the read-set-go-phase isn't skipped.

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -2522,8 +2522,8 @@ void Kart::updateGraphics(float dt, const Vec3& offset_xyz,
     {
         // fabs(speed) is important, otherwise the negative number will
         // become a huge unsigned number in the particle scene node!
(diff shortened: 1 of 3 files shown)


Flakebi:
I moved the `startMusic()` call completely to the `reset()` method so a valid music is set before a race is started. That fixes `-N`, which crashed because the SFXManager wants to update the music, but there is no music set when the start screen is skipped.

PR #1994, opened 2015-02-21
Fix transparency of fallback characters

Flakebi:
This patch fixes transparency of fallback characters in title fonts.
The brackets in the credit screen didn't fade out and in which was caused by the shader that didn't respect transparency (I hope this isn't expected behaviour)
and the colors that didn't change because they were static.

Additionally I changed the order of yellow and orange, so it fades from yellow to orange now.


auriamg:
Thanks for the patch, it seems to work well

PR #2005, opened 2015-02-24
Add a tooltip to explain usage of per-player handicaps

Flakebi:
Nothing more than the title says. Feel free to change the text.

### DIFF
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -1849,7 +1849,7 @@ void Skin::drawTooltip(Widget* widget, bool atMouse)
     if (atMouse)
     {
         pos = irr_driver->getDevice()->getCursorControl()->getPosition()
-            + core::position2di(15, 15);
+            + core::position2di(10 - size.Width / 2, 20);
     }
 
     core::recti r(pos, size);
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -1609,6 +1609,12 @@ void Skin::drawCheckBox(const core::recti &rect, Widget* widget, bool focused)
                                              0 /* no clipping */, 0,
                                              true /* alpha */);
     }
+
+    if (focused && widget->hasTooltip())
+    {
+        m_tooltip_at_mouse.push_back(true);
+        m_tooltips.push_back(widget);
+    }
 }   // drawCheckBox
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


deveee:
I just noticed that graphics settings level tooltip is not moveable. I think it is better solution than allow to move this tooltip in whole screen.

And TBH I don't like "handicap" word because it suggests adjusting game for blind or deaf men, not higher difficulty.

Maybe something like:
"Enable difficulties for players"
And in tooltip:
"It allows to choose players with additional difficulties in kart selection screen."

What do you think?

hiker:
One comment: while I am not a native speaker, 'handicap' appears to be a quite normal term to compare strength in games with (e.g. in Go and Gold people have a  handicap). But we should go through the strings to try to make the term clearer, e.g.: give a handicap would likely be clearer. Maybe "Disadvantage players"  or "slow down" (though that might not hint enough why this option exists, while handicap in terms of games appears to be frequently used to compare strength and to offset different strengths).

deveee:
Maybe just "handicap for players" instead of "handicapped players"?

auriamg:
I would indeed suggest reworking this text a little :
```Handicap stronger players in the kart selection```

But it's actually quite hard to find something that is both clear and short.
Maybe something like

```Let each players optionnally select a handicap on the kart selection screen```

Any other suggestions? I really want to avoid the feature seeming like it's an accessibility feature

Flakebi:
I made the tooltip moveable because it looks better in my opinion (and it's consistent with the tab-tooltips). It's only visible when the mouse hovers the checkbox.

I have another suggestion for the string:
`Players can select handicapped profiles on the kart selection screen`
or `Players can select more difficult profiles on the kart selection screen`

auriamg:
I like your suggestions, why not a hybrid?

```Players can select handicapped (more difficult) profiles on the kart selection screen```

or perhaps

```Players can select handicapped profiles (higher difficulty) on the kart selection screen```

Flakebi:
The explanation is a good idea, I would like to take the first.
The second is maybe a bit unclear because higher difficulty is similar to easy, medium, hard and supertux difficulty, which could indicate that the kart is faster instead of slower :wink:.

Flakebi:
@deveee sorry, now I noticed that you can indeed move the label around the whole screen...
I wanted to work on the focus handling sometime...

The good thing is, I can move the tooltip to see the whole (new) text because otherwise it's cut off at the right side.

I centered tooltips and now it's visible without moving the mouse.

PR #2025, opened 2015-03-07
Sliders react to mouse

ZyperPL:
You can change slider/spinner's value by mouse click.
It works only on gauge spinners.

### DIFF
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -1378,6 +1378,20 @@ void Skin::drawSpinnerBody(const core::recti &rect, Widget* widget,
                                             0 /* no clipping */, 0,
                                             true /* alpha */);
 
+        // ---- Set value by mouse position
+        const float max_value = (float)(w->getMax() - w->getMin()) /
+                                       (w->getMax() - w->getMin());
+
+        const core::recti mouse_area(rect.UpperLeftCorner.X + handle_size,
+                                    rect.UpperLeftCorner.Y,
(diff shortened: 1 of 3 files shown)


auriamg:
Hi,

this is a step in the right direction, but I see a few things that could be improved

1) I don't quite get the following line. Unless there's something I missed, this division always yields 1 :

```
const float max_value = (float)(w->getMax() - w->getMin()) / (w->getMax() - w->getMin());
```

2) Placing the click handling code in the skin is very hackish, why not move the click handling code directly in onClick where it belongs?

ZyperPL:
@auriamg 
Ok. I moved all the code to onClick() function of widget SpinnerWidget. 
But there's one thing that is wrong, I don't have m_left_border and image's height of SkinConfig::m_render_params. Is it possible to get those values from inside Widget class?

auriamg:
Hi, that's a lot better, nice work! One last little think, hardcoding 110 and 128 is uneeded since you have access to coordinates in this way :

```
core::recti body_rect = m_children[1].getIrrlichtElement()->getAbsolutePosition();
```

This coordinate is not 100% perfect due to the skin adding some margins, but should suffice to work reasonbly well

ZyperPL:
Thanks @auriamg 
I've changed that. Everything works, but now it's not pixel-perfect.
However I think user won't see the difference.

auriamg:
Hi, thanks I've applied your pull request!
I made a few small changes, like remove unused code (Calculating x and y was never used) and also using roundf instead of adding +0.5

PR #2028, opened 2015-03-08
Improvements for the first person camera

Flakebi:
This pull request contains two improvements/changes for the fps camera:
* I changed the way, how the mouse position is handled: There is now a limit for the movement per event, that works better than ignoring a number of events.
* I added a new "mode" for the first person camera, that attaches it to a kart. That allows to drive a kart from a different (fixed) position than normal.


samuncle:
I have tested and everything seems to work perfectly. Hopefully I won't break something ^^.

PR #2030, opened 2015-03-09
added ability to tab backwards when shift is pressed

xXFlorianXx:
My first commit.

I compiled the game and was annoyed that i couldn't tab backwards in the menus...


deveee:
I wanted to do it rather long time ago, but I forgot about it ;) There was even a bug report somewhere, but probably marked as invalid.

I think that use ```event.KeyInput.Shift``` should be safer. Irrlicht already checks it for us.

deveee:
Fixed in 0a7234df84b9d38889fc54ab117d6c7d59ff0784. Thanks a lot!

PR #2034, opened 2015-03-14
Highscores update

ZyperPL:
On track_info_screen high scores were not visible or updated (after entering or after changing number of ai).
Now they are updated on init() or after click on ai-spinner.


ZyperPL:
Sorry for those unwanted commits above "Fixed highscores update".

PR #2035, opened 2015-03-14
Fixed highscores update

ZyperPL:
High scores are now updated on init() and after change of number of AIs.


auriamg:
Thanks, merged

PR #2042, opened 2015-03-23
Make CREDITS file to be UTF-8 and seen as a text file by GIT.

leyyin:
It was very annoying when someone modified the CREDITS file, you could not see the changes made ;)


deveee:
I think it's good idea because we had problems with this file in first beta (wrong encoding). But it needs also modifications in code. Atm. utf8 credits file will just crash STK.

leyyin:
Yes, I forgot about that, currently it segfaults.

hiker:
use vimdiff, this can handle utf-16 files natively. Otherwise iirc git has options to define an encoding attribute for files, and my understanding is that git diff will then use this to convert the files before comparing. Check:
git help gitattributes and search for encoding.

Flakebi:
diff itself can handle utf-16 too (I think it's `-t` or so to force text mode). GitHub can't and if I remember correctly, setting the value in .gitattributes reads the file as utf-8 which doesn't look better because every second char is `\0` (only affects git, not GitHub). :wink:

PR #2053, opened 2015-03-28
Fix glVertexAttribDivisor redefinition warning

tobbi:
This fixes this warning:

supertuxkart/stk-code/src/graphics/gl_headers.hpp:16:17: warning: 
      'glVertexAttribDivisor' macro redefined
#        define glVertexAttribDivisor glVertexAttribDivisorARB
                ^
supertuxkart/stk-code/lib/glew/include/GL/glew.h:2382:9: note: 
      previous definition is here
#define glVertexAttribDivisor GLEW_GET_FUN(__glewVertexAttribDivisor)
        ^

### DIFF
--- a/src/graphics/gl_headers.hpp
+++ b/src/graphics/gl_headers.hpp
@@ -13,6 +13,9 @@ extern "C" {
 #    include <OpenGL/gl3.h>
 #    define OGL32CTX
 #    ifdef GL_ARB_instanced_arrays
+#        ifdef glVertexAttribDivisor
+#            undef glVertexAttribDivisor
+#        endif
 #        define glVertexAttribDivisor glVertexAttribDivisorARB
 #    endif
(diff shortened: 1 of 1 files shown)


tobbi:
That would result in a behaviour change, right? I don't know about the implications of that. 
It would then switch from being defined as
#define glVertexAttribDivisor glVertexAttribDivisorARB
to 
#define glVertexAttribDivisor GLEW_GET_FUN(__glewVertexAttribDivisor)

Is that what we want?

hiker:
@auriamg preferred the suggested solution since it will 'more likely' be stable. But @tobbi has a point, too. So what about: tobbi's patch for 0.9; and auria's patch for master?

auriamg:
Honestly I don't know enough to give an informed decision.

SInce tobbi's patch is basically status quo minus warning, I suggest to go with that

PR #2057, opened 2015-03-29
Allocate memory dynamically in fribidize

Flakebi:
Fribidize crashed when strings in RTL languages exceeded 512 chars length.
I tested this code on Windows but not on linux where the string shouldn't be copied.

### DIFF
--- a/src/utils/translation.cpp
+++ b/src/utils/translation.cpp
@@ -283,26 +283,22 @@ const wchar_t* Translations::fribidize(const wchar_t* in_ptr)
 #if ENABLE_BIDI
     if(this->isRTLLanguage())
     {
-        const int FRIBIDI_BUFFER_SIZE = 512;
-        FriBidiChar fribidiInput[FRIBIDI_BUFFER_SIZE];
-        int len = 0;
-        int n = 0;
-        //std::cout << "fribidi input : ";
-        for (n = 0; ; n++)
-        {
-            fribidiInput[n] = in_ptr[n];
-            //std::cout << (int)fribidiInput[n] << " ";
-            len++;
+        std::size_t length = wcslen(in_ptr);
+        FriBidiChar *fribidiInput;
 
-            if (n == FRIBIDI_BUFFER_SIZE-1) // prevent buffeoverflows
-            {
-                Log::warn("Translations::fribidize", "translated string too long, truncating");
-                fribidiInput[n] = 0;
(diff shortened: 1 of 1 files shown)


src/utils/translation.cpp
```diff
@@ -283,26 +283,22 @@ const wchar_t* Translations::fribidize(const wchar_t* in_ptr)
...
 
-            if (n == FRIBIDI_BUFFER_SIZE-1) // prevent buffeoverflows
-            {
-                Log::warn("Translations::fribidize", "translated string too long, truncating");
-                fribidiInput[n] = 0;
-                break;
-            }
-            if (fribidiInput[n] == 0) break; // stop on '\0'
+        if (sizeof(wchar_t) == sizeof(FriBidiChar))
+            fribidiInput = (FriBidiChar*) in_ptr;
```
auriamg: I am not sure that this is right. I believe fribidiInput will be modified, and we certainly don't want to modify the original string while translating, so pretty sure we do need a copy

src/utils/translation.cpp
```diff
@@ -283,26 +283,22 @@ const wchar_t* Translations::fribidize(const wchar_t* in_ptr)
...
+        else
+        {
+            // On windows FriBidiChar is 4 bytes, but wchar_t is 2 bytes.
+            // So we simply copy the characters over here (note that this
+            // is technically incorrect, all characters we use/support fit
+            // in 16 bits, which is what irrlicht supports atm).
+            fribidiInput = new FriBidiChar[length + 1];
```
auriamg: This is incompatible with the
```
delete[] fribidiInput;
```
statement later

src/utils/translation.cpp
```diff
@@ -283,26 +283,22 @@ const wchar_t* Translations::fribidize(const wchar_t* in_ptr)
...
+            std::memset(fribidiInput, 0, (length + 1) * sizeof(FriBidiChar));
+            for (std::size_t i = 0; i <= length; i++)
+                fribidiInput[i] = in_ptr[i];
```
auriamg: Why use memset here since there is a for loop immediately after that sets the content of the buffer? Presumably you forgot to zero-terminate the string

Flakebi:
I tested it on linux too and it seems to work without problems.

auriamg:
This pull request seems included in #2060

PR #2060, opened 2015-03-29
Fix some issues with RTL languages

Flakebi:
I fixed some issues with RTL languages and removed some unnecessary code.
There were some missing `fribidize` invocations and I changed the position the track/gp-name to the right side of the screen to avoid issues with long names.

There's a new debug option to force text to be right-to-left in `translations.cpp`.

Edit: I forgot to mention that this pull request depends on the changes I made in  #2057 (Allocate memory dynamically in fribidize).

### DIFF
--- a/src/states_screens/edit_gp_screen.cpp
+++ b/src/states_screens/edit_gp_screen.cpp
@@ -155,7 +155,7 @@ void EditGPScreen::init()
     {
         LabelWidget* header = getWidget<LabelWidget>("title");
         assert(header != NULL);
-        header->setText(m_gp->getName(), true);
+        header->setText(translations->fribidize(m_gp->getName()), true);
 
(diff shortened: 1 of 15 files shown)


src/states_screens/edit_gp_screen.cpp
```diff
@@ -281,7 +281,8 @@ void EditGPScreen::setModified(const bool modified)
 
     LabelWidget* header = getWidget<LabelWidget>("title");
     assert(header != NULL);
-    header->setText(m_gp->getName() + (modified ? L" (+)" : L""), true);
+    //I18N: Indicate that the grand prix is modified and not saved
+    header->setText(modified ? _(L"%s (+)", m_gp->getName()) : L"", true);
```
auriamg: This makes "%s (+)" translatable, do we really need that to be translatable?
In any case we have a string freeze so we can't introduce new strings at this stage

Issue #2073, opened 2015-04-04
SuperTuxKart fails to start when /data exists


jturner314:
Version: 0.9rc1-1 (Arch Linux package)

Description
--------------
When a directory `/data` exists, STK experiences a fatal error and fails to start.

Steps to reproduce
--------------------
1. Install `supertuxkart`. The Arch Linux package places the `supertuxkart` binary at `/usr/bin/supertuxkart` and the data files in `/usr/share/supertuxkart/data`.
2. Create a `/data` directory. (That's a directory named `data`, immediately under the root directory.)
2. Run `supertuxkart` at the command line.

Actual behavior
-----------------
This results in the following output:
```
$ supertuxkart
[verbose  ] main: Error messages and other text output will be logged to /home/jim/.config/supertuxkart/0.8.2/stdout.log.
[info   ] [FileManager]: Data files will be fetched from: '../../data/'
[info   ] [FileManager]: User directory is '/home/jim/.config/supertuxkart/0.8.2/'.
[info   ] [FileManager]: Addons files will be stored in '/home/jim/.local/share/supertuxkart/addons/'.
[info   ] [FileManager]: Screenshots will be stored in '/home/jim/.cache/supertuxkart/screenshots/'.
[info   ] [FileManager]: User-defined grand prix will be stored in '/home/jim/.local/share/supertuxkart/grandprix/'.
[warn   ] [FileManager]: Directory 'challenges' not found, aborting.
[warn   ] [FileManager]: Directory 'fonts' not found, aborting.
[warn   ] [FileManager]: Directory 'gfx' not found, aborting.
[warn   ] [FileManager]: Directory 'grandprix' not found, aborting.
[warn   ] [FileManager]: Directory 'gui' not found, aborting.
[warn   ] [FileManager]: Directory 'library' not found, aborting.
[warn   ] [FileManager]: Directory 'models' not found, aborting.
[warn   ] [FileManager]: Directory 'music' not found, aborting.
[warn   ] [FileManager]: Directory 'tracks' not found, aborting.
[warn   ] [FileManager]: Directory 'sfx' not found, aborting.
[warn   ] [FileManager]: Directory 'shaders' not found, aborting.
[warn   ] [FileManager]: Directory 'skins' not found, aborting.
[warn   ] [FileManager]: Directory 'textures' not found, aborting.
[warn   ] [FileManager]: Directory 'po' not found, aborting.
[fatal  ] [FileManager]: Not all assets found - aborting.
```

[...]

PR #2074, opened 2015-04-04
Re-prioritize SUPERTUXKART_DATADIR build option

jturner314:
This is a draft proposal for fixing #2073, and I have not tested it at all.

Before, when selecting the directory containing the data files, STK checked whether some paths such as `../../data` existed before considering the `SUPERTUXKART_DATADIR` build option. This commit makes the build option take priority over those checks, although it remains lower priority than the `SUPERTUXKART_DATADIR` environment variable.

Now, when `SUPERTUXKART_DATADIR` is defined to the correct value during compilation, the existence of an empty `/data` directory (when the actual data files are in e.g. `/usr/share/supertuxkart/data`) does not cause STK to crash on startup.

### DIFF
--- a/src/io/file_manager.cpp
+++ b/src/io/file_manager.cpp
@@ -160,6 +160,10 @@ FileManager::FileManager()
 #ifdef __APPLE__
     else if( macSetBundlePathIfRelevant( root_dir ) ) { root_dir = root_dir + "data/"; }
 #endif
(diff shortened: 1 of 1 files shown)


hiker:
I've implemented the suggestions from #2073 (which might find other installed stk data files instead).

PR #2076, opened 2015-04-06
Fix message dialog for RTL languages

### LINKED ISSUE #1823: [RTL] Wrapped text appears from bottom to top.
from bottom to top.
Wrapped text appears

screenshot - 141214 - 140847

Flakebi:
This should be the easiest way to fix #1823. When a string is loaded from the xml file, the RTL option of irrlicht is set to that of the current language, per default it would be `false`.


auriamg:
This fix is not quite right, because if a RTL translation is incomplete, it may result in an untranslated (english) string to be displayed upside down. It also adds a dummy string that will appear in transifex even though it's unused

auriamg:
Fixed by merging other branch in 9a905cb5796dd254eeb9fae009356f3233852bc0

PR #2103, opened 2015-04-23
Fix compilation with -DUSE_FRIBIDI=0

nado:
Last commits introducing some new functions for fribidi didnt have preprocessor conditions, this lead to a break in compilation without fribidi.


auriamg:
Thanks for the fix!

PR #2104, opened 2015-04-23
Fix #2102.

### LINKED ISSUE #2102: Move parsing of `--log` and `--help` before the execution of any other command generating output.
I think the title is clear enough, but here an example : https://gist.github.com/nado/7cfb5c54939233699ac7

nado:
Moving output modifier and early exit options (--help, --version,
--log, etc) earlier in the execution so nothing useless is printed.

English is not my native language and I have to admit, I hesitated a moment about the function name. Tell me if its not clear enough.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -555,14 +555,10 @@ void cmdLineHelp()
 }   // cmdLineHelp
 
 //=============================================================================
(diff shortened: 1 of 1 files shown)


auriamg:
Thanks for the fix!

PR #2106, opened 2015-04-24
Simplify StringUtils::insertValues

Flakebi:
I removed all functions that take a static amount of arguments, so the variadic ones are used always now. This should make the code a lot easier to read.

Unfortunately, the simple `insertValues` template function that passes it's only argument isn't selected by the compiler anymore. This introduces some bugs in combination with `fribidize` (because it returns a `wchar_t*` of a maybe freed `stringw`). The only place where I found that issue so far is the grand prix editor.


hiker:
Is this ready to be merged? While I am not that much of a friend of variadic templates (people tend to find it difficult to understand them, and they were not well supported with older compilers), we are already using them in the graphics, so I don't see a reason not to include them. While I didn't actually run the code, the patch itself looks fine - though it's still marked as WIP, so not sure. @auriamg include it?

src/utils/string_utils.hpp
```diff
@@ -136,210 +136,66 @@ namespace StringUtils
...
-    }   // insertValues(s, v1, v2)
-    // ------------------------------------------------------------------------
-    /** Shortcut insert_values taking three values, see above for
-     *  full docs.
-     *  \param s String in which all %s, %d are replaced.
-     *  \param v1 Value to replace.
-     */
-    template <class T1>
-    std::string insertValues(const std::string &s, const T1 &v1)
+    /** Intermediate struct to fill a vector using variadic templates */
```
leyyin: Can you explain more in detail what this does? Variadic templates are black magic ;) and having a nice long explanation will definitely help.
Also can we not use double underscore '__'? As far as I know the standard says that two underscore is reserved for the compilers internal use.

src/utils/translation.cpp
```diff
@@ -358,11 +359,34 @@ Translations::Translations() //: m_dictionary_manager("UTF-16")
...
 #if ENABLE_BIDI
     if(this->isRTLLanguage())
     {
+        // Test if this string was already fribidized
+        for (std::vector<std::pair<wchar_t*, wchar_t*> >::iterator it =
+             m_fribidized_strings.begin(); it != m_fribidized_strings.end(); it++)
+        {
+            if (wcscmp(it->first, in_ptr) == 0)
+                return it->second;
+        }
```
auriamg: One thing I don't like is that is that on everytime we get a string, this will perform a linear aka O(n) search, performing a string comparison on every step. While probably not catastrophic, it's not very good performance wise

Also I'm not sure I understand why using variadic templates would cause problems with fribidi?

PR #2117, opened 2015-04-25
cppcheck iterator prefix fixes

tobbi:
Changes iterator++ to ++iterator because cppcheck claims it's faster.

### DIFF
--- a/src/achievements/achievement_info.cpp
+++ b/src/achievements/achievement_info.cpp
@@ -103,7 +103,7 @@ irr::core::stringw AchievementInfo::toString() const
     {
     case AC_ALL_AT_LEAST:
         // If all values need to be reached, add up all goal values
-        for (iter = m_goal_values.begin(); iter != m_goal_values.end(); iter++)
+        for (iter = m_goal_values.begin(); iter != m_goal_values.end(); ++iter)
         {
             count += iter->second;
         }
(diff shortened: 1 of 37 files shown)


deveee:
I checked it in a single example:

```
#include <cstdio>
#include <vector>

int main ()
{
	std::vector<int> tab;
 
	tab.push_back(1);
	tab.push_back(2);
	tab.push_back(3);
 
	std::vector<int>::iterator it;

	//for (it=tab.begin(); it != tab.end(); it++)
	for (it=tab.begin(); it != tab.end(); ++it)
	{
		printf("%i\n", *it);
	}
	
	return 0;
}
```

It gives slightly different output when you compile it without optimizations (I didn't check which one is faster).

But it gives byte in byte exactly the same output when you use at least -O1 as an optimization parameter. It doesn't matter if you use it++ or ++it.

DanWin:
The optimization here is to remove the need of creating a temporary variable when increasing the value. As long, as the complier knows the class e.g. int, it can automatically optimize this operation. However when increasing the value of an iterator, the compiler might not know, whether it can be optimized or not and thus it is good practise to use prefix instead of postfix operators for iterations. I found a good explanation of the performance impact here: http://www.tantalon.com/pete/cppopt/asyougo.htm#UsePrefixOperators

hiker:
We are aware of the reasons for the warnings, we have looked at this issue in the past, and as a result our coding style (http://supertuxkart.sourceforge.net/Coding_Style#Coding) clearly states that we like the postfix operators more.
Now, would I create a code base with new coding rules and a new team from scratch, I might actually use prefix - even though I don't like the look of it (it just feels wrong to first have the operator without specifying what it applies to ... "i is 1 plus i", it feels easier to understand that "i is i plus 1" - and I also think that a=i++ is plain code obfuscation ;) ). But now we devs are all used to a certain style, like it, it does not have any performance impact in our cases, so no need to change this.

Issue #2132, opened 2015-04-26
Missing translation strings in 0.9


Wuzzy2:
I have discovered some untranslatable strings in 0.9.

They are:
* “Mosquito Hunter”
* “Please select a Grand Prix”

Post here if you find more untranslatable strings in 0.9.

hiker:
Please select a Grand Prix was fixed recently, and I do see Mosquito Hunter in the old pot file. What makes you think that this is not translatable?

I'll leave this ticket open for now so what we make sure that the next release will indeed include the fix for "Please select a Grand Prix".

Wuzzy2:
Oops, I was indeed wrong about “Mosquito Hunter”.

nado:
No, its just that the counter is not incremented until success, that could be interesting to update to the best while we dont get the achievement.

hiker:
So it seems like it's all fixed ;)
Thanks!
   Joerg

Issue #1815, opened 2014-12-18
Reversed RTL text (track name) in unlocked track screen


GreenLunar:
neercs kcart dekcolnu ni (eman kcart) txet LTR desreveR

deveee:
@GreenLunar Maybe it was workarounded directly in translations. Tracks names can't be workarounded because they are displayed in different palces.

Anyone from Arabia? :P

GreenLunar:
No workarounds in Hebrew translation as far as I can see.

I have seen a clever RTL workaround in SMPlayer, where strings that appear both in MPlayer (where RTL strings get reversed like in some parts in STK as reported above) AND Qt (SMPlayer) interface are translated like םולש and have LRO mark to reverse them in Qt interface to שלום.

auriamg:
I would recommend against writing translations in reverse, this seems especially unclean. I know there's no one available at the moment, but getting some contributor speaking an RTL language is still the best option

auriamg:
We have recently merged RTL fixes from Seebi, could you please confirm if it's now fixed?

cap44:
Looks good on my end: 
![overworld-2015 03 30_07 44 57](https://cloud.githubusercontent.com/assets/8017401/6890293/eb2a6c9c-d6b0-11e4-8502-47d0260a02f5.png)

@GreenLunar is it fixed for you?

GreenLunar:
It seems so.  Thank you @Seebi

GreenLunar:
Or @ Flakebi

GreenLunar:
Or @Flakebi

GreenLunar:
Texts within `%s` in `Completed achievement "%s".` is reversed.

PR #2133, opened 2015-04-26
Fix some translation and bidi issues

### LINKED ISSUE #2125: Custom Grand Prix name is “translated” by Gettext
The name of a custom grand prix is subject to Gettext. This should not be the case.

For example, if you call your grand prix “Help”, its name will be displayed as “Hilfe” when you use the German locale.

Very weird, and it may lead to unexpected results. IMO all custom grand prix names should be left untouched.

How to reproduce:
* Set German locale
* Start STK
* Start grand prix editor
* Create new grand prix and call it “Help”
* Look at the displayed name

STK version: 0.9

Flakebi:
I hope I didn't mix up too many unrelated changes here. There are some little changes that
- fix #2125,
- the second part of #2132 (“Please select a Grand Prix”),
- the achievement display (@GreenLunar mentioned it in #1815),
- use isRTLText instead of isRTLLanguage in the BubbleWidget (if text is untranslated) and
- I partially fixed the TEST_BIDI debug option (it's still not working when you try to choose a kart in multiplayer).

The bigger change (fae12f3) is to fix a bidi issue in the multiplayer KartSelection screen where a multiline string (with `\n`) should be displayed. That caused issues with fribidi so now the text is splitted into lines before it gets fribidized.

Update: TEST_BIDI works in the multiplayer kart-selection screen now

### DIFF
--- a/src/achievements/achievement_info.hpp
+++ b/src/achievements/achievement_info.hpp
@@ -113,7 +113,7 @@ public:
     irr::core::stringw getDescription() const { return _(m_description.c_str()); }
     // ------------------------------------------------------------------------
     /** Returns the name of this achievement. */
-    irr::core::stringw getName() const { return _(m_name.c_str()); }
+    irr::core::stringw getName() const { return _LTR(m_name.c_str()); }
     // ------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


auriamg:
Seems good, thanks

PR #2188, opened 2015-05-12
Don’t sort strings alphabetically (makes the translators’ work much easier)

huftis:
Here’s a patch to remove the sorting of the strings for translation. Currently, they’re sorted alphabetically, which makes translation more difficult and error-prone, since one can’t see the strings in proper context (e.g., the strings ‘Ready!’, ‘Set!’ and ‘Go!’ appear in completely different parts of the translation file). After applying the patch, the strings will appear in ‘natural’ order, i.e., in the source code order.

### DIFF
--- a/data/po/update_pot.sh
+++ b/data/po/update_pot.sh
@@ -37,13 +37,13 @@ echo "---------------------------"
 echo "    Generating .pot file..."
 
 # XML Files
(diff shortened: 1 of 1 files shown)


auriamg:
I have no idea if transifex will respect the order, but why not, this seems good

Issue #2187, opened 2015-05-10
The letter Æ is missing in the title font


huftis:
The letter Æ is missing in the title font (`title_font_2.png`, I believe). This is needed for the Norwegian Nynorsk (`nn`) translation, and possibly a few other languages too (Danish and Icelandic?).

The problem can be seen when one enters the tutorial bubble in story mode. The name of the level should be OPPLÆRINGSBANE, but the Æ is shown as an ugly lowercase æ from a different font.

(Note that the other two ‘special’ letters used in Norwegian Nynorsk, Ø and Å, *is* included in `title_font_2.png`.)

Flakebi:
The problem is that the title font is not generated from an existing font, but afaik it was created by hand (#1814). If I get it right, the only solution would be to create the other letters too and include them in the font.

huftis:
This is now fixed by pull request #2189.

PR #2189, opened 2015-05-12
Add ‘Æ’ glyph to the title font

huftis:
This pull request adds a glyph for the letter ‘Æ’ to the title font. This fixes bug #2187.

The glyph was created by combining the existing ‘A’ glyph and ‘E’ glyph, and then retouching the result to make it look good.


auriamg:
Seems good, thanks

PR #2191, opened 2015-05-14
Improved English grammar and punctuation

huftis:
Here are some very minor grammar and punctuation fixes in the English strings. For example, in English, there shouldn’t be a space before a colon (e.g., it should be ‘Password:’, not ‘Password :’) (This is different than in French, where there *should* be a space.)

The commit also includes some punctuation changes to make similar strings more consistent (e.g., trailing periods).

Finally, the translation template files are updated with the changed strings (and some new ones which were added since the last translation template update).


hiker:
Hi,
I've applied this pull request (d4f3fc88d63662b865001c6352b543e67b446a00, no idea why this request wasn't automatically closed). Some of the changes had to be applied to svn (r16305) due to changes in our code base.

Thanks a lot!
   Joerg

Issue #2120, opened 2015-04-26
User config for 0.9 is still saved in the "0.8.2" folder 


konstin:
This is happens because the version number is hardcoded in `src/io/file_manager.cpp:827`. It would be better to evaluate the current version number at compile time using cmake.

hiker:
There is a constant in utils/constant.cpp that could be used. But the main reason not to use it is that we actually wanted people to be able to use 0.8.2 beta and 0.9rc configs (and as long as we don't change the config, even for future releases).

On the other hand, this needs at least to be documented in our release page, and perhaps discussed. I should have just given it an independent version number ('1'), instead of the STK version number - since this is what this version number actually means.

konstin:
> On the other hand, this needs at least to be documented in our release page, and perhaps discussed. I should have just given it an independent version number ('1'), instead of the STK version number - since this is what this version number actually means.

I changed the title accordingly. This issue can probably also be labeled as an easy one.

Alayan-stk-2:
Disconnecting entirely the folder version number from the release version number would be impractical for someone trying to edit the config files. (Imagine the confusion : "You run version 1.1.0 ? You must edit the files in the "3" folder. Version 1.2.1 ? That's the "4". Version 1.2.4 ? That's the "5".")

Using a version numbering system guaranteeing that all versions sharing a given numbering (say, 0.9.x) are config-compatible, and bumping the version number when incompatibilities arise, seems like the sanest approach.

PR #2196, opened 2015-05-17
User conf dir versioned naming appears to be wrong

captnfab:
Confdir name is versioned and seems to be outdated.
Maybe an automatic naming (using macros) would be more robust.


konstin:
That has already been discussed at #2120

hiker:
While it's not ideal, we want people who used e.g. the rc or beta to be able to keep on using their data files (esp. progress). So unless we break compatibility in the file format, we will keep this (indeed ... 'strange') version number.

Thanks for your patch!

PR #2198, opened 2015-05-18
Fix #2164 : Resized max length of mail address up to 254 chars

### LINKED ISSUE #2164: Arbitrary e-mail address length restriction of 50 characters
According to a string I found while translating, STK 0.9 appears to reject e-mail adresses which are longer than 50 characters (bytes/octets/…). The string is question is “Email has to be between 4 and 50 characters long!”. I have not actually tested if STK actually rejects e-mail adresses of this length. But if this applies, this is a problem.
Because this is an abitrary restriction, because e-mail addresses can be much longer, although they are rarely encountered.

According to RFC5321 (and its errata), valid e-mail adresses can be up to 254 characters long, so SuperTuxKart should be prepared to handle such addresses.

Sources:
* https://tools.ietf.org/html/rfc5321#section-4.5.3.1
* http://www.rfc-editor.org/errata_search.php?rfc=3696 (refer to errata ID 1690)

Wikipedia's informal description:
* https://en.wikipedia.org/wiki/Email_address#Syntax

Also, is there any justification for the 4 character limit? I am not sure if such addresses are even possible, but if you can show me the shortest length of any possible e-mail adress, I am willing to accept it as mandatory lower bound.

nado:
Also, i raised the min value to 5 since as i explained in the issue thread, we might want to be ready for a 1 character tld.

### DIFF
--- a/data/po/supertuxkart.pot
+++ b/data/po/supertuxkart.pot
@@ -1067,7 +1067,7 @@ msgid "Email"
 msgstr ""
 
 #: src/states_screens/register_screen.cpp:292
-msgid "Email has to be between 4 and 50 characters long!"
+msgid "Email has to be between 5 and 254 characters long!"
 msgstr ""
 
 #: src/states_screens/register_screen.cpp:298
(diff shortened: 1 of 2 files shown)


hiker:
Thanks!

PR #2195, opened 2015-05-16
Fix angelscript build on non-x86 arches - fixes #2119 and fixes #2194

jcowgill:
This is my fix for #2119 and #2194. I have compile tested it on x86_64 and mips64el. Incredibly I have no ARM hardware so I am currently building it there with QEMU. That'll probably take at least another hour though.

----
For the build failures on 64-bit arches, angelscript was using `__LP64__` to test for x86_64. Use the correct define instead.

For ARM architectures, ensure the native function calling file is built on all ARM arches and not just when building for android.

Issue #2194, opened 2015-05-15
STK 0.9 fails to build on armel/armhf/arm64/ppc64el/s390x


Vincent-C:
STK 0.9 (just uploaded today to Debian) fails to build on a variety of non-i386/amd64 architectures in Debian sid; see [here](https://buildd.debian.org/status/package.php?p=supertuxkart&suite=unstable) for build status and logs (you can click on any of the rows under the "Status" column for the entire log for that arch). Note that STK 0.8.1 built fine on all archs supported by Debian.

All of the build failures have this in common:
```
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp: In function 'asQWORD X64_CallFunction(const asQWORD*, int, funcptr_t, asQWORD&, bool)':
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%rcx' in 'asm'
     "%rdi", "%rsi", "%rax", "%rdx", "%rcx", "%r8", "%r9", "%r10", "%r11", "%r15");
                                                                                  ^
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%rdx' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%rax' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%rsi' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%rdi' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%xmm7' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%xmm6' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%xmm5' in 'asm'
/«PKGBUILDDIR»/lib/angelscript/source/as_callfunc_x64_gcc.cpp:144:82: error: unknown register name '%xmm4' in 'asm'
[...]

PR #2211, opened 2015-06-04
More angelscript fixes

Vincent-C:
This is a followup for PR #2195; the earlier PR contributed by @jcowgill fixes build failures on most archs mentioned in the initial bug report in #2194 except for armel. This PR fixes the armel build on Debian as well (build logs can be found [here](https://buildd.debian.org/status/package.php?p=supertuxkart&suite=unstable)). Thanks for considering!


jcowgill:
Sorry about my ARMv4 patch not working.

It looks like you actually fixed the assembly to work on ARMv4 so do you need my second patch at all? Also the commit message is wrong because you ported it to ARMv4 and haven't actually disabled anything :)

Vincent-C:
No worries! Thanks for all the work you've put in to make STK compile on the other archs that were failing.

I actually haven't tested a build without your second patch yet (it took forever to finish a single build on the Debian armel porterbox I was using), but it doesn't hurt to include it. Agreed about the silly commit message though, not sure what I was thinking. :)

auriamg:
The main issue with these pull requests is that while we integrate the angelscript sourcecode into STK for convenience, we haven't actually forked angelscript nor do we intend do. So modifications to angelscript itself should be probably submitted upstream, unless there's some bits specific to the STK build system of course

Vincent-C:
Ack, I've already submitted a patch upstream (via email to the upstream author); I just wasn't sure earlier whether or not the embedded angelscript copy in STK was intended as a convenience copy or a fork. I can just keep this as a Debian-specific patch for now. Thanks anyways!

hiker:
If the patch is useful and was submitted upstream, I think we can include it?

Vincent-C:
Just FYI, the patch was applied upstream as svn r2177.

auriamg:
Thanks! We will eventually upgrade to a newer version of angelscript and then we'll get this fix. Given that STK doesn't currently support ARM I'd say there is no urgent need to ugprade to a newer AS

auriamg:
We have just updated to a newer version of angelscript, that should hopefully address those issues. Please reopene if there are more issues

PR #2212, opened 2015-06-05
Strip ICC headers from png files to silence warning

Flakebi:
This strips ICC headers from png files to silence warnings of newer libpngs:
```
[warn   ] Irrlicht: PNG warning: iCCP: known incorrect sRGB profile
[warn   ] Irrlicht: PNG warning: iCCP: profile 'Photoshop ICC profile': 'RGB ': RGB color space not permitted on grayscale PNG
```
Additionally the removal of this headers gains a bit more space :-).

### DIFF
--- a/data/optimize_data.sh
+++ b/data/optimize_data.sh
@@ -20,8 +20,9 @@ jpegtran=true
 advdef=true
 advzip=true
 optipng=true
+convert=true
 
-#WARNING! SETTING TO TRUE MAY POSSIBLY INCREASE LOAD TIMES ON A SLOW CPU (UNTESTED) OR LEAD TO FILE NOT FOUND ERRORS WHEN RUNNING SUPERTUXKART! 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Has anyone experimented with oxipng, zopflipng and more aggressive compression/newer methods?

I'm seeing a ~12% savings against v1.4 PNG assets with oxipng without zopfli, so with zopflipng this could go up to 15% or even more.
Assuming the assets don't change too frequently, it seems like a significant improvement.

```
439 MB (460.833.829 bytes)
386 MB (405.636.168 bytes)
```

jpegoptim should also help save more bytes and it's lossless.

deveee:
No idea why it's in 10 year old bug report but you can compare with
https://github.com/supertuxkart/stk-code/blob/master/android/generate_assets.sh
that uses pngquant and converts some files to jpeg. Then if you get smaller
size and better quality then maybe it's worth it.

PR #2217, opened 2015-06-13
Fix compiler warning about shift count being bigger than size of type

tobbi:
/stk-code/src/network/network_string.hpp:170:24: warning: shift count >= width of
      type [-Wshift-count-overflow]

### DIFF
--- a/src/network/network_string.hpp
+++ b/src/network/network_string.hpp
@@ -167,7 +167,8 @@ class NetworkString
             T result = 0;
             while(a--)
             {
-                result <<= 8; // offset one byte
+                result <<= 7; // offset one byte
+                result <<= 1;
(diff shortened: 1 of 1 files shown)


auriamg:
Hi, this does not seem like an actual fix; while it does suppress the warning, it doesn't actually change the behavior of the code, so if there actually was the issue the warning is telling us about then the original issue is not actually fixed, merely hidden

tobbi:
I was told on IRC that this is the intended behaviour for this code.

leyyin:
Hi, @Flakebi also fixed this issue some time ago in his PR #1907 (commit Flakebi/stk-code@19ce80656330293a).

tobbi:
This has been fixed, presumably by the above pull request.

PR #2228, opened 2015-07-04
Fix a couple of deref after NULL check in states_screens

tobbi:
This fixes various issues reported by Coverity


auriamg:
Hi, thanks for the PR though in this case I'm pretty sure the pointer can actually never be null, so the check is incorrect. Probably the if check should be removed and replaced with assert(ribbon != NULL);

tobbi:
Here we go! I replaced the if statements with proper assets.

auriamg:
Thanks, merging!

PR #2244, opened 2015-07-21
Fix reversed rtl string part

### LINKED ISSUE #2226: STK treats quote marks inside Hebrew acronyms as a blank space
SuperTuxKart treats quote marks inside Hebrew acronyms as a blank space.

When viewing addon size information, the string %s מ"ב  (that's %s MB - string 6)  looks like this: %s מ ב . a blank space. 

Could the comix font lack a character, or STK treats that string literally ?

Flakebi:
The string was reversed twice. This fixes #2226.


auriamg:
Merged, thanks

PR #2245, opened 2015-07-22
Kart properties

Flakebi:
These changes heavily modify the property system like it was suggested on the [forum](http://forum.freegamedev.net/viewtopic.php?f=17&t=6195).
It is possible to give relative values (factors and more) for each property (that is already contained in the new system).

The format (as a regex) is: `([+-*/](%d+|x))*`
An empty value means no change, a value without sign or with a `-` replaces the previous value. If an operator prefixes the string, e.g. `*1.5` (`-` doesn't work because it's a sign). `x` will get replaced by the previous value which means `*1.5` is equal to `x*1.5`. It's also possible to combine operations:
`+1.5*20` means "add 1.5 to the previous value and then multiply by 20".
Note that it's computed sequentially from left to right and there is no operator precedence. Also whitespaces aren't allowed because they are used to split arrays.
If you want to return the square of something use `x*x` or only `*x`.

TODO:
 - [x] Remove random wheel rotation
 - [x] Remove wheel radius
 - [x] Rename TracelCm to Travel
 - [x] Maybe add field of view? (No)
 - [x] Maybe remove wheel position?
 - [x] Remove previously not applied turn-radius
 - [x] Use skidding properties from characteristics
 - [x] Getters are moved to KartProperties


src/karts/abstract_kart.cpp
```diff
@@ -79,6 +99,12 @@ void AbstractKart::reset()
 }   // reset
 
 // ----------------------------------------------------------------------------
+const AbstractCharacteristic* AbstractKart::getCharacteristic() const
+{
+    return &(*m_characteristic);
```
leyyin: Can you use http://en.cppreference.com/w/cpp/memory/unique_ptr/get please? I see the same deference in other places where you used unique_ptr.

src/karts/combined_characteristic.cpp
```diff
@@ -0,0 +1,43 @@
...
+        const SkiddingProperties *skid = characteristic->getSkiddingProperties();
+        if (skid)
+            return skid;
+    }
+    return nullptr;
+}
+
+void CombinedCharacteristic::process(CharacteristicType type, Value value, bool *is_set) const
+{
+    for (std::vector<const AbstractCharacteristic*>::const_iterator it = m_childs.cbegin();
```
leyyin: Use C++11 ranged for loops? http://en.cppreference.com/w/cpp/language/range-for

src/karts/abstract_kart.cpp
Flakebi: I can't use unique_ptr here because it wouldn't be unique then anymore. The safe way would be a weak_ptr, but you would need a shared_ptr for that and there would be overhead at runtime and in the code (because you need to cast a weak_ptr before you use it and so on...).
I decided to use a raw pointer here because it's most understandable. The only thing that has to be guaranteed for this to work, is that the pointer isn't used anymore when the parent object is deleted.

src/karts/combined_characteristic.cpp
Flakebi: I would really like to use it, but travis doesn't like it (too old compiler).

leyyin: Why does the the loop above work then? https://github.com/Flakebi/stk-code/blob/kart-properties/src/karts/combined_characteristic.cpp#L28

src/karts/abstract_kart.cpp
leyyin: I do not understand, it is unique_ptr in your code https://github.com/Flakebi/stk-code/blob/kart-properties/src%2Fkarts%2Fabstract_kart.hpp#L81. I am missing something?
What you are basically doing here is getting the value pointed by the pointer then getting the address of the value, which is basically the pointer, what .get() does for you.
And I think it's safe to call .get on an unique_ptr, as long as the other parts of the code base just use it to call other methods, and only one class manages it's lifetime.

Flakebi: Ups, sorry I misread your comment, I thought you meant I should return a unique_ptr... `get()` looks nice :smiley:

src/karts/combined_characteristic.cpp
Flakebi: Hm, interesting iirc that didn't work on a slightly newer gcc but I'll try if travis accepts this.

src/karts/xml_characteristic.cpp
```diff
@@ -0,0 +1,466 @@
...
+                                }
+                                if (!*is_set)
+                                {
+                                    Log::error("XmlCharacteristic::process", "Can't process %s",
+                                        pair[1].c_str());
+                                    value.fv->clear();
+                                    break;
+                                }
+                            }
+                            if (!found)
```
leyyin: You never set `found` in the for loop, is this statement will always be true.

src/karts/xml_characteristic.cpp
```diff
@@ -0,0 +1,466 @@
...
+                                // The searched value was not found so we have
+                                // a different format
+                                shouldReplace = true;
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        } else
```
leyyin: Newline here:
`}
else
`

hiker:
Hi,
thanks for the patch, I had (finally - my apologies) a look. I like the idea of the implementation, treating handicaps etc all the same just by stacking up characteristics is great. I see one major problem though: unless I have misread the code, whenever you access a kart-properties (now characteristics) value, the value is re-computed, and that means not only the arithmetic - the strings are parsed again etc etc., and that is not ideal performance-wise.
My suggestion would be that instead the KartPropertiesManager creates a KartProperties object 'on demand' (i.e. when a race is started). All values will only be computed once and stored in that KartPropertiesObject. Besides performance benefit, this will also mean the patch will be much smaller, since all existing code that uses KartProperties can stay the way it is and does not need to be modified to use the characterisic instead  (though if it helps we might want to replace all the individual names with an enum as an index - though that makes the patch as big again ;) ).

Additionally, several files do not follow our coding style: There are lines >100 characters (80 is the official limit, though I won't complain if this is somewhat exceeded); no "---" lines between methods, the '}' closing a function has no comment with the name of the functions, no comments explaining the functions.

Thanks a lot!
   Joerg

Flakebi:
Hi,

thanks for taking a look at my patch!
I also thought about the performance and there is a class called CachedCharacteristics, that could be used to parse the strings only once. But that part of the code is not yet used.

I'm not sure if it's possible to make the patch a lot smaller. But the changes of all files that use the KartProperties could be taken out, that's right. Maybe a KartProperties-object can hold a CachedCharacteristics-object and all the functions like `getMass` just return the result of the characteristics? On the other hand, that would be another part of the code that could be script-generated (I don't want to write all that code by hand), which is maybe a bit useless for that part because it would only forward function calls.

What I'm wondering about are the skidding properties. Currently they have their own class, but it would make sense to switch them to the characteristics-system too (if that will be used in it's current form). In my opinion, the easiest and probably best solution would be to integrate them again into the main characteristics/properties. That would have the disadvantage of longer compile times if they are changed, but I think it's a lot easier to use that structure.

Sorry for the missing comments and so on, I will look over that and @leyyin's comments. :-)

Cheers,
Seebi

Flakebi:
I finally had some spare time again, so the coding style should be better now and the characteristics won't be recomputed every time.

src/guiengine/widgets/kart_stats_widget.cpp
```diff
@@ -81,15 +79,18 @@ KartStatsWidget::KartStatsWidget(core::recti area, const int player_id,
         m_children.push_back(skill_bar);
     }
 
-    m_skills[SKILL_MASS]->setValue((int)(props->getMass()/5));
+    // Scale the values so they look better
+    m_skills[SKILL_MASS]->setValue((int)
+            ((props->getCombinedCharacteristic()->getMass() - 20) / 4));
```
auriamg: It may be good to leave a little comment explaining where those numbers come from? 20 and 4 (and on lines below, 15 and 6, 30 and 20). I know they weren't detailed before, but if we're going to improve this could we might as well do it all the way

src/karts/cached_characteristic.cpp
```diff
@@ -0,0 +1,203 @@
...
+                break;
+            case TYPE_INTERPOLATION_ARRAY:
+                delete static_cast<InterpolationArray*>(v.content);
+                break;
+            case TYPE_BOOL:
+                delete static_cast<bool*>(v.content);
+                break;
+            }
+            v.content = nullptr;
+        }
```
auriamg: This works well, though I'm wondering, couldn't the destructor of Value handle that? Seems like it would belong there ideally

src/karts/kart.cpp
```diff
@@ -2107,11 +2137,11 @@ void Kart::updatePhysics(float dt)
     // Only apply if near ground instead of purely based on speed avoiding
     // the "parachute on top" look.
     const Vec3 &v = m_body->getLinearVelocity();
-    if(/*isNearGround() &&*/ v.getY() < - m_kart_properties->getSuspensionTravel()*60)
+    if(/*isNearGround() &&*/ v.getY() < - m_kart_properties->getSuspensionTravel() * 0.01f * 60)
```
auriamg: Same thing, could we leave some comment explaining where this 0.01 * 60 value comes from? Same thing for a few lines below

src/karts/skidding.cpp
```diff
@@ -47,8 +47,7 @@ Skidding::Skidding(Kart *kart, const SkiddingProperties *sp)
     m_actual_curve->setVisible(false);
 #endif
     m_kart = kart;
-    copyFrom(sp);
-    m_skid_reduce_turn_delta = m_skid_reduce_turn_max - m_skid_reduce_turn_min;
+    m_skid_reduce_turn_delta = m_kart->getKartProperties()->getSkidReduceTurnMax() - m_kart->getKartProperties()->getSkidReduceTurnMin();
```
auriamg: General remark, that's not really bad but a stylistic remark. Many of these lines are quite long and repeat "m_kart->getKartProperties()" over and over. It may be better to create a "kart_properties" variable at the start, calling "kart_properties = m_kart->getKartProperties()" only once and then avoid the repeated calls

auriamg:
So I reviewed the code and overall it seems pretty good :) 

I must say I'm not a big fan of unions unless strictly necessary, it's unsafe, risky, adds memory management issues. Though in this case it might still be the best alternative. I guess in general I would have preferred the union type to be a little more self-contained, but maybe that can't be helped

src/karts/kart.cpp
hiker: That actually is commented just 
just before the shown snippet ;)

I expect a conflict there, originally the suspension travel was in cm, and I changed the units a while ago to make it all consistently use m.

auriamg:
Okay I tried the branch and unfortunately it crashes for me :

```
>	supertuxkart.exe!AbstractCharacteristic::getSkidEnabled() Line 1719	C++
 	supertuxkart.exe!KartProperties::getSkidEnabled() Line 1143	C++
 	supertuxkart.exe!Kart::loadData(RaceManager::KartType type, bool is_animated_model) Line 2448	C++
 	supertuxkart.exe!Kart::init(RaceManager::KartType type) Line 227	C++
 	supertuxkart.exe!World::createKart(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & kart_ident, int index, int local_player_id, int global_player_id, RaceManager::KartType kart_type, PerPlayerDifficulty difficulty) Line 307	C++
 	supertuxkart.exe!World::init() Line 190	C++
 	supertuxkart.exe!WorldWithRank::init() Line 33	C++
 	supertuxkart.exe!LinearWorld::init() Line 61	C++
 	supertuxkart.exe!RaceManager::startNextRace() Line 494	C++
 	supertuxkart.exe!RaceManager::startNew(bool from_overworld) Line 409	C++
```

I have no more information at this time

Flakebi:
Sorry for that crash. It didn't occur with my normal build (that was using gcc). I was able to reproduce it with clang (not sure if that's really related because I didn't test more). Now it works for me with both compilers.

auriamg:
Merged, thanks!

PR #2252, opened 2015-07-26
Fix crash when loading story mode

tobbi:
This fixes a crash when trying to load story mode with latest master.


Flakebi:
The crash is caused by the missing `trigger-type` attribute for action-triggers. It happens for the overworld and the tutorial.

auriamg:
Thanks for the fix

PR #2253, opened 2015-07-26
Mac compilation fixes (fix calls to std:abs)

tobbi:


auriamg:
Hi, I've applied a slightly different version of your fix : I just use <cmath> on all systems, that should work and avois the #if checks

PR #2271, opened 2015-08-08
Allow enabeling Network Multiplayer through CMake flags

konstin:
This makes working on the network mode with git much easier


deveee:
If you are working on network mode and this modification makes your life easier, then why not. This option is disabled by default anyway.

hasufell:
Can someone explain to me what this does? I enabled and disabled it and I see no difference.

konstin:
First of all, network multiplayer doesn't yet work and is only meant for developers who understand what they're doing. But if you're really willing to do that anyway, here's how:
 * go to your build directory
 * run `cmake .. -DENABLE_NETWORK_MULTIPLAYER` and then compile, e.g. with `make -j4`
 * click on the online/network entry in the main menu
 * have fun with an undocumented and broken part of STK...

hasufell:
I'm just packaging this, so I guess I'll leave this option out.

konstin:
Definitely. A simple plain `cmake ..` should be pretty fine. Oh, and please don't try to use system versions of libraries shipped with STK because they're normally modified. See http://supertuxkart.sourceforge.net/Build_STK_on_Linux for the details.

hasufell:
I'm already aware of that, thanks.

PR #2272, opened 2015-08-08
Added documentation, refactored large code pieces and implemented a consitent error handling for the STUN requests

konstin:
This PR mainly consists of refactoring and documenting, but it also ensure that a new STUN request is send and an error message is printed every time a response contained invalid data.

### DIFF
--- a/src/network/protocols/get_public_address.cpp
+++ b/src/network/protocols/get_public_address.cpp
@@ -29,6 +29,8 @@
 
 #include <assert.h>
 
+#include <string>
+
 #ifdef __MINGW32__
 #  undef _WIN32_WINNT
 #  define _WIN32_WINNT 0x501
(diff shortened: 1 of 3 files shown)


src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
     m_state = NOTHING_DONE;
 }
 
-void GetPublicAddress::asynchronousUpdate()
+/**
+ * Gets the response from the STUN server, checks it for its validity and
+ * then parses the answer into address and port
+ * \return "" if the adress could be parsed or an error message
```
leyyin: Typo 'adress'

konstin: damn, I always do that mistake.

src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
...
-        m_listener->requestTerminate(this);
+        struct sockaddr_in* current_interface = (struct sockaddr_in*)(p->ai_addr);
+
+        m_stun_server_ip = ntohl(current_interface->sin_addr.s_addr);
+        m_transaction_host = new STKHost();
+        m_transaction_host->setupClient(1,1,0,0);
+        m_transaction_host->sendRawPacket(bytes, 20, TransportAddress(m_stun_server_ip, 3478));
+        m_state = TEST_SENT;
+
+        freeaddrinfo(res); // free the linked list
```
leyyin: Comment seems redundant for me. The same as the one below.

konstin: Indeed, though the comment wasn't from me. I'll fix this with the next commit

src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
...
+    // The address and the port are known, so the connection can be closed
+    m_state = EXITING;
+    // terminate the protocol
+    m_listener->requestTerminate(this);
+
+    return "";
+}
+
+
+/** Creates a request and sends it to a random STUN server randomly slected
```
leyyin: Typo 'slected'

src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
...
+    if (message_size == 0)
+        return "STUN response does not contain any information.";
+    if (message_size < 4) // cannot even read the size
+        return "STUN response is too short.";
+
+
+    // Those are the port and the address to be detected
+    uint16_t port;
+    uint32_t address;
+    while(true)
```
leyyin: while (true)

leyyin:
Btw, In the this file there is a function that seeds the number generator and returns a pseudo random number: https://github.com/konstin/stk-code/blob/dev/src/network/protocols/get_public_address.cpp#L49.
I searched the whole project and the only other place where srand() is called it is in main.cpp (where it should be). This function seems pointless to me, we can simply use rand().

src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
...
+
+    // resolve the name into an IP address
+    struct addrinfo hints, *res, *p;
+    int status;
+
+    memset(&hints, 0, sizeof hints);
+    hints.ai_family = AF_UNSPEC; // AF_INET or AF_INET6 to force version
+    hints.ai_socktype = SOCK_STREAM;
+
+    if ((status = getaddrinfo(stun_servers[rand_result].c_str(), NULL, &hints, &res)) != 0) {
```
leyyin: '{' on new line

src/network/protocols/get_public_address.cpp
```diff
@@ -68,188 +70,188 @@ void GetPublicAddress::setup()
...
+        attributes = attributes + 4 + size;
+        message_size -= 4 + size;
+        if (message_size == 0)
+            return "STUN response is invalid.";
+        if (message_size < 4) // cannot even read the size
+            return "STUN response is invalid.";
+    }
+
+    // finished parsing, we know our public transport address
+    Log::debug("GetPublicAddress", "The public address has been found: %i.%i.%i.%i:%i", address>>24&0xff, address>>16&0xff, address>>8&0xff, address&0xff, port);
```
leyyin: This is over the 80 char limit on a line.

konstin:
After all those changes I'ld say that I've rather rewritten the whole `GetPublicAddress` class.

auriamg:
The travis builf fails, this should be fixed

Excerpt:
```
/home/travis/build/supertuxkart/stk-code/src/network/protocols/get_public_address.hpp:43:59: error: a brace-enclosed initializer is not allowed here before ‘{’ token
```

konstin:
The problem is that it uses a C++11 initialiser. This can propably be solved by adding a -std=c++11 flag or by upgrading gcc.

auriamg:
The travis build system already uses -std=gnu++0x, so it should be okay on that front (AFAIK gnu++11 is like c++11, just more compatible)

At this time whatever version is installed on travis sort of serves us as minimum-requirements gcc version, so if your code doesn't build on travis this probably means it uses a feature that was only introduced in a recent gcc?

konstin:
I fix the compilation. The Problem was a C++11 feature that allows initializing array in headers files, which was introduced in 4.7, while travis uses 4.6.

btw #2271 is good for testing the changes ;)

Issue #2232, opened 2015-07-06
Feature request: Looking back while start countdown is running


lightonflux:
At the moment it is not possible to look back while the start countdown is counting. This would be very handy in multiplayer so you know if your arch enemy is behind you, so you can prepare to be rammed or worse: overtaken!

BTW: it is possible to accel, brake, steer left and right, so why not turn your head?

temeliak:
Hi! 

This is the first time I try to contribute to a project, so please excuse me if I do something wrong. 
So I found how to allow the player to look backward during the countdown : how can I submit the modification?

Thank you!

deveee:
Fixed in current git

PR #2280, opened 2015-08-11
Offers the possibility to look backward during the start phase. #2232

temeliak:
https://github.com/supertuxkart/stk-code/issues/2232

Moving the camera "manager" before the penalty time check allows the player to look backward during the countdown.

### DIFF
--- a/src/karts/controller/player_controller.cpp
+++ b/src/karts/controller/player_controller.cpp
@@ -326,8 +326,8 @@ void PlayerController::update(float dt)
     // if automatic reverse camera is active
     if (m_camera->getMode() != Camera::CM_FINAL)
     {
-        if (m_controls->m_look_back || (UserConfigParams::m_reverse_look_threshold>0 &&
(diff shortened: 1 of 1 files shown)


src/karts/controller/player_controller.cpp
```diff
@@ -321,6 +321,24 @@ void PlayerController::update(float dt)
     if (!history->replayHistory())
         steer(dt, m_steer_val);
 
+
+    // look backward when the player requests or
+    // if automatic reverse camera is active
+    if (m_camera->getMode() != Camera::CM_FINAL)
+    {
+        if (m_controls->m_look_back || (UserConfigParams::m_reverse_look_threshold>0 &&
+            m_kart->getSpeed()<-UserConfigParams::m_reverse_look_threshold))
```
leyyin: Hi, can we make it so that there is space around the '<'.
```  m_kart->getSpeed() < -UserConfigParams::m_reverse_look_threshold```.

deveee:
I don't see any problem in having ability to look backwards before start. And the modification seems to be safe.

Issue #2182, opened 2015-05-07
Add sfx for nitro


hiker:
As mentioned in 
http://forum.freegamedev.net/viewtopic.php?f=17&t=6289&sid=4b0c120d0f50b6c60d33c8ca32133341&start=25#p64138

http://opengameart.org/content/racing-speed-boost-sound was recommended.

auriamg:
Nitro now has sound effects, this ticket is outdated

PR #2283, opened 2015-08-12
Added SFX to nitro

temeliak:
https://github.com/supertuxkart/stk-code/issues/2182
I added some SFX to nitro. Beside the changes in the code, some changes have to be done in stk_assets :
- adding http://www.bigsoundbank.com/dl.php?ext=ogg&id=0851 in stk-assets/sfx/ and renaming it nitro.ogg
- adding the following line to stk-assets/sfx/sfx.xml :   
 `<sfx filename="nitro.ogg"            volume="3.0" positional="true" rolloff="0.5" />`

The result can be seen here : https://youtu.be/fU1IYIvejnw?t=46s

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -164,6 +164,7 @@ Kart::Kart (const std::string& ident, unsigned int world_kart_id,
     m_boing_sound   = SFXManager::get()->createSoundSource( "boing" );
     m_goo_sound     = SFXManager::get()->createSoundSource( "goo"   );
     m_skid_sound    = SFXManager::get()->createSoundSource( "skid"  );
+    m_nitro_sound   = SFXManager::get()->createSoundSource( "nitro" );
     m_terrain_sound          = NULL;
     m_previous_terrain_sound = NULL;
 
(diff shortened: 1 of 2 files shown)


src/karts/kart.cpp
```diff
@@ -1710,18 +1715,25 @@ void Kart::updateNitro(float dt)
     bool increase_speed = (m_controls.m_nitro && isOnGround());
     if (!increase_speed && m_min_nitro_time <= 0.0f)
     {
+        if(m_nitro_sound->getStatus()==SFXBase::SFX_PLAYING)
```
leyyin: Hi, can you make so that it looks like:
```if (m_nitro_sound->getStatus() == SFXBase::SFX_PLAYING)```

The same with the other ifs, as it makes the code more readable. Thanks.

auriamg:
Hi, the code seems good, though we would need licensing information regarding the sound effect, so that we can give proper credit in sfx/License.txt

However, this sound seems to be a chainsaw sound, not sure it really fits for nitro? It's also too long and starts too slowly, people usually use nitro in short bursts so we need a sound that sounds good in quick bursts. Also, it may be nice to have a little fade-out when releasing the nitro button?

auriamg:
Hi, I have merged your pull request, thanks for your contribution!

If anyone wants to add a nice little fadeout, help is still welcome ;)

PR #2284, opened 2015-08-12
Check size of window; don't assume it equals requested size

torso:
Some window managers don't honour window size requests. If the window is bigger than requested, only the top left corner of cutscenes will fade.

### DIFF
--- a/src/states_screens/cutscene_gui.cpp
+++ b/src/states_screens/cutscene_gui.cpp
@@ -16,10 +16,10 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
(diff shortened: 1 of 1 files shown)


deveee:
There was long discussion about it in one bug report. Width and height config params can't be changed because they are remembered in xml file and then used in next run, in settings screen etc.

But we are using real renderer size everywhere, where it's needed. It was m_actual_width or similar variable.

If there is a place where config params are used and they shouldn't, it should be simply changed to m_actual_width/height

torso:
I'm glad to hear that. It seemed like user config was updated to reflect screen size in other cases and used on purpose for these things. I've changed it to use irr_driver->getActualScreenSize() in the affected case.

auriamg:
Thanks for the fix, I have merged the pull request

Issue #1552, opened 2014-09-21
Kart particles effects are displayed multiple times with some settings


deveee:
Currently skidding particles are barely visible. I changed rate and life time values in in xml file to higher values. Now I see it twice:

![soccer_field-2014 09 21_10 52 50](https://cloud.githubusercontent.com/assets/5108558/4347843/61325398-416d-11e4-9aaf-38f624dafbdb.png)

Flakebi:
Even if the game runs at high framerates, I see this issue while I'm skidding. It seems like the particles use the direction that the graphical model of the kart is facing, but the real (physical) direction is another because the kart is turned while skidding. So the line of particles that gets created in one frame is far away from the one created in the last frame, because it's created with the wrong direction. That creates the effect that the nitro stream is split into parts.

Flakebi:
Now I'm sure that the divided nitro effect is caused by rotating the visual kart while skidding. The nitro box-emitter is added to the kart node and creates particles every frame, but in the wrong direction (as stated in my previous comment).

I didn't find a way to rotate a box-emitter in irrlicht so I think the only way to fix it is to create another node for the kart that has the same rotation as the physical kart and add the nitro emitters there.

hiker:
Removed invalid, since it appears to be not necessarily related to FPS rate.

deveee:
Kart particles work mostly in the same way as in 0.8.1 version. Simply lifetime values can't be too high. I think that ```lifetime max="50"``` is safe.

I only changed lifetime to:
```
    <lifetime  min="30"
               max="50"          />
```
and it looks fine for me. Though it is still noticeable with high fps values during skidding.

hiker:
The implementation is totally different from 0.8.1. On my hd3000 I see the skid particles too far behind the kart, but they are at the right place when the kart is standing still (with a little patch to always show skid particles). I also see two rows. On the other hand, not really a release blocker, since it works for 'most' cards, so postponed.

deveee:
I mean that I had similar issues in STK 0.8.1 when I was experimenting with different values. I didn't look into code.

Again in STK 0.8.1 when you had very low fps and velocity parameter was different than 0, particles were displayed too far or too high. Personally I workarounded it by setting velocity to 0. But I didn't observe it in current git.

I can reproduce it with intel hd4000 and nvidia 635m. Rather common, but low-fps graphics cards.

I suggest to workaround it by setting lifetime min=30 and max=50. It doesn't look that bad.

[...]

PR #2285, opened 2015-08-13
Fix kart particles #1552

Elderme:
https://github.com/supertuxkart/stk-code/issues/1552

There is no more visual artifacts with nitro and skidding particles whatever the framerate.


deveee:
You are great, nitro particles finally work perfectly for me :)

I will wait for hiker because he was looking at this code some time ago and he knows it better than me. But it works for me with intel and nvidia graphics cards and very different framerates.

Issue #2110, opened 2015-04-24
[Chinese translation] Text rendered out of area in serveral SuperTuxKart 0.9 prompts


Vdragon:
As the following screenshot:
![supertuxkart_507](https://cloud.githubusercontent.com/assets/1192163/7316590/cfb73522-eaaa-11e4-8745-99d99cdfce04.png)
![image](https://cloud.githubusercontent.com/assets/1192163/7316702/cdf8e5e0-eaab-11e4-8880-bbd3417bbae6.png)

#### OS
Ubuntu 14.04 x86(32-bit)

#### Locale
zh_TW.UTF-8

#### STK source
<http://sourceforge.net/projects/supertuxkart/files/SuperTuxKart/0.9/supertuxkart-0.9-linux.tar.xz/download>

#### Graphics
NVIDIA GTS-250 using proprietary driver 331.113 from Ubuntu 14.04 software source

#### Desktop Environment
KDE from KDE Backports PPA

auriamg:
This happens because STK will only wrap line on whitespaces, but in Chinese there appear to be very long words without spaces, like 可能會連線到伺服器下載附加元件並提醒您更新資訊。我們也會收集匿名的硬體統計資訊以協助 . Not sure what can be done about this? I am not every sure how to determine where it is legal to wrap. Maybe soft-line-break characters could be added to indicate where it's okay to split words?

auriamg:
Will need feedback from someone who speaks chinese to know how to fix this properly

Benau:
I think it's a problems for the translation po files..

according to the po file, for english the long part are split already but not the chinese one:

#: src/main.cpp:1201
msgid ""
"SuperTuxKart may connect to a server to download add-ons and notify you of "
"updates. We also collect anonymous hardware statistics to help with the "
"development of STK. Please read our privacy policy at "
"http://privacy.supertuxkart.net. Would you like this feature to be enabled? "
"(To change this setting at a later time, go to options, select tab 'User "
"Interface', and edit \"Connect to the Internet\" and \"Send anonymous HW "
"statistics\")."

msgstr "SuperTuxKart 可能會連線到伺服器........" (one line only)

So just need to update the po files (both simplifed and traditional).

I can do that later.

Benau:
Also it seems that when you use a bigger resolution, the chinese words will be placed correctly.

But the english words will be seperated itself in small resolution.

Benau:
[...]

PR #2301, opened 2015-08-29
Fix the ugly hack for displaying Chinese in #2110, and then fix #2297

### LINKED ISSUE #2297: Fix the ugly hack for Chinese display
Hi!

I wrote a patch for stk/irrlicht to make it be able wrap line for CJK automatically in #2110 , but there're lots problem discovered later. 

Like space between chinese and english get ignored, incorrect linebreak, wrong wrapping position for english with chinese. 

Luckily, today I fully solved it by rewriting the print part of statictext, I posted the patch to irrlicht devs before, but getting no response, so I post here to hear comments.

Now stk/irrlicht can handle most utf breaking rules for CJK (can add more in utfwrapping.h) , like no opening brackets at ending line or punctuation mark at starting line. 

Here's screenshots:

As u can see, even the most complex combination of english, chinese and brackets, stk/irrlicht can print without wasting too much space *i copy the translation from another msgstr because i don't know how to print that box

As shown in the upper middle box, there's a unneeded newline for chinese, which fixed with the new patch.

Now some questions, is the code good enough? I don't know if I should add a extra .h for defination of 
brackets and punctuation. Do u think this should be included in irrlicht trunk too?

Patch: http://pastebin.com/zU7BQL9G

Benau:
Result:

Image
Before any patch
Image
Ugly hack and the final one
*I replace the text in po file as I don't know how to generate the box.

### DIFF
--- a/lib/irrlicht/include/utfwrapping.h
+++ b/lib/irrlicht/include/utfwrapping.h
@@ -1,0 +1,104 @@
+// Copyright (C) 2015 Ben Au
+// This file is part of the "Irrlicht Engine".
+// For conditions of distribution and use, see copyright notice in irrlicht.h
+
+namespace irr
+{
+namespace gui
+{
+
+//Here a list of characters that don't start or end a line for chinese/japanese/korean
+//Only commonly use and full width characters are included
+//You should use full width characters when writing CJK, like using "。"instead of a "."
+//You can add more characters if needed
+//For full list please visit http://webapp.docx4java.org/OnlineDemo/ecma376/WordML/kinsoku.html
+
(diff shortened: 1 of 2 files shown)


auriamg:
Thanks for the fix! We are currently close to releasing a bugfix release, this change is probably a little too big and too late to go in this bugfix release, we'll see depending on our free time - if it doesn't go in this upcoming release we will take a look for the one right after

Benau:
ok!

auriamg:
Okay turns out I had enough time to test this PR, seems to work well, thanks

PR #2307, opened 2015-09-03
Fix typo and and add quotation

Benau:
Correct comment of namespace

The quotation added is for Chinese Simplified, which I discovered usage in translation work, to display better.

Sorry for extra pull

### DIFF
--- a/lib/irrlicht/include/utfwrapping.h
+++ b/lib/irrlicht/include/utfwrapping.h
@@ -17,6 +17,10 @@ bool UtfNoStarting (wchar_t c)
 {
 	switch (c)
 		{
+			case 8217:	//’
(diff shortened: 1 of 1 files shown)


Benau:
By the way, I think you may need a update for font pngs of both chinese version, as some "new" chinese words are used/changed.

Is stk only pre-load chinese that's inside po file?

But let me finish translation for chinese simpfiled first, when you see both 100% then it should be ok.

auriamg:
Thanks for the fix

And yes, we regenerate the chinese font before each release, so don't worry about that, by the time we release all characters will be in there! The reason is that we can't make fonts containing all chinese characters, that would overload the GPU

PR #2317, opened 2015-09-18
Fix SuperTux difficulty button state not updating correctly

MiroslavR:
If the race setup screen is entered while the SuperTux difficulty is unavailable (locked), the SuperTux difficulty button will retain its locked badge and inactive state forever -- that is even if said difficulty becomes available (e.g. by changing the current user).


auriamg:
Thanks for the fix!

Issue #2330, opened 2015-09-30
STK rc failed on clear xp


Benau:
After further investigation, seems the the openal32.dll bundled in mingw zip, requires msvcr90.dll which is not found in my system, I replace openal32.dll with the one provided by oalinst.zip and it runs.

Even so, it crashed when I tried to play, but it's a less important issues as it's a Vbox xp. But overall the gui display as properly as in linux/ normal windows.

Should we consider use the openal32.dll provided by oalinst.zip instead?

deveee:
It's not possible to run STK on Virtualbox because of this issue:
https://github.com/supertuxkart/stk-code/issues/2326

The openal32.dll library is the same one, which is used in our VS package.
Indeed it would be nice to not depend on msvc libraries, even in old
versions. I will look what can be done.

Benau:
Thank you for a quick reply!

deveee:
After quick look, even libgcc depends on msvcrt.dll ;) But only OpenAL
depends on msvcr90.dll:
MSVCR90.dll =>
/cygdrive/c/Windows/WinSxS/x86_microsoft.vc90.crt_1fc8b3b9a1e18e3b_9.0.30729.6161_none_50934f2ebcb7eb57/MSVCR90.dll
(0x72110000)

And indeed version which I have installed in system32 doesn't depend on
msvcr90.dll. I think that we should use it.

But I remember that OpenAL32 installer contains two libraries and both are
needed. Maybe I just were using newer installer from their website, and our
installer is in older version? I don't have any tool to unpack it atm.

Benau:
![ms](https://cloud.githubusercontent.com/assets/3252341/10187400/e67b16a4-678a-11e5-8598-2137d6378507.png)

I used the latest oalinst.zip from offical website, do u find the screenshot useful?

deveee:
Okay, I updated dependencies for mingw. Now it doesn't depend on msvcr90.dll. You can check it if you want:
https://github.com/supertuxkart/dependencies/tree/master/windows-mingw

Current OpenAL version needs both files: OpenAL32.dll and wrap_oal.dll. If you have only the first one, there is no sound in game. Actually I noticed that Blender uses only wrap_oal.dll library, but this would be a bit confusing.

I checked it on few different windows installations and it works for me.

PR #2331, opened 2015-10-01
stk-code: Rework Windows installer to build directly from source tree

MattDevo:


MattDevo:
added a slightly more paranoid sanity check on the installation dir per leyyin's guidance

deveee:
Note that in case of removing OpenAL installer, we need also update dependencies for Visual Studio. Current openal32.dll library from our dependencies, depends on msvcr90.dll library (which we don't want at all).

Newest OpenAL needs two files to make it working: openal32.dll and wrap_oal.dll.

More info in this ticket:
https://github.com/supertuxkart/stk-code/issues/2330

MattDevo:
@deveee I agree, the non-mingw Windows dependencies need to be updated as well.  I'm happy to modify the Windows build stuff to use the dependencies repo instead of a packaged distribution if that's the preferred method going forward

Benau:
Hmm, so what's the idea for this PR, which is for 0.9.1?

qwertychouskie:
What is the status of this?  I know we had some Installer work since this PR was made, is this still of any use?

PR #2354, opened 2015-10-20
Fix compilation with system angelscript

hasufell:
Upstream names the shared library angelscript_s.
In addition, add more common PATH_SUFFIXES.


hasufell:
The situation of angelscript is a bit difficult. SONAME and library names are a bit weird. I've contacted the developer for both things. If we are lucky, we may get
* proper SONAMEs
* same library name for shared/static
* pkgconfig file

PR #2356, opened 2015-10-20
Remove trailing whitespaces

jubalh:


auriamg:
One issue I have with this PR is that it doesn't distinguish trailing whitespace from indented empty lines - your PR deindents some empty lines. This is minor but having empty lines pre-indented is IMO not an issue and means it's quicker if you want to insert code there

deveee:
Also removing trailing spaces in libraries doesn't have sense. It
makes harder to apply upstream patches (you won't be able to just
cherry-pick a commit).

Flakebi:
There is a script in the `tools/` directory (`whitespaces.py`). That should be used as it should fit all these needs.

hiker:
We won't apply this patch for the mentioned reasons. Could I strongly recommend to talk to us devs first when trying to contribute? This way we can help deciding what is useful etc.

Thanks a lot for your help!
  Joerg

PR #2365, opened 2015-10-30
Add default precision qualifier in fragment shaders

Elderme:
I tried to fix shader compilation with some drivers:
http://forum.freegamedev.net/viewtopic.php?f=17&t=6600

I don't know if it will work since I can't reproduce the issue described in this topic. According to GLSL specification, precision qualifer are useless with OpenGL (they only have an effect with OpenGL ES), so this PR should not break anything, at worse it is useless.


auriamg:
Thanks a lot! Let's hope it fixes the issue

PR #2367, opened 2015-11-01
Fix a couple of memory leaks.

CruzR:
There are a few other things in there that make me wonder whether you use a custom operator new or Boehm GC or something, but I could not find any information. For example, pretty much all Dialog classes are just dynamically allocated but never deleted anywhere. Any ideas?

### DIFF
--- a/src/physics/physical_object.cpp
+++ b/src/physics/physical_object.cpp
@@ -18,6 +18,7 @@
 
 #include "physics/physical_object.hpp"
 
+#include <memory>
 #include <string>
 #include <vector>
 
@@ -325,7 +326,6 @@ void PhysicalObject::init()
     case MP_EXACT:
     {
         extend.setY(0);
-        TriangleMesh* triangle_mesh = new TriangleMesh();
 
         // In case of readonly materials we have to get the material from
         // the mesh, otherwise from the node. This is esp. important for
@@ -361,6 +361,8 @@ void PhysicalObject::init()
                 return;
         }   // switch node->getType()
 
+        std::auto_ptr<TriangleMesh> triangle_mesh(new TriangleMesh());
+
         for(unsigned int i=0; i<mesh->getMeshBufferCount(); i++)
         {
             scene::IMeshBuffer *mb = mesh->getMeshBuffer(i);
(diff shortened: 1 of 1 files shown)


src/network/stk_host.cpp
```diff
@@ -214,18 +214,18 @@ void STKHost::setupServer(uint32_t address, uint16_t port, int peer_count,
                             int channel_limit, uint32_t max_incoming_bandwidth,
                             uint32_t max_outgoing_bandwidth)
 {
-    ENetAddress* addr = (ENetAddress*)(malloc(sizeof(ENetAddress)));
-    addr->host = address;
-    addr->port = port;
+    ENetAddress addr;
+    addr.host = address;
+    addr.port = port;
```
hiker: Please do not do any changes to the networking code, it is currently being completely rewritten in a different branch, so applying this would cause a lot of unnecessary conflicts later.

src/network/stk_host.cpp
```diff
@@ -375,6 +375,7 @@ uint8_t* STKHost::receiveRawPacket(const TransportAddress& sender,
             TransportAddress a(m_host->address);
             Log::verbose("STKHost", "No answer from the server on %s",
                          a.toString().c_str());
+            delete[] buffer;
```
hiker: I've applied this to the networking branch (6946306c4c29f23998faf4b2807cc22b629fb057). Thanks

hiker:
Any gui element is just created once, and we never bothered with freeing them, since they are potentially needed till the very end anyway.

I am not to keen to have an automatic garbage collector, since it could run at the wrong time causing more jitter during a race.

src/network/stk_host.cpp
CruzR: Alright, I've removed the changes to netcode from this commit.

auriamg:
The fix seems clean, thanks. Note that as far as I can tell this is a rare leak that could only occur when reading a malformed data file (normal execution paths would not leak so it wasn't a huge issue), but doesn't hurt to fix it anyway

CruzR:
Yeah, to be completely honest, I did not find these myself, clang's ``scan-build`` tool found them.

I've got a couple of other potential errors that also only occur in rare occasions, including calling methods on pointers that might be ``NULL``, but I haven't looked at them in detail yet and they also seem to be on error paths.

PR #2375, opened 2015-11-08
Add matrix configuration to travis file

leyyin:


leyyin:
Unfortunately they moved to a newer infrastructure (https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future). So no more access to 32 threads :).
I changed the code so that it uses the new container based infrastructure (this means no more sudo) and using the APT addon https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon.
I did benchmark different scenarios and I could not see any significant speed improvement.

PR #2379, opened 2015-11-13
Color the text base on current skin

Benau:
I tried my best to make STK define less color as possible, without changing the look comparing to current one (Peach skin) very much. Now there are 3 types of color as you can see. Feel free to tell if I missed any hard-coded color for text (except for white/black).

Now this is the Ocean skin font color I used:

I use purple to make the text more readable when it's above the sky. For the rest should be OK as long as it has a black border.

Regarding skin, I think it's good idea to make 2 more (red and green  color??) which are base on peach and ocean, so the (choosing) skins in STK are more useful.

4 skins which are not too less or more also cover most people prefered color (keep in mind that there can be more color mixed with one skin, like I mixed "Ocean skin" with purple). I can do that, if you don't mind me just GIMP-color the current skins :)


data/skins/Ocean.stkskin
```diff
@@ -257,4 +257,9 @@ when the border that intersect at this corner are enabled.
 <!-- Rating star image -->
 <element type="rating" state="neutral" image="ocean/rating_star.png" />
 
+<!-- Font color -->
+<color type="top" state="font" a="255" r="100" g="200" b="220" />
+<color type="bottom" state="font" a="255" r="80" g="80" b="220" />
+<color type="normal" state="font" a="255" r="40" g="40" b="180" />
+
```
auriamg: This is reversed :) type should be "font" and state top/bottom/normal

Benau:
Two new skin (don't know if it's a good name):



Don't know if the best color is used too...

I also fix some alignment issue with some selection boxes.

PR #2380, opened 2015-11-17
Fixes few warnings at compilation

nado:
-Wunused-variable for gcc and clang, and some only for clang.


deveee:
This rather doesn't have a sense:
```
if ((!sscanf(s, "reverse: %c", &r)) != 1)
```
sscanf returns integer value. It should return "1" if success and "0" or "-1" if failure. It will work in most cases because we get ```if (!1 != 1)``` if success. But it should be rather:
```
if (sscanf(s, "reverse: %c", &r) == 1)
```
or at least
```
if (!(sscanf(s, "reverse: %c", &r) != 1))
```

nado:
Fixed.

PR #2391, opened 2015-12-04
Switch to new container based travis

nado:
Its a one step further than https://github.com/supertuxkart/stk-code/pull/2375 I didnt know if I should have PR to @leyyin or here so I put it here directly.

The `sudo: false` is to force the switch to the new container based infrastructure from travis, in the near future, there will be an automatic cache for apt dependencies, so it might interesting to speed up a bit the build time.

Also, the containers boot up faster than legacy infrastructure vm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,44 +3,43 @@
 # Configuration manual:
 # http://docs.travis-ci.com/user/build-configuration/
 #
+sudo: false
+
 language: cpp
 compiler:
   - gcc
-# - clang
+  - clang
 #branches:
 #  only:
 #    - master
-before_install:
-  # Update repos
-  - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


nado:
After comparing build times I also updated the value of THREADS to 16 as seen in @leyyin PR

leyyin:
@nado You should not use 16 threads for the travis containers (last time I checked they only have 2 threads).
Also the travis containers are up to 3 times slower than the old infrastructure.
https://travis-ci.org/nado/stk-code/builds/94970460
vs
https://travis-ci.org/leyyin/stk-code/builds/94986730

nado:
Wow ._. Indeed, I did not check the detailed build, 3 minutes only is kinda amazing. I’ll close this one then.

PR #2394, opened 2015-12-12
Purpel.stkskin

qaugustijn:


Benau:
First, u miss purpel skin data (png file), also no font color define in your stkskin, which is required for next stk version

qaugustijn:
How to upload a png file to this repository of Github?

Benau:
U cannot file to github, u need a local copy of your fork.

Than you can push

Also it would be good if your new skin is drawn from scratch, so some skin(ruby?) can be replaced, which it's gimp-color

qaugustijn:
I have created a folder with the png files of the new skin and I have tested in supertuxkart.

Benau:
You should test with current (git) version of supertuxkart for future compatibility.

deveee:
This pull request doesn't contain "purpel" directory, which as I see you use in your skin. Please, prepare proper pull-request with all needed files, so that it could be useful for others.

If you think that your skin is really nice, and you don't know how to prepare pull-request, then just upload it somewhere as a zip file.

PR #2396, opened 2015-12-12
Update stk_config.xml

qaugustijn:


Benau:
Notice that it meaningless to have 50 max karts, first slow pc will make the game unplayable, also there may not be enough starting position

Btw, 3 strikes battle won't support 50 karts

deveee:
You just modify one parameter, without any description why do you change it and without any argument why this value is the best one. I would like to close this pull request doing the same ;)

The 50 karts in race is certainly too much:
- it can slowdown the game on slow PC
- gameplay has been tested up to 20 karts and it can be just one big mess with more karts
- as Benau said, 3 strikes battle won't support 50 karts
- did you test all official tracks to make sure that for example 45th kart doesn't stay on a gift or on a banana before start?

And still similar changes should be discussed before.

PR #2405, opened 2015-12-18
Some few C++ code corrections and slighty better alternative OS support

devnexen:
[patch-supertuxkart.txt](https://github.com/supertuxkart/stk-code/files/66987/patch-supertuxkart.txt)
Hi,

I was trying to compile this version under OpenBSD and encountered few issues during the building hence this patchset.

Hope it helps.

Thanks.


auriamg:
Hi, most of your patch seems fine, though the associated pull request doesn't seem right, maybe double-check the pull request was created correctly?

Otherwise, just one question, why is this required?

```
-		Value = (void *) tmp;
+		Value = reinterpret_cast<void *>(tmp);
 ``

devnexen:
Hi and thanks.
That might depend on compiler flags (maybe you generally use -fpermissive ?) for me it could not compile, in C++ it is better to use C+ cast operator I think.

PR #2416, opened 2016-01-02
highscores shared for all users

devnexen:


auriamg:
Overall the code seems clean, but I am not fully sure I understand what the coding code is used for. Could you elaborate on that?

Benau:
Seems the /var/games/supertuxkart directory is not writable by a normal user.. ( i tested with touch a file there)

Also after the first highscore is written, different users may not be able to write it again

Notice: in my linux, the highscore.xml is in -rw-r--r--, not sure in BSDs

hiker:
The locking is quite complex, especially since we don't really support running stk twice on the same machine. But one use case I see is to use a shared file system at an event to implement a 'global' highscore list ... but: the highscore manager reads all high scores at start up, and writes all back when a new high score was achieved ... which means if you have to copies running at the same time (on different machines but with a shared highscore file), the locking does not really help, two instances would overwrite each other's data each time a new highscore is achieved.
So we either need to also fix this, or officially state that we do not support running two instances at the same time (and add some tests to detect this).

devnexen:
Hi and thanks for feedback.
@auriamg @hiker There are few use cases when the shared highscore might need locking if the shared file is under NFS folder for example and two remote machines are launched in same time. Otherwise with my branch the shared highscore is not corrupted by my several users, did not test yet concurrent accesses though.
@Benau indeed chmod command via cmake might be a necessary update.

hiker:
@devnexen That's what I meant: if you have two concurrent runs, both will read the highscores at startup. Then say the first machine will have a race with a new highscore: it will update its internal data structures and write out its data to the file. If now the second machine has a highscore, it will update its internal data structures and write out its data to the file - overwriting the new highscore from the first machine. So in order for the locking to make a nfs mount actually usable, this needs to be changed as well (and special cases, e.g. a machine might detect a new highscore locally, but a new highscore was achieved on a different machine in between, so highscore messages should only be issued after locking and updating the highscore file).

PR #2423, opened 2016-01-09
shared highscore file + blocking several instances launches

devnexen:
Hi,

Regarding other feedbacks, finally the current version does not allow severe instances of stk running.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,9 +32,16 @@ endif()
 
 if(MINGW OR CYGWIN)
     option(USE_WIIUSE "Support for wiimote input devices" OFF)
+    if (NOT STK_SHARED_DIR)
+        set(STK_SHARED_DIR "${CMAKE_INSTALL_PREFIX}/supertuxkart/")
+    endif()
 else()
     option(USE_WIIUSE "Support for wiimote input devices" ON)
+    if (NOT STK_SHARED_DIR)
+        set(STK_SHARED_DIR "/var/games/supertuxkart/")
+    endif()
 endif()
+add_definitions(-DSTK_SHARED_DIR=\"${STK_SHARED_DIR}\")
 
 if(UNIX AND NOT APPLE)
     option(USE_XRANDR "Use xrandr instead of vidmode" ON)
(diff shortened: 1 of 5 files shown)


deveee:
I'm not convinced to this change. Some people prefer private hiscore file and other people prefer shared hiscore file, and it's just matter of taste what is better.

Most games don't use shared hiscore file, so I can assume that this is expected behaviour.

Also if someone wants to have shared hiscore file, then doing it is quite easy. You can just create a symlink or add one line to the script before running STK: 
```export HOME=/path/to/your/shared/config/directory```
But if someone wants to have private hiscore file, then creating second ```/var/games``` directory is not that easy.

And it potentially can cause some issues:
- read/write privileges to the shared file by multiple users (This has been solved for example in gnome-games by creating "games" group and adding the user to this group. But what if someone will create new user and won't add it to this group? This is potentially the root of issues.). I know that it's easy to say that we will just do ```chmod a+rwx```, but in practice after almost every big upgrade my hiscores in gnome-games were broken ;)
- running several STK instances is probably not usual, but I don't see any reason that I shouldn't do it if I want. I can start every other game twice and it doesn't complain that it is already running.
- if you have several STK versions installed on your computer, then all versions would need to use the only one and true shared hiscore file.

Also I didn't look carefully at your whole code, but I noticed for example, that you set ```STK_SHARED_DIR``` to ```${CMAKE_INSTALL_PREFIX}/supertuxkart/``` for MinGW and Cygwin. The ```${CMAKE_INSTALL_PREFIX}``` is set for MinGW to ```C:\Program Files\Supertuxkart``` by default, so probably there will be a problem with creating the file. Most probably it would require administrator privileges.

@hiker, @auriamg what do you think about it?

auriamg:
I appreciate the work you are trying to do with this PR but this is a can of worm that opens the door to so many issues

For instance, your PID check to prevent multiple instances - what if STK crashes, or is killed? Then I think it would be impossible to start it again (without knowing what file to delete). This certainly causes more issues than it solved

As for the shared highscore file, we are certainly going to run into permission issues here. Maybe it should be off by default, and distributions can set it up when building packages if they want? Overall the idea of shared highscores is not a bad one but we should not do it in a way that opens the door to so many issues

qwertychouskie:
Due to the aforementioned issues, should this be closed then?

PR #2430, opened 2016-01-27
Fix warnings

leyyin:


auriamg:
Seems fine to me, though since this touches physics and networking code I'll also let @hiker take a look

PR #2439, opened 2016-02-18
Update AngelScript to version 2.30.2

Odd0002:
Update the version of AngelScript bundled with SuperTuxKart from 2.30.0 WIP to 2.30.2.
This fixes compilation on ARM architectures and switches to a stable build of AngelScript.
Preliminary tests show no regressions: the tutorial still works correctly.

### DIFF
--- a/lib/angelscript/include/angelscript.h
+++ b/lib/angelscript/include/angelscript.h
@@ -58,8 +58,8 @@ BEGIN_AS_NAMESPACE
 
 // AngelScript version
 
-#define ANGELSCRIPT_VERSION        23000
-#define ANGELSCRIPT_VERSION_STRING "2.30.0 WIP"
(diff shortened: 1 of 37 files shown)


deveee:
Thanks for applying the cmake fix.

There was also one less important modification to silence cmake warning (commit 6af30b68f0ad5fa09fe0bf5160d8ea141f5e3575)

We certainly should update Angelscript before next release. Or at least apply the fix for ARM architectures. It's a shame that since two releases package maintainers have to patch STK to make it working :P

Odd0002:
OK, I applied that modification as well. Updating to 2.32.2 was pretty easy, but when trying to update to 2.31 WIP, the scriptmanager code in STK didn't compile.

auriamg:
Merged, thanks!

PR #2451, opened 2016-03-13
Update README

egirsova:
The inclusion of "The Xcode project file is in `/src/ide/Xcode/`" in the current README is misleading, since the `stk-code/src/ide` directory does not even exist in the project.


auriamg:
Indeed, I've merged it, thanks

PR #2454, opened 2016-03-17
Fix uninitialized scalar fields in addon.hpp

tobbi:
Reported by coverity.

### DIFF
--- a/src/addons/addon.hpp
+++ b/src/addons/addon.hpp
@@ -127,7 +127,18 @@ private:
     static SortOrder m_sort_order;
 
 public:
-         Addon() {};
+         Addon():
+            m_revision(),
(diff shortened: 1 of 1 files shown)


auriamg:
I don't think this constructor is used for anything other than creating an empty object to be imemdiately copied over, but this is a legitime fix still. However I think the proper fix would be to take the initialisation code in "Addon::Addon(const XMLNode &xml)" and move it into Addon::Addon, and have the XML constructor delegate to that code (or use a reset method as we usually do). THat would be a lot more consistent with our coding conventions (we bsaically never use initialisers for ints/other basic types)

hiker:
Fixed differently in e1ce01b5d2acdd9412928b9d1c5833df274338d1

PR #2455, opened 2016-03-17
Add missing unlock

tobbi:
Fixing a coverity issue.


auriamg:
assigning to hiker since I'm not familiar with this sound locking code

PR #2464, opened 2016-03-27
Fix debug build on GCC 5.3.0

Odd0002:
Replaced instances of \<math.h\> with \<cmath\>, and replaced all instances of isnan() with std::isnan(). This fixes the debug build on GCC 5.3.0. 
Requires testing on Windows and Mac.

### DIFF
--- a/lib/bullet/src/BulletCollision/CollisionDispatch/btCollisionObject.h
+++ b/lib/bullet/src/BulletCollision/CollisionDispatch/btCollisionObject.h
@@ -16,19 +16,7 @@ subject to the following restrictions:
 #ifndef BT_COLLISION_OBJECT_H
 #define BT_COLLISION_OBJECT_H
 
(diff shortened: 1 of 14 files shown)


auriamg:
This pull request seems good overall, however there is one issue : you have modified bullet, we generally try to avoid modifying bullet. Do you know if a newer upstream version of bullet has already fixed this issue, or if there is another less intrusive fix? The STK parts are fine but the bullet parts are more problematic, while we do host the bullet sources we haven't actually forked it nor want to maintain our own fork of bullet

hiker:
Using a newer version of bullet might introduce a lot of problems, they don't really maintain compatibility. The code changed was mostly (if not exclusively) introduced by us anyway, so I am ok with applying the patch (assuming it does not break anything else)

auriamg:
Alright, if hiker agrees then we can merge. THanks!

PR #2467, opened 2016-04-01
Possible to turn volume all the way down

scootergrisen:


Benau:
the min_value 0 is a quick hack for just for you (private use), the proper fix is to detect whether this widget is zero, and then SFXManager::get()->toggleSound(false) or music_manager->stopMusic().

qwertychouskie:
Isn't this what the "Enabled" checkbox right above the volume control for?

auriamg:
The enabled checkbox definitely needs to be considered here - if we allow the sound to go to 0 then the code will need to detect it and actually turn the feature off completely, that would be preferable

qwertychouskie:
I think due to the "Enabled" toggle right above the volume level control, (and a lack of activity on this PR,) this should be closed.

scootergrisen:
Why not fix the problem instead?

nado:
Because you’re the only one to see a problem here ?

scootergrisen:
Have you ever seen a volume control anywhere that is not made to turn the volume all the way down?

auriamg:
Should we receive a PR that fully disables sound when it's turned down (i.e. really disables audio) then we could merge it. Meanwhile though, this is an extremely minor feature (you can get the same effect in one click) so I don't think anyone from the main team will be working on this

qwertychouskie:
Should this be closed then?

auriamg:
Yes, closing, this PR won't be merged as-is. If anyone wants to make another improved pull request, please create a new PR

PR #2475, opened 2016-04-26
Support disabling race intro sound

antoniusriha:
I've added a user config entry for disabling the sound at race start. This is useful when you're listening to some music on your PC and don't want it to clash acoustically with the race start sounds.

### DIFF
--- a/src/modes/cutscene_world.cpp
+++ b/src/modes/cutscene_world.cpp
@@ -58,7 +58,8 @@ CutsceneWorld::CutsceneWorld() : World()
     m_aborted = false;
     WorldStatus::setClockMode(CLOCK_NONE);
     m_use_highscores = false;
-    m_play_racestart_sounds = false;
+    m_play_track_intro_sound = false;
+    m_play_ready_set_go_sounds = false;
     m_fade_duration = 1.0f;
 }   // CutsceneWorld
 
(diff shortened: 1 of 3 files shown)


auriamg:
Hi,

this is not a bad idea but I'm really not sure adding a new flag is necessary. I think it may be better to just force these sounds off if music is disabled, since they are kind of like music? That would avoid the creation of yet another flag. That is kind of a hack but is much simpler

antoniusriha:
I personally wouldn't have a problem with disabling the race start sounds with the music flag. Some people though might need the acoustic cue to time the race start better. This is why I added a new flag.

auriamg:
I was only referring to the little "get ready" music that plays _before _"ready-set-go". The little chimes that play on ready-set-go are definitely sound effects and should stay there

antoniusriha:
I've removed the new additional flag. Race intro sound is now controlled by the music_on flag. Ready-set-go sounds are always on.

Benau:
Minor problem:
If no m_track_intro_sound is played, the race will start faster (ie thunderbird will appear immediately)

antoniusriha:
I have fixed the timing issue by adding a delay. This resulted in an awkward silence at the beginning of the race, so I decided to make the kart engines start earlier, if the intro music is off.

auriamg:
Seems good, thanks. I'm not sure I like having a special case for the engines, but I don't have anything better to suggest so that will have to do

PR #2486, opened 2016-05-04
Fix jumbled up human names in gp ranking

antoniusriha:
In the ranking shown in between races during a gp the names of human players have been determined
using the assertion that the human players start from the rear of the field. This is not true for
gps, where the start position is determined by the overall standings, which lead to the human's
name being displayed with the wrong kart in the rankings. This commit fixes the issue by accurately
determining the start position of a human's kart.

### DIFF
--- a/src/states_screens/race_result_gui.cpp
+++ b/src/states_screens/race_result_gui.cpp
@@ -493,14 +493,7 @@ void RaceResultGUI::backToLobby()
             RowInfo *ri = &(m_all_row_infos[position - first_position]);
             ri->m_is_player_kart = kart->getController()->isLocalPlayerController();
 
-            // Identify Human player, if so display real name other than kart name
-            const int rm_id = kart->getWorldKartId() -
-                (race_manager->getNumberOfKarts() - race_manager->getNumPlayers());
-
-            if (rm_id >= 0 && !race_manager->isWatchingReplay())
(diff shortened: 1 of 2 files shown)


src/states_screens/race_result_gui.cpp
```diff
@@ -1446,3 +1433,32 @@ void RaceResultGUI::backToLobby()
         assert(m_font != NULL);
         return m_font->getDimension(L"A").Height; //Could be any capital letter
     }
+    
+    // ----------------------------------------------------------------------------
+    irr::core::stringw RaceResultGUI::getKartNameForResult(const AbstractKart *kart, World *world)
+    {
+        // Identify human player, if so display real name other than kart name
```
Benau: should be identify AI kart name

auriamg:
Thanks for the pull request!

I still think we could better to convert between locate/remote IDs though. @hiker, could you take a look at this? I'm pretty sure we can come up with something better than subtracting the number of AI karts

hiker:
Hi,

thanks a lot for the patch. I've applied a slightly different patch that should fix the issue, see 6c9f3d73b54bd43108c16d9cbb0439be9f45fe3c and (to my embarrassment) bc418acb05cada255c129e8a03ef3c994041f6f1.

Cheers,
   Joerg

PR #2494, opened 2016-05-11
Fix dependencies list for Ubuntu compilation

qwertychouskie:
A few more packages were needed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,7 +27,7 @@ Ubuntu command:
 
 ```
 sudo apt-get install autoconf automake build-essential cmake libogg-dev libvorbis-dev libopenal-dev libxxf86vm-dev \
-libgl1-mesa-dev libglu1-mesa-dev libcurl4-openssl-dev libfribidi-dev libbluetooth-dev libxrandr-dev libfreetype6-dev
+libgl1-mesa-dev libglu1-mesa-dev libcurl4-openssl-dev libfribidi-dev libbluetooth-dev libxrandr-dev libfreetype6-dev libpng12-dev pkgconf libjpeg-dev
 ```
   
 Unpack the files from the tarball like this:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,7 +26,7 @@ First, make sure that you have the following packages installed:
 Ubuntu command:
 
 ```
-sudo apt-get install autoconf automake build-essential cmake libogg-dev libvorbis-dev libopenal-dev libgl1-mesa-dev libglu1-mesa-dev libcurl4-openssl-dev libfribidi-dev libbluetooth-dev libxrandr-dev libfreetype6-dev libpng12-dev pkgconf libjpeg-dev
+sudo apt-get install autoconf automake build-essential cmake libogg-dev libvorbis-dev libopenal-dev libglu1-mesa-dev libcurl4-openssl-dev libfribidi-dev libbluetooth-dev libxrandr-dev libfreetype6-dev libpng12-dev pkgconf libjpeg-dev
 ```
   
 Unpack the files from the tarball like this:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,7 +26,7 @@ First, make sure that you have the following packages installed:
 Ubuntu command:
 
 ```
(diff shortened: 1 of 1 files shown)


deveee:
We also don't need autotools anymore and I'm not sure why we need pkgconf?

Here is the list of dependencies from debian package:
 cmake (>= 2.8.1),
 debhelper (>= 9),
 docbook-to-man,
 libalut-dev,
 libbluetooth-dev [linux-any],
 libbz2-dev,
 libcurl4-gnutls-dev | libcurl4-dev,
 libenet-dev (>= 1.3.1),
 libfribidi-dev,
 libjpeg-dev,
 libopenal-dev,
 libpng-dev,
 libsdl-image1.2-dev,
 libsdl-ttf2.0-dev,
 libvorbis-dev,
 libxrandr-dev,
 libxxf86vm-dev,
 mesa-common-dev,
 zlib1g-dev

So I'd say that we need:
build-essential cmake libalut-dev libbluetooth-dev libcurl4-gnutls-dev  libenet-dev  libfribidi-dev libjpeg-dev libopenal-dev  libpng-dev  libvorbis-dev  libxrandr-dev  mesa-common-dev zlib1g-dev

But I didn't check it on clean system. You can easily check it using debootstrap if you want.

qwertychouskie:
According to http://packages.ubuntu.com, `libgl1-mesa-dev` depends on `libxxf86vm-dev` for all listed Ubuntu versions.

qwertychouskie:
pkgconf is needed to detect certain lib-devs.

qwertychouskie:
`libalut-dev` does not seem to be needed, as it is not installed on my system, and supertuxkart works just fine (including directional sound).

qwertychouskie:
`libenet-dev` also does not seem to be needed, as it is not installed on my system, and supertuxkart works just fine (including network connections).

deveee:
Here is the complete list (tested on Ubuntu Wily using debootstrap):
```
sudo apt-get install build-essential cmake libbluetooth-dev libcurl4-gnutls-dev libfreetype6-dev libfribidi-dev \
libgl1-mesa-dev libglu1-mesa-dev libjpeg-dev libogg-dev libopenal-dev libpng-dev libvorbis-dev libxrandr-dev \
mesa-common-dev zlib1g-dev
```

Pkgconfig is not needed and also autotools are not needed. 

The mesa-common-dev is needed because it contains most of the header files. 

And I would leave libgl1-mesa-dev in the list because it contains libGL.so library, which is probably needed for linking. It is indeed installed as a dependency, but you can't be sure that it will be still a dependency in next few years ;)

The same situation is with zlib. It is a libpng dependency, but we are sure that we need both of them.

And I think that we should use libcurl4-gnutls-dev instead of openssl because it is used by default in debian/ubuntu packages for STK and also for other projects. I also use it for our static packages.

I don't want to make a commit myself while you already created a pull request with 4 commits. Do you want to correct it yourself? Ideally as a separate single-commit pull-request to make it easier to port it to 0.9.2 branch.

qwertychouskie:
Form my testing, STK failed to compile without `libglu1-mesa-dev`, and according to http://packages.ubuntu.com, `libglu1-mesa-dev` depends on `libgl1-mesa-dev` for all listed Ubuntu versions, and `libgl1-mesa-dev` depends on `mesa-common-dev` for all listed Ubuntu versions.  I personally would prefer to let apt handle as much as possible to avoid weirdness with packages like `mesa-common-dev-lts-wily`.  Also, on my system, certain libs would not be detected without `pkgconf`, though `pkg-config` should work too.

deveee:
Ok, pkg-config is indeed required too. It is used in cmake scripts for finding freetype and fribidi library:
```
elseif(UNIX)
    include(FindPkgConfig)
    pkg_check_modules(FREETYPE freetype2)
else()
```

> libglu1-mesa-dev depends on libgl1-mesa-dev for all listed Ubuntu versions, and libgl1-mesa-dev depends on mesa-common-dev

Sure, but GLU library is just few minor tools such as setting camera view. And what we really need is libgl.so library and opengl headers (gl.h, glext.h, glx.h etc...)

qwertychouskie:
I would personally rather only list `libglu1-mesa-dev` and let apt handle the rest.  In theory, this list should be the most concise and updated list, and Debian and Ubuntu packages' build-depends should be based on this, not the other way around.

auriamg:
Merging the current state as it seems like a good start. Anyone can still point out required tweaks/additions

PR #2512, opened 2016-06-06
Updated zlib by a minor release, and updated the .gitignore to remove…

bkeys:
… the generated build files
Let me know if I need to change anything, hopefully this helps!


deveee:
I'm not sure why you added ```lib/zlib/``` to .gitignore. These files simply shouldn't be built there :) As I see you didn't touch CmakeLists.txt file, so everything should work as it was before. You also don't need to add '''lib/zlib/Makefile''' when you already added ```lib/zlib/```

There are also some useless files, for example ```lib/zlib/Makefile``` which contains data from your personal build, eg: ```cd /home/bkeys/Devel/DMUX && $(MAKE) -f CMakeFiles/Makefile2 lib/zlib-1.2.8/clean```

Is there a particular reason to update zlib?

ComradeKeys:
Whoops sorry about that, I have cleaned up that section of the .gitignore, if the files should not be built there then it leaves me to wonder why they do I believe that I copied the old one from the previous version of zlib
No real reason you guys should update, we just have the most recent version of zlib in our project so I wanted to update it for you guys too

auriamg:
Hi, we're not ignoring this PR but we're in the process of making a release, and upgrading libraries at the last minute is not a good idea. We can come back to this PR after the release

nado:
Since 0.9.2 is out, how lib updates should be treated now ?

deveee:
This PR still contains unused files, such as CTestTestfile.cmake, Makefile, libz.so and still modifies .gitignore file, which IMO doesn't have a sense.

nado:
Also, why not use git submodule ? https://github.com/madler/zlib/tree/master is their stable branch (1.2.8 atm), almost nothing to configure.

deveee:
There is available newer zlib version and the pull request contains unwanted files from local build, so I'm closing this as invalid

PR #2513, opened 2016-06-08
Toggle graphical options from command line

### LINKED ISSUE #2092: Adding CLI options to toggle shadows/lights/etc
We should have a way to change the default configuration by CLI option (maybe an option to load an alternate file, or something similar) it will become easier to do benchmarks with STK.

http://www.phoronix.com/forums/showthread.php?117175-Running-Benchmarks-On-The-Newest-Open-Source-OpenGL-3-1-Game&p=484950#post484950

LoadingPleaseWait:
This closes #2092. User configuration options can be done in the command line.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -469,7 +469,7 @@ void setupRaceStart()
         Log::warn("main", "Kart '%s' is unknown so will use the "
             "default kart.",
             UserConfigParams::m_default_kart.c_str());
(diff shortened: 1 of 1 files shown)


auriamg:
Thanks, seems good!

Now as a complement, if you also want to document them in the string that is printed when you run with --help, that would be perfect

PR #2516, opened 2016-06-14
More information in race result GUI

### LINKED ISSUE #1805: More information in summary chart
Do not only display order of contestant and time, show also best Lap Time, Difficulty and number of laps.

LoadingPleaseWait:
Best lap time, difficulty and number of laps are displayed. This closes #1805. `linear_world.hpp` was edited to add an accessor for `m_fastest_lap`.

### DIFF
--- a/src/states_screens/race_result_gui.cpp
+++ b/src/states_screens/race_result_gui.cpp
@@ -828,7 +828,7 @@ void RaceResultGUI::backToLobby()
         if (race_manager->getMajorMode() != RaceManager::MAJOR_MODE_GRAND_PRIX ||
             m_animation_state == RR_RACE_RESULT)
         {
-            displayHighScores();
+            displayPostRaceInfo();
         }
     }   // renderGlobal
 
@@ -1322,7 +1322,7 @@ void RaceResultGUI::backToLobby()
     }   // cleanupGPProgress
 
     // ----------------------------------------------------------------------------
-    void RaceResultGUI::displayHighScores()
+    void RaceResultGUI::displayPostRaceInfo()
     {
         // This happens in demo world
         if (!World::getWorld())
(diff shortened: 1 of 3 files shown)


Benau:
m_top + 550 will lead to text rendered outside the border is smaller resolution is used

My suggestion:

Put your new code in displayHighScore (rename a good one later), than you can share the "current_y" there (and white_color too)

Also \_(xxx) is for translation, consider something like_("Best lap time: %s", StringUtils::timeToString(best_lap_time).c_str())

if (xxx) { should be
if (xxx)
{
Details here:
http://supertuxkart.sourceforge.net/Coding_Style

LoadingPleaseWait:
The race result GUI now looks like this. https://imgur.com/a/apAM4

Benau:
Nice PR

src/states_screens/race_result_gui.cpp
```diff
@@ -1410,6 +1413,36 @@ void RaceResultGUI::backToLobby()
                     false, false, NULL, true /* ignoreRTL */);
             }
         }
+
+        if (race_manager->getMinorMode() != RaceManager::MINOR_MODE_SOCCER
+            && race_manager->getMinorMode() != RaceManager::MINOR_MODE_EASTER_EGG)
+        {
+            // display lap count
+            if (race_manager->getNumLaps() < 9000)
```
leyyin: Does the number `9000` hold any significance here and below `best_lap_time < 9000`, why not `9001`?

src/states_screens/race_result_gui.cpp
```diff
@@ -1410,6 +1413,36 @@ void RaceResultGUI::backToLobby()
                     false, false, NULL, true /* ignoreRTL */);
             }
         }
+
+        if (race_manager->getMinorMode() != RaceManager::MINOR_MODE_SOCCER
+            && race_manager->getMinorMode() != RaceManager::MINOR_MODE_EASTER_EGG)
```
auriamg: Here it would be pfereable to use race_manager->modeHasLaps() to determine whether it is appropriate to show this section. then the "9999" laps check could also be removed. This would be preferable because if we add more modes in the futures, we won't need to search for places where lists of modes were hardcoded.

PR #2525, opened 2016-06-21
Angelscript 2.32.0

ignatenkobrain:


qwertychouskie:
All checks have failed
1 failing check
continuous-integration/travis-ci/pr — The Travis CI build failed

Hmm... It seems Clang does not like your code...

ignatenkobrain:
@qwertychouskie actually it's bug somewhere in angelscript.

auriamg:
ok thanks, but I don't think we will merge if the build is failing...

Also, is there a specific reason for updating, or just the good practice of keeping up to date with no specific reason?

ignatenkobrain:
@auriamg I will report this issue to angelscript upstream, so just don't close this (once we will have fix, I will push it here).

About update.. In Fedora we are using system-wide angelscript and we are keeping it up2date so it was a bit problem when STK stopped compiling. Also when new version of angelscript is supported in upstream it is always better.

qwertychouskie:
Is there a link for the upstream report?

ignatenkobrain:
some quick investigation:
```
#ifdef AS_CAN_USE_CPP11
...
#include <type_traits>
...
#endif
```

I guess we must define it when compiling, if CI will fail again - I will do that (we use c++11, so it's okay)

ignatenkobrain:
@qwertychouskie I was not able to find bugtracker in upstream, so I sent bugreport to main developer via mail.

ignatenkobrain:
@qwertychouskie actually it is a bug in clang: https://llvm.org/bugs/show_bug.cgi?id=25880

deveee:
I noticed a fix for clang build, though I didn't check it yet:
https://sourceforge.net/p/angelscript/code/2335/

ignatenkobrain:
So, I rebased this to newest 2.31.2 which should fix all secondary architectures and build with this version of clang.

ignatenkobrain:
So, it still fails...
```cpp
/home/travis/build/supertuxkart/stk-code/lib/angelscript/source/as_string_util.cpp: In function ‘asQWORD asStringScanUInt64(const char*, int, size_t*, bool*)’:

/home/travis/build/supertuxkart/stk-code/lib/angelscript/source/as_string_util.cpp:189:28: error: ‘UINT64_MAX’ was not declared in this scope

/home/travis/build/supertuxkart/stk-code/lib/angelscript/source/as_string_util.cpp:216:29: error: ‘UINT64_MAX’ was not declared in this scope
```

but this is due upstream needs c++11 std as only its version defines UINT64_MAX in `cstdint`..

auriamg:
Indeed, both Clang and vc++ builds fail for different reasons :/ they should work before we upgrade.

qwertychouskie:
@ignatenkobrain It may be worth a try using the latest WIP (http://www.angelcode.com/angelscript/wip.php), as relying on an old bundled version could cause headache for packagers packaging 0.9.3.

nado:
I believe I made a working copy for travis by installing newer versions of gcc/clang back in the day.
Never got any reply about that from @ignatenkobrain

ignatenkobrain:
Hopefully now it should work (I also updated angelscript to newly released 2.32.0)

ignatenkobrain:
@qwertychouskie @nado @auriamg @deveee Yeah, it worked!

Feel free to re-review this PR and sorry for such long delay ;)

deveee:
Modifications in .travis.yml are not needed, because as I see travis already uses clang 5.0.

I tested and it compiles for me without issues on gcc 4.7.2 and clang 4.0.

@auria are you ok with clang >= 4.0? If so, then I would merge it.

It only needs to revert .travis.yml and fix cmake warning in angelscript CMakeLists.txt:
```
if(NOT (CMAKE_MAJOR_VERSION VERSION_LESS 3))
  cmake_policy(SET CMP0048 OLD)
endif()
```

auriamg:
@deveee sounds good to me, ok

deveee:
Already commited in https://github.com/supertuxkart/stk-code/commit/aec7ca0ce911544e18722cf6387050acf347f852
Thanks!

deveee:
It looks that travis with clang 5.0 fails even if compilation with clang 4.0 works on my linux, so that it depends on something more than just clang version (libstdc++ ?)

It should be possible to revert 
https://sourceforge.net/p/angelscript/code/2278/

I already asked Andreas if it's safe and if there is a better fix.

PR #2526, opened 2016-06-27
couple of resource leaks fixes

devnexen:


auriamg:
Seems good, thanks for the contribution

PR #2531, opened 2016-06-28
Update TODO.md

qwertychouskie:


qwertychouskie:
Just realized I had not updated the link, just fixed.

auriamg:
I'm actually wondering if we should keep this file at all, we're not really using it.

In any case, I wonder why you added "and background images" to the TODO ?

qwertychouskie:
I didn't add it, just changed the comma to an "and".  More easily seen this way: https://github.com/supertuxkart/stk-code/pull/2531/files?short_path=0ff387c#diff-0ff387c7e4dec76c01cb77aa4d338ee1

PR #2532, opened 2016-06-28
small mem leak fix for player XML data

devnexen:


auriamg:
I believe this is usually freed in "initRemainingData", are there cases where it's not the case?

devnexen:
ah true

PR #2554, opened 2016-07-03
Minor fixes for issues found while cross-building

ezequielgarcia:
Issues found while adding stk as a package for Buildroot, which allows playing it on embedded x86 platforms such as the Minnowboard family.
See https://patchwork.ozlabs.org/patch/634103/


auriamg:
Do you think you could submit your changes as a pull request and not a raw patch? Pull requests are the preffered way to go when using git(hub)

ezequielgarcia:
Isn't this a pull request?

/me confused

auriamg:
Oops I'm sorry, I saw the link to a patch and didn't notice this was a PR :) by bad, github displays issues and PR almost the same

Seems good for the most part, though you removed an include to "/usr/X11R6/include". I understand this may not be available everywhere, but this include was not added without reason, on some linux/unix systems it is required, so I don't think we can just remove it like that

Issue #2544, opened 2016-07-02
License says "Copyright Â© 2007" with a Â charachter


scootergrisen:
Windows installer for STK 0.9.2 have a Â character in the GPL license that is showed during installation just before "© 2007 Free Software Foundation, Inc."

It look like "Copyright Â© 2007 Free Software Foundation, Inc."

scootergrisen:
![stk1](https://cloud.githubusercontent.com/assets/1836590/16540437/022281b8-4065-11e6-9c77-c97aa2bc894b.png)

scootergrisen:
Also there is a long __________ line that wraps.
Maybe it would be better to show the license in it own box or something so that line is not needed.
![stk2](https://cloud.githubusercontent.com/assets/1836590/16540440/0ee98248-4065-11e6-9eee-f2371422c87d.png)

auriamg:
@hiker maybe an encoding issue in the installer?

As for the wrapping "____", yes it happens but it's exessively minor and I don't think anyone cares about that...

qwertychouskie:
`COPYING` seems to be UTF-8, but that needs [a Unicode version of NSIS](http://www.scratchpaper.com/).

qwertychouskie:
Also see https://en.wikipedia.org/wiki/Nullsoft_Scriptable_Install_System, towards the bottom.  Maybe we just need to use v3.0-rc2?

qwertychouskie:
From http://nsis.sourceforge.net/Download:
> NSIS 3.0
> Released July 24, 2016 
One day after I suggested 3.0-rc2 :)
Now that it is officially stable, it probably should be upgraded in STK.

qwertychouskie:
Obsoleted by #2661 (since the issue is really about Unicode support in the installer), can be closed.

hiker:
Fixed in 6476af46cdf1c0279e6b11a208059eec827ff45b.

PR #2568, opened 2016-07-11
Fix COPYING (including wrapping underline in installer)

qwertychouskie:
See https://github.com/supertuxkart/stk-code/issues/2544#issuecomment-230100514.


auriamg:
Assigning to hiker. Not sure changing our license file is the best way to deal with this...

qwertychouskie:
Any chance of getting this merged? It's a pretty small change....

qwertychouskie:
Also ping @MTres19, as he handles license stuff (thanks for your work!).

auriamg:
Also, you added "or at your option, any later version.". You definitely should NOT change the term of the license.

qwertychouskie:
@auriamg https://github.com/supertuxkart/stk-code/blob/master/src/main.cpp#L9 includes the "or any later version", which is why I added it.  https://github.com/supertuxkart/stk-code/search?utf8=%E2%9C%93&q=%22or+%28at+your+option%29+any+later+version.%22&type= confirms this is probably correct and that this file is not.  If you want though, I can remove it, though this file probably should match the rest of the code.

PR #2573, opened 2016-07-14
Place buttons in confirm dialog horizontally

LoadingPleaseWait:
When you start SuperTuxKart for the first time the buttons will now be next to each other horizontally.

### DIFF
--- a/data/gui/confirm_dialog.stkgui
+++ b/data/gui/confirm_dialog.stkgui
@@ -5,7 +5,7 @@
 
         <spacer height="25" width="10" />
 
-        <ribbon id="buttons" height="30%" width="30%" align="center">
+        <buttonbar id="buttons" height="30%" width="30%" align="center">
 
             <icon-button id="confirm" icon="gui/green_check.png" I18N="In a 'are you sure?' dialog"
                 text="Yes" align="center"/>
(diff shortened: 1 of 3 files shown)


auriamg:
Thanks for the pull request, but this will not navigate correctly using a keyboard or gamepad (will not be possible to navigate between buttons with left/right). The way to do left-right navigation in STK is with ribbon controls

LoadingPleaseWait:
Would it be alright if I use a `RibbonWidget` with `IconButtonWidgets` containing the green check mark image and the red dash?

auriamg:
Yes that would work

LoadingPleaseWait:
I've added a commit to make the confirm dialog use a ribbon widget.

auriamg:
Hi,

the "m_cancel_selected" part is an hack that should be avoided, and does not work properly. The proper solution is to use <buttonbar> instead of <ribbon> (ribbon is meant as some kind of "radio button", for instance to select race difficulty, whereas buttonbar requires the select button to be pressed)

auriamg:
Oh also, your pull request contains changes to file "message_dialog.hpp" but you don't appear to have changed anything in this file, your editor just de-indented empty lines (probably as part of "remove trailing whitespace" feature, but we don't consider empty indented lines to be trailing whitespace)

LoadingPleaseWait:
A private instance variable is added in the file, `message_dialog.hpp` but you are correct Atom has been removing the indentation from blank lines and I will change that. I will also try using a buttonbar.

LoadingPleaseWait:
The message dialog should now use a button bar. Also, the white space on blank lines is back. Sorry that this has taken so long.

auriamg:
Everything seems good, thanks! I've merged your pull request

PR #2582, opened 2016-07-24
fix freetype header issue as in some distros the root folder could be freetype2

devnexen:


Benau:
```
/home/travis/build/supertuxkart/stk-code/src/font/font_with_face.cpp:29:21: fatal error: ftoutln.h: No such file or directory
```

Same on my linux :/

Benau:
In my linux:
```
bash-4.3$ ls -l /usr/include/freetype2/
total 4
drwxr-xr-x 1 root root  832 6月  17 22:56 freetype
-rw-r--r-- 1 root root 2383 6月  17 22:56 ft2build.h

**Comment by devnexen** [comment]
normally FREETYPE_INCLUDE_DIRS include root folder of free type right ? or in older version of cmake  it does not ?

**Comment by Benau** [comment]
Seem I used pkgconfig to detect freetype, any idea?
```
include(FindPkgConfig)
pkg_check_modules(FREETYPE freetype2)

Benau:
```
/usr/bin/c++   -DDEBUG -DENABLE_BIDI -DENABLE_LIBBFD -DENABLE_WIIUSE -DGLEW_NO_GLU -DHAVE_OGGVORBIS -DSUPERTUXKART_DATADIR=\"/usr/share/supertuxkart\" -I/data/game/stk-code/lib/bullet/src -I/data/game/stk-code/lib/enet/include -I/data/game/stk-code/lib/glew/include -I/data/game/stk-code/lib/irrlicht/include -I/data/game/stk-code/lib/wiiuse -I/data/game/stk-code/src -I/data/game/stk-code/lib/angelscript/include -I/usr/include/AL -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/harfbuzz -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/fribidi  -fopenmp -std=gnu++0x -Wall -Wno-unused-function -g   -o CMakeFiles/supertuxkart.dir/src/font/font_with_face.cpp.o -c /data/game/stk-code/src/font/font_with_face.cpp
```
This is what I do in my linux

devnexen:
never noticed the included findfreetype module ... is it better if we simply use the existing cmake module instead ? it might exists already even for cmake 2.8.4

Benau:
Because we need to handle freetype in windows and mac too

devnexen:
ok then we can remove the else if UNIX case ? what do you think @Benau ?

Benau:
Did you mean:
```
elseif(APPLE)
    find_path(FREETYPE_INCLUDE_DIRS NAMES freetype/freetype.h PATHS "/Library/Frameworks/FreeType.Framework/Versions/2.4/unix/include")
    find_library(FREETYPE_LIBRARY NAMES freetype PATHS "/Library/Frameworks/FreeType.Framework/Versions/2.4/")
    include_directories(/Library/Frameworks/FreeType.Framework/Versions/2.4/unix/include)
    set(FREETYPE_FOUND 1)
    set(FREETYPE_LIBRARIES ${FREETYPE_LIBRARY})
else()
    include(FindPkgConfig)
    pkg_check_modules(FREETYPE freetype2)
endif()
```

Works for you (maybe because you use *BSD?)

devnexen:
Not exactly I thought renaming this included cmake module which is used only for win32 and darwin platforms, removing then the useless elseif(UNIX) case in it and include for unix the existing cmake module which. so in the root CmakeLists, it would look like this ?

if (APPLE OR WIN32)
    include(FreetypeAppleWin32)
else()
    include(Freetype)
endif()

Just a rough idea. In fact I found the build issue in first place in debian stable although indeed I use BSD mainly ;)

Benau:
```
elseif(APPLE)
    find_path(FREETYPE_INCLUDE_DIRS NAMES freetype/freetype.h PATHS "/Library/Frameworks/FreeType.Framework/Versions/2.4/unix/include")
    find_library(FREETYPE_LIBRARY NAMES freetype PATHS "/Library/Frameworks/FreeType.Framework/Versions/2.4/")
    include_directories(/Library/Frameworks/FreeType.Framework/Versions/2.4/unix/include)
    set(FREETYPE_FOUND 1)
    set(FREETYPE_LIBRARIES ${FREETYPE_LIBRARY})
else()
endif()
```

include(FindFreetype) crash my cmake :/

Benau:
Maybe only keep commit b3b907f974095e5ed74381550383c74877d0bd2d?

Benau:
Hang on, can you remove all patch and use:

#include FT_OUTLINE_H in font_with_face.cpp instead?

Benau:
https://lists.nongnu.org/archive/html/freetype-devel/2013-12/msg00045.html

it said:
in x2go-nx-libs.  Applications should use header file name macros, as
shown in all FreeType example files and demo programs.  If you change
that to the canonical

  #include <ft2build.h>
  #include FT_OUTLINE_H

everything works just fine, regardless of the internal structure of
the FreeType header directories.

My fault :/

devnexen:
works fine in my side and you solved it that s the most important :)

Benau:
Fixed, thanks for your help!

Issue #34, opened 2014-02-06
Ready, Set, Go! Penalty Removal and Revving


TorbenCarrington:
This ticket makes echo to this forum thread: http://forum.freegamedev.net/viewtopic.php?f=17&t=5028

Pressing the accelerate key to soon leads to a penalty that is frustrating, particularly to new players. We should have a way to make noise with the engine before the race.

Proposal:
- Remove the penalty time.
- Improve the startup boost mechanism to incorporate revving and potentially burning out.
- Adjust AI to work with the new setup.

Alayan-stk-2:
Can this be considered done, considering the changes to the startup boost done because of networking ?

PR #2586, opened 2016-07-27
Show boost at startup

qwertychouskie:
#34 seems abandoned, this is (hopefully) the first in a series of PRs to get the same functionality, as #34 has merge conflicts.

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -2110,7 +2110,7 @@ void Kart::updatePhysics(float dt)
     {
         m_has_started = true;
         float f       = getStartupBoost();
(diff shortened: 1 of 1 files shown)


auriamg:
Seems good, I just wonder if we can avoid hardcoding "0.9f" ? I'm not sure where that number comes from

qwertychouskie:
I think it comes from the above line. :)

Benau:
(above line) float Kart::getStartupBoost() const will return 0 if no startup boost is triggered, see kart.cpp:1011

Maybe you meant if (f >= 0.0f)?

Also please don't copy code from elsewhere without studying what it's really do

qwertychouskie:
OK, fixed.

auriamg:
Merged, thanks

Issue #1763, opened 2014-12-05
Display allow stk to connect to internet when player try to download addons


samuncle:
Hi 

While I was testing stk I found this rather annoying thing. If you don't allow stk to connect to internet the first time, later you might open the addons and have no clue why you can't download them.

The easiest and user friendly way of doing it would be to simple display to users the popup that is shown the first time
![connect](https://cloud.githubusercontent.com/assets/6406045/5310934/33c02c5a-7c3c-11e4-9bb7-44150ae8a361.jpg)

hiker:
Is this actually an issue? I just tried it, and if you click on addons, you get a popup window telling you what you need to enable. I think that is good enough. I would consider this a 'Wont' fix".

hiker:
It is indeed more user friendly, but not necessary for 0.8.2, so postponing it.

auriamg:
Fixed in #2605

PR #2605, opened 2016-08-18
In the addons screen, remember that the Internet connection is disabled 

Alpt:
This commit is for closing issue #1763 .

In the addons screen, remember if the Internet connection is disabled.

### DIFF
--- a/data/gui/addons_screen.stkgui
+++ b/data/gui/addons_screen.stkgui
@@ -33,5 +33,8 @@
         <box proportion="1" width="98%" align="center" layout="vertical-row" padding="6">
             <list id="list_addons" x="0" y="0" width="100%" height="100%"/>
         </box>
+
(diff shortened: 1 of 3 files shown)


auriamg:
The code seems good, thanks

One question, why does the text scroll? There seems to be enough room to display it normally (scrolling is usually used if the text is longer than available space). Also, the warning seems to disappear after it's done scrolling, is it intended? I think it would be beneficial to just leave the text there

Alpt:
One can stay with no Internet connection by choice. Then, why should he always be annoyed by being reminded that the Internet connection is disabled? This is why I did not want to be too annoying and make the text disappear after one scroll.
The text scrolls for effect. It seemed a good reminder effect to me.
Shall I make the text static and stop making the text disappear?
Let me know.

auriamg:
Okay thanks, your explanation sounds good enough. Will merge, thanks!

PR #2611, opened 2016-08-23
Documentation and tooling improvements

konstin:


TODO.md
```diff
@@ -1,30 +0,0 @@
-##TODO
```
qwertychouskie: I personally think it is best to leave this here...

auriamg:
what is this pull request about? there's no description.

Some notes: there's a reason why the installation file only contains links, we don't want to have instructions in 2 places, that makes it annoying to update/maintain. 
You also moved the credits file, that will probably break the credits screen. 
The svn-config file you deleted is still partly relevant for the media repo
You also deleted a bunch of helper scripts, why?

doc/conventions.txt
```diff
@@ -1,25 +0,0 @@
-Here are the guidelines to follow when coding SuperTuxKart:
```
qwertychouskie: Why delete this file?  It seems like this would be useful for first-time contributers...

konstin:
This Pull Request is about removing outdated files and improving the documentation in the repository. Here's each change explained:
 * Removing "AUTHORS": This file was just a stub with no content
 * Moving "CREDITS": I did forget about hte credits screen; the file is moved back now.
 * Removing "SVN-CONFIG": Unless svn does some very weird stuff I don't know about, there is no use in having this file in the stk-code folder whilst svn is only used for the stk-assest folder.
 * Removing "run-me.sh", "batch.py" and "test.sh": They are for a workflow in a very specific environment and afaik not used anymore, so they shouldn't be in the git repository.
 * Removing "conventions.txt": This file were incomplete and misleading as the information is maintained in the wiki
 * Removing "TODO.md": This file only repeats the obvious and there is a better version in the wiki. 
 * Removing the nightlybuilder: This set of scripts is broken since we move to git and there is no use case for them.
 * Rewriting the build guide: In current master, we have the windows instructions in the readme, linux instructions in a separate file and OS X instructions in the wiki, while the instructions in the wiki partially differ from the git repository. My plan how to fix that is to put all the build instructions in the readme and then linking from the wiki to the readme. The reason why this information should be in the git repository and not in the wiki is that the readme should tell you how to set up the project and because the readme is version controlled together with the code. If we e.g. decided to change the build system for version 1.0 and had the instructions only in the wiki, one would not know how to build version 0.9.2 anymore.

auriamg:
@hiker what do you think?

hiker:
I agree with most things:
- The svn-config should be moved to the assets svn repo
- Considering that our build instructions don't change much, I think it might be a good idea to have one set of instructions (for all three platforms) included in the src distro. What about we put the full instructions for all three platforms into the INSTALL.md file, and on our web page just point to the github version of that file? This way we have only to maintain one document

konstin:
> The svn-config should be moved to the assets svn repo

I can't move the config to the svn repo as I don't have access to it, so this pull request contains only deleting part of the moving.

> Considering that our build instructions don't change much, I think it might be a good idea to have one set of instructions (for all three platforms) included in the src distro. What about we put the full instructions for all three platforms into the INSTALL.md file, and on our web page just point to the github version of that file? This way we have only to maintain one document

That was what I proposed with the small difference that I'ld have the instructions in one conprehensive readme and not in an extra install file, but that's only a stylistic question.

Issue #2613, opened 2016-08-23
warning: comparison between signed and unsigned integer expressions


aapo:
stk-code/src/tracks/check_line.cpp:174
    stk-code/src/tracks/check_line.cpp:215
    stk-code/src/tracks/check_line.cpp:218
    warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
       if (kart_index == -1)

And indeed: 

    unsigned int kart_index

auriamg:
Yes, though in this case -1 is just a synonymous for 0xFFFFFFFF. So there's no bug, just annoying warnings. Feel free to shut the warnings by adding the proper casts if you wish though there is no real issue

nado:
Or instead of cast you can use UINT_MAX
```
#<cstdint>
...
if (kart_index == UINT_MAX)
```

nado:
Fixed in 4c0b4d80780cb35f4e36fbf8f96970c3723b0fae
Can be closed

deveee:
Define something like INVALID_KART_INDEX or change the kart index to signed integer seems to be better solution, but indeed the warning should be now fixed. Thanks!

PR #2615, opened 2016-08-24
Fixes warnings

nado:
See #2613

Edit: I did not change any of the networking code to avoid long merging headache to @hiker.

### DIFF
--- a/src/graphics/render_geometry.cpp
+++ b/src/graphics/render_geometry.cpp
@@ -1319,9 +1319,9 @@ void multidraw2ndPass(const std::vector<uint64_t> &Handles, Args... args)
     T::InstancedSecondPassShader::getInstance()->use();
     glBindVertexArray(VAOManager::getInstance()->getInstanceVAO(T::VertexType,
                                                                 T::Instance));
-    uint64_t nulltex[10] = {};
     if (SolidPassCmd::getInstance()->Size[T::MaterialType])
     {
+        uint64_t nulltex[10] = {};
         HandleExpander<typename T::InstancedSecondPassShader>::template
             Expand(nulltex, T::SecondPassTextures, Handles[0], Handles[1],
                    Handles[2]);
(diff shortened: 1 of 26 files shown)


src/graphics/render_geometry.cpp
```diff
@@ -1410,9 +1410,9 @@ void IrrDriver::renderSolidSecondPass()
                 GrassMat::InstancedSecondPassShader::getInstance()->use();
                 glBindVertexArray(VAOManager::getInstance()->getInstanceVAO(GrassMat::VertexType,
                                                                            GrassMat::Instance));
-                uint64_t nulltex[10] = {};
                 if (SolidPassCmd::getInstance()->Size[GrassMat::MaterialType])
                 {
+					uint64_t nulltex[10] = {};
```
qwertychouskie: This line looks too indented.

nado: Forgot to set expandtab :-°

src/karts/skidding.cpp
```diff
@@ -325,14 +325,13 @@ void Skidding::update(float dt, bool is_on_ground,
             m_predicted_curve->setVisible(true);
             m_predicted_curve->setPosition(m_kart->getXYZ());
             m_predicted_curve->setHeading(m_kart->getHeading());
-            float angle = kp
-                                ->getMaxSteerAngle(m_kart->getSpeed())
+            //FIXME: why successive assignations with different values ?
```
nado: Oh, right I dont intend to let this, @hiker I believe this is you who introduced the code, does it have any side effect outside of this scope ?

src/main.cpp
```diff
@@ -429,7 +429,6 @@ bool isEasterMode(int day, int month, int year, int before_after_days)
         }
         return (month > easter_start_month || (month == easter_start_month && day >= easter_start_day)) &&
                (month < easter_end_month   || (month == easter_end_month   && day <= easter_end_day));
-        break;
     }
```
hiker: I would leave the break here. Having it does not affect anything negatively, but if that that code should be rewritten later (and the return removed), that could cause some unnecessary time to debug.

LoadingPleaseWait: Is it actually necessary to use a switch statement here? It seems like it might be cleaner to just use an if statement and that would remove the need for a break.

qwertychouskie:
This should be updated, now that walldriving has been merged. ("This branch has conflicts that must be resolved")

PR #2621, opened 2016-09-01
Disable Compute Shader for ATI / Radeon GPUs on Linux

### LINKED ISSUE #2595: Compute shaders make a crash with Radeon HD 7640G and drivers version 15.302
When I set a graphics setting of 3 or higher SuperTuxKart crashes when loading a track.I am using the latest version from the stable PPA but it also happens when using the one on the website.

HP Pavilion g7 2269wm

CPU - AMD A8-4500M
GPU - AMD Radeon HD 7640G (using fglrx driver)
RAM - 6GB DDR3
OS - Windows 10 , elementary OS (currently using) , Remix OS

SuicSoft:
Compute shaders cause a segmentation fault on Linux with AMD / ATI cards with fglrx.I do not have this problem if I run STK from Windows 10 so this issue only applies the Linux :smiley:.

I don't have a Mac (with AMD graphics) to test for the segmentation fault on OS X.

Feel free to revert the PR when the compute shader crash with fglrx is fixed.This won't close #2595 since the issue is still there and compute shaders improve performance

### DIFF
--- a/data/graphical_restrictions.xml
+++ b/data/graphical_restrictions.xml
@@ -29,4 +29,6 @@
   <card contains="Radeon"         os="windows"  version="<14.300"  disable="DriverRecentEnough"/>
   <card contains="ATI"            os="windows"  version="<14.300"  disable="DriverRecentEnough"/>
   <card contains="ATI"            os="windows"  version="<=3.1.8787" disable="ForceLegacyDevice"/>
+  <card contains="Radeon" os="linux" disable="ComputeShader"/>
+  <card contains="ATI" os="linux" disable="ComputeShader"/>
 </graphical-restrictions>


deveee:
But it's not that trivial :( It for sure works fine with other Radeon graphics cards or with other fglrx drivers versions (at least it was the only one bug report). Totally disable it on linux doesn't have a sense because it works for most people.

You can make some tests with older or newer drivers version to see which particular versions cause a crash, and then we can disable it for this one particular graphics card.

SuicSoft:
@deveee I am using elementary OS freya 64-bit. It crashes when using fglrx or Catalyst from AMD website.The open source drivers show nothing but a black screen and also says that the driver is out of date

SuicSoft:
@devee Any updates ?

deveee:
@SuicSoft The pull-request in current state is not acceptable, because it disables compute shaders for all radeon graphics cards with all drivers versions on linux. 

We don't really know the reason of this crash. The fglrx 15.201 drivers are available on Ubuntu since over 2 years (version 14.04.1), so I can assume that it works fine for other people. Maybe it is related to your particular graphics card model, but we can't say it for sure.

Anyway fglrx drivers are now marked as deprecated and not available in Debian/Ubuntu testing versions anymore. IIRC fglrx drivers for xserver >= 1.18 are not available.

I think that we should leave it as is for now because it doesn't affect many people. We can try to find a reason when we will get more similar bug reports.

SuicSoft:
@devee FGLRX is unmaintained  now so I think this should be closed.

I've moved to elementary OS Loki using Liqourix kernel 4.7 and oibaf's PPA and the performance is almost the same (in STK)

PR #2622, opened 2016-09-01
[WIP] Optimize shaders for STK

SuicSoft:
I am optimizing the shaders using glsl-optimizer in STK to improve perfomance / FPS

### DIFF
--- a/data/shaders/irrlicht/COGLES2FixedPipeline.fsh
+++ b/data/shaders/irrlicht/COGLES2FixedPipeline.fsh
@@ -1,142 +1,94 @@
 precision mediump float;
-
-/* Definitions */
-
-#define Solid 0
-#define Solid2Layer 1
-#define LightMap 2
-#define DetailMap 3
-#define SphereMap 4
-#define Reflection2Layer 5
-#define TransparentAlphaChannel 6
-#define TransparentAlphaChannelRef 7
-#define TransparentVertexAlpha 8
-#define TransparentReflection2Layer 9
-
-/* Uniforms */
-
 uniform int uMaterialType;
-
 uniform bool uTextureUsage0;
-uniform bool uTextureUsage1;
-
 uniform sampler2D uTextureUnit0;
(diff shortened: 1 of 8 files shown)


Benau:
I don't think using shader optimizator worth a pr, and i doubt the code is used in stk anyway...

Benau:
It is used in gles... anyway, your fix remove comments and copyright, which is not acceptable imo

do you want to do some real improvements?

deveee:
These shaders are used in legacy pipeline in GLES renderer.

I didn't check everything, but as I see it just converted input/output parameters to newer GLSL version ("in" instead of "attribute" etc...), which btw. probably won't work when you don't have defined #version at the beginning. 

Other modifications are just style changes, for example:
```
  Color_1 = vVertexColor;
  if (uUseTexture) {
    Color_1 = (vVertexColor * texture (uTextureUnit, vTexCoord));
```
instead of
```
	vec4 Color = vVertexColor;

	if(uUseTexture)
		Color *= texture2D(uTextureUnit, vTexCoord);
```

or 
```
vVertexColor = inVertexColor.bgra;
```
instead of 
```
vVertexColor = inVertexColor.zyxw;
```
which is equal.

It also removed 
```
#define MAX_LIGHTS 2
```
which is IMO less readable now.

Anyway I think that it won't magically make it working 2x faster, and as Auria said on the forum, shaders should be manually tweaked by hand if there is something wrong.

I think that we can close it as invalid.

SuicSoft:
@deveee @Benau I will create a another PR which uses a script at compile time to optimize the shaders so the readability is not compromised

auriamg:
@SuicSoft before you work on any of that, keep in mind that we should consider carefully consider the situation before adding new dependencies. If your tool identifies some shaders that can be identified, maybe we can manually improve those shaders without adding a new dependency in our pipeline

SuicSoft:
@auriamg I'll make it optional to run the script so that the shader optimizer is optional

PR #2624, opened 2016-09-01
Provide nightly builds from Travis

qwertychouskie:
This is a PR and an issue in one.  The issue part: It would be nice to provide builds from Travis, so that it is easier for people to test the current git version.  The PR part: From https://docs.travis-ci.com/user/deployment/, don't delete the built files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,4 +62,4 @@ notifications:
       - "Build: %{build_url}"
 
 deploy:
-	skip_cleanup: true
+  skip_cleanup: true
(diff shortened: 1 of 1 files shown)


hiker:
Hi,
afaik just not deleting the files does not provide us with anything useful, we need to deploy the binaries to a server somewhere. And the main problem: without assets this is completely useless (often we have changes to assets going along with code changes). Additionally, a linux binary that does not include the right set of libraries might not work on most platforms anyway.

Thanks!
   Joerg

Issue #2625, opened 2016-09-02
Path for shaders is hardcoded


SuicSoft:
I am trying to get the shaders to be used from another directory (see https://github.com/SuicSoft/stk-code/blob/shader_optimization/sources.cmake , it also gives the **-- Using optimized GLSL shaders** message which shows that it not to do with my coding) but it does not use the new shader directory.

I have tried deleting and re adding the `cmake_build` directory and running `make clean` but it still doesn't work.The shader directory is most likely hard coded somewhere

Benau:
If you really want to have it non-hardcoded, it's your task to make it smartly choose one for your optimized GLSL shaders, not our job to do that

SuicSoft:
@Benau Then why does the cmake.sources have an option to change it.Didn't you test it when you added that code (maybe it has broken recently)

deveee:
cmake.sources doesn't have an option to change it

SuicSoft:
@deveee @Benau Then what is https://github.com/supertuxkart/stk-code/blob/master/sources.cmake#L5

PR #2626, opened 2016-09-02
[WIP] Shader optimization

SuicSoft:
This PR adds a script to the tools directory which is optional.This does not add a dependency to STK since it is not required but is recommended (and should be run for our release builds).The readability of the GLSL is the same since the optimized shaders are in a seperate directory ignored by git.

I have tested for glitches on my laptop and fixed them so currently there are no glitches but you should also test it on your computer.

This will increase your FPS by **+10** . I could not optimize all the shaders since https://github.com/aras-p/glsl-optimizer does not support OpenGL 3.3

This is not ready for merging now since it is work in progress

### DIFF
--- a/tools/optimize_shaders.sh
+++ b/tools/optimize_shaders.sh
@@ -5,10 +5,14 @@ cp -R data/shaders data/shaders_optimized # Copy shaders to data/shaders_optimiz
 
 cd data/shaders_optimized
 
-find . -type f -name "*.vert" -exec sed -i '1i  #version 140' {} \;
-glslopt -v coloredquad.vert coloredquad.vert
+find . -type f -name "*.vert" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -v $1 \;
 
-find . -type f -name "*.vert" -exec sed -i '/#version 140/d' {} \;
+find . -type f -name "*.frag" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -f $1 \;
 
-sed -i '1s/^/#version 140\n/' ./pass.vert # To fix the crash
-
+# Use unoptimized version of faulty shaders
+cp ../shaders/particlesimheightmap.vert .
+cp ../shaders/texturedquad.vert .
+cp ../shaders/pointemitter.vert .
+cp ../shaders/importance_sampling_specular.frag .
+cp ../shaders/screenquad.vert .
+cp ../shaders/colortexturequad.vert .
--- a/tools/optimize_shaders.sh
+++ b/tools/optimize_shaders.sh
@@ -5,9 +5,9 @@ cp -R data/shaders data/shaders_optimized # Copy shaders to data/shaders_optimiz
 
 cd data/shaders_optimized
 
-find . -type f -name "*.vert" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -v $1 \;
+find . -type f -name "*.vert" -exec ../../tools/optimize_glsl.sh {} -v $1 \;
 
-find . -type f -name "*.frag" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -f $1 \;
+find . -type f -name "*.frag" -exec ../../tools/optimize_glsl.sh {} -f $1 \;
 
 # Use unoptimized version of faulty shaders
 cp ../shaders/particlesimheightmap.vert .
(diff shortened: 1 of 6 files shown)


tools/optimize_shaders.sh
```diff
@@ -0,0 +1,18 @@
+#/bin/bash
+
+echo Copying data/shaders to data/shaders_optimized
+cp -R data/shaders data/shaders_optimized # Copy shaders to data/shaders_optimized
+
+cd data/shaders_optimized
+
+find . -type f -name "*.vert" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -v $1 \;
```
qwertychouskie: This should be `find . -type f -name "*.vert" -exec tools/optimize_glsl.sh {} -v $1 \;`.

tools/optimize_shaders.sh
```diff
@@ -0,0 +1,18 @@
...
+
+find . -type f -name "*.frag" -exec /mnt/storage/princ/Documents/Programming/stk-code/tools/optimize_glsl.sh {} -f $1 \;
```
qwertychouskie: This too: `find . -type f -name "*.frag" -exec tools/optimize_glsl.sh {} -f $1 \;`

SuicSoft:
#2625 Is preventing this from being mergable

Arthur-D:
My FPS increased not at all, with the zipped shaders you shared on IRC.

> Cocoa Temple 0 AI
 53 min
 67 max
 
> Optimized shaders
 53 min
 66 max

Additionally the headline font glitches a lot:
optimized_shaders_glitch

SuicSoft:
My FPS increased by +6 on that track with the optimized shaders

(Optimized)
Min - 21
Max - 34

(Normal)
Min - 17
Max - 28

Using git version of STK on HP Pavilion g7 2269wm with A8-4500M and 7640G discrete-class graphics. Graphics settings : 
screenshot from 2016-09-03 18 34 50
At 1280x720 resolution (not native 1600x900).

Which version of STK are you using and what GPU do you have and what drivers.AMD's fglrx driver is known to be _fast and buggy_ so maybe it doesn't optimize correctly.

If you have an AMD GPU using fglrx run `rm -rf ~/.AMD/GLCache` to delete the shader cache before benchmarking the difference

Arthur-D:
I'm using git version as well, under Kubuntu 16.04 with
```
 -Computer-
Processor		: 4x Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz
Memory		: 16382MB (6265MB used)
Operating System		: Ubuntu 16.04.1 LTS
-Display-
Resolution		: 1920x1080 pixels
OpenGL Renderer		: GeForce GTX 960/PCIe/SSE2
X11 Vendor		: The X.Org Foundation
```

My graphics driver is the proprietary NVIDIA 361.42 driver. I ran the game at 1280x720 resolution with these graphics settings:
optimized_shaders_gfx_settings

SuicSoft:
Your GPU is  much more powerful than mine and also NVIDIA not AMD.The fglrx driver by AMD is very buggy but fast and _just works_ and needs manual optimization.

And try this new shaders.zip [shaders.zip](https://github.com/supertuxkart/stk-code/files/453414/shaders.zip) which fixes the glitches.

Arthur-D:
Yep, fixes the glitches with headlines. However FPS still remains between 53 and 66.

SuicSoft:
@Arthur-D Your drivers are already doing optimization but mine aren't.

(Waiting for AMD to release new open source driver).

I am going to test it on Windows 10 tommorow

Benau:
I don't think anyone in the team should waste more time with it, sorry

SuicSoft:
@Benau Yes , it is a time waster to get everyone to test it and will take another week or so to get a lot of results

deveee:
I just checked your optimized shaders on Android. I mean these files:
https://drive.google.com/file/d/0B08cCnnU-zt-VldWMHU3dXcxV28/view?usp=sharing

It works fine without any artifacts, but there is no performance difference at all. I get exactly the same fps numbers.

I don't want to hurt you because you want to do something for STK, but there are only some style modifications in these optimized shaders and I don't think that it will cause any significant performace boost.

I noticed that it removed some of our definitions, for example #ifdef SRGBBindlessFix in object_unlit.frag. The "SRGBBindlessFix" is set for Radeon graphics cards, so maybe you get few more fps, but some objects in the scene are darker than before?

Or maybe our shaders are not that bad ;) Or shaders compilers are not that bad now.

You can optimize shaders when you for example avoid computing constant values once per fragment/vertex and you pass this to a shader as uniform variable. Or you use some more advanced OpenGL techniques, for example use subroutines instead of "if" conditions where it's possible etc... But I don't think that there is a tool that will do it for you.

SuicSoft:
@deveee They aren't for OpenGL ES

PR #2632, opened 2016-09-10
Fix #1929

### LINKED ISSUE #1929: Doesn't create shortcut on the desktop automatically
I have installed the game in windows 8.1 64-bit and windows 7 32-bit. It doesn't  create a shortcut on the desktop.We can do something so that it ask while installation if we want the application on our desktop.

qwertychouskie:
Create link on desktop.  Although there is no option to choose whether or not to put the icon there, it probably doesn't matter that much, as it is really easy to delete an icon from the desktop.

According to [this](http://nsis.sourceforge.net/Simple_tutorials), the $DESKTOP variable is automatically defined by NSIS.


Flakebi:
I don't like it when applications create icons on my desktop, I'm annoyed if I have to uncheck the option (and I like installers that have it unchecked by default) so I think you should at least provide an option to evade the desktop icon.

hiker:
Sorry, I dislike an app creating links on my desktop. So unless we get a checkbox (default off) that allows a user to select if the link is wanted or not, I won't change the installer.

Issue #2065, opened 2015-04-01
Add icons for different kart specs


hiker:
The bars showing the different kart specs have different sizes in split screen, most likely caused by the text in front of the first bar.

![bar-length](https://cloud.githubusercontent.com/assets/6138652/6954937/45013132-d923-11e4-96f6-84495be425a5.png)
Using icons would be better - less to translate, less space requirement.

fHachenberg:
Tried to work on this issue but it appears to be rather cumbersome to change labels in the SkillLevelWidget into icons. For instance edit move- and size-Routines have to be modified (obscure dimension calculations). I'd be happy if someone could direct me into the right direction how to approach this problem. I seem to do something wrong because this was labeled as a beginner's task ^^

auriamg:
@fHachenberg you're on the right back, I think in this case the text label part can be removed (and removing m_label_w, m_label_h, etc.). The interesting part is that once the label is removed, the size calculation should become a lot easier. We can probably presume square icons, so determining the width should now be pretty easy.
And "LabelWidget" in the constructor can be replaced by an IconButtonWidget

something like (pseudo-code)
```
icon  = new IconButtonWidget(IconButtonWidget::SCALE_MODE_KEEP_TEXTURE_ASPECT_RATIO, false);
icon->setImage("gui/example.png", IconButtonWidget::ICON_PATH_TYPE_RELATIVE);
```

fHachenberg:
I introduced icons instead of text. But there are a couple of followup issues:
* The icons seem to be rendered without anti-aliasing, while other icons on screen (main menu buttons for instance) are definitely rendered using AA
* The icons I introduced are rather dull and missing a better idea I used a horse as a symbol for power ("horse power" ^^). Replacements/Enhancements are needed
* I was not sure whether the original intent was to provide EVERY skill bar (not just for player 1) with icons but I tried that and ran into problems with widget placement/sizes. See attached screenshots. So I finally ended up with simply replacing labels for player 1 with icons. Further player do not have icons on their skill bars.
* Even good icons might not be self-explanatory. Tooltips maybe?

![screen_single](https://cloud.githubusercontent.com/assets/524304/18668968/e1461d84-7f38-11e6-9402-bfae7b4308f6.png)
![screen_multi](https://cloud.githubusercontent.com/assets/524304/18668972/e6da11f6-7f38-11e6-828d-594323ae0dd1.png)
[...]

PR #2637, opened 2016-09-20
Fixes issue #2065 (labels in kart stats bars replaced by icons)

fHachenberg:
Changes applied:

 *symbol files added
 *LabelWidget removed from SkillLevelWidget and replaced by IconButtonWidget
  Subsequently variables and method containing Label or Text renamed
  setSize method was rewritten
 *Unused property "m_label_name" removed from SkillLevelWidget
 *Substitution label->icon also introduced renamings in KartStatsWidget.
  For instance, setDisplayText is now named setDisplayIcons


auriamg:
Also see #2065

auriamg:
So, this was never merged because there were a few points missing, but after all this time I figure there won't be any more activity here so I'll just handle the remaining bits myself. Thanks for your contribution!

Issue #2358, opened 2015-10-23
Antarctica as a standalone game engine?


eaglgenes101:
I would like to make use of the Antarctica game engine in my own (FOSS) game, but there is no documentation (as far as I can tell) about using Antarctica except in the context of Super Tux Kart. Can that documentation be provided please?

auriamg:
HI, thanks for your interest. Antartica is currently not separate from STK, I'm afraid there would be quite a bit of work to separate it into a truly independent engine. We do plan to keep on improving the modularity of our code but unless we get help from new developers it's probably going to take quite a while to reach proper modularity where the engine is more independent of the game

deveee:
Vlj started some work in these repositories:
https://github.com/vlj/YAGF
https://github.com/vlj/YAGE
but now he is busy with other things.

Also someone on the forum was writing that he is working on a game based on supertuxkart code:
http://forum.freegamedev.net/viewtopic.php?f=17&t=6551
You can look what he already did.

Making standalone engine would be nice, but it's not a priority since we are developing the game and not the engine ready to use by everyone.

I'm marking this as invalid, at least for now.

Extex101:
Hello! Us in Minetest are quite worried about irrlicht dying (Breaking from being so old) see [here](https://forum.minetest.net/viewtopic.php?f=3&t=22618).
They were considering moving it to a new engine.. 
But that would require a complete reconstruction of the code and all user provided content.
As mentioned [here (comment)](https://forum.minetest.net/viewtopic.php?f=3&t=22618#p348992) we could join forces and use Antarctica in Minetest.
Although this would require the license to be changed as noted in #2381

Thank you :) Sincerely - Extex (On behalf of Minetest)

Benau:
atm we ~30-40% people who agreed to dual license stk-code, let me finish so more stuff at the moment and send other email (maybe i can add your name minetest officially there lol)

luckily most of the rest doesn't have siginificant contribution to stk

MetalMaxMX:
I would like to know. But how is Antartica working as a standalone game engine? Any progress on that front?

PR #2639, opened 2016-09-23
Integrate Irrlicht with STK source code

SuicSoft:
Since in the future Antarctica will be a stand-alone OpenGL engine (see #2358) we should move Irrlicht into the main source code folder since it is our own version and not just a library.


deveee:
Omg but as I see you just deleted irrlicht?

SuicSoft:
@deveee Irrlicht is now in https://github.com/SuicSoft/stk-code/tree/integrate_irrlicht/src/stkirrlicht.

auriamg:
I know you're well-intended, but this doesn't help us. Just because we would like to make the code modular doesn't mean we need to put it all in the same folder - in fact, in which folder resides the code is probably the least important aspect, making the code more modular would be much more important than just moving stuff around for no practical again. For future contributions, I strongly suggest you discuss with us before attempting large scale changes

PR #2647, opened 2016-10-04
Auto-install vcrun

### LINKED ISSUE #2545: Improve "Installer Microsoft VC++ runtime libraries?" dialog
When i installer STK 0.9.2 on Windows 7 i get dialog with this text:

```
Install Microsoft VC++ runtime libraries?
```

Maybe the dialog should explain to the user what its needed for and how the user would know if it is needed.
And maybe check if already installed and what what is installed and what version is being offered to be installed.
And what will happen if not choosing to install it... and maybe how to install it later if i select not to installer but want to installer later.

qwertychouskie:
Auto-install vcrun, since it is needed.
Fixes #2545

### DIFF
--- a/tools/windows_installer/supertuxkart.nsi
+++ b/tools/windows_installer/supertuxkart.nsi
@@ -206,11 +206,11 @@ SectionEnd
 
 Section -Prerequisites
   ;SetOutPath $INSTDIR\Prerequisites
-  MessageBox MB_YESNO "Install Microsoft VC++ runtime libraries?" /SD IDYES IDNO endVC
+  ;MessageBox MB_YESNO "Install Microsoft VC++ runtime libraries?" /SD IDYES IDNO endVC
     File "prerequisites\vcredist_x86.exe"
     ExecWait "$INSTDIR\prerequisites\vcredist_x86.exe /q"
-    Goto endVC
-  endVC:
+    ;Goto endVC
+  ;endVC:
 SectionEnd
 ;--------------------------------
 ;Uninstaller Section
--- a/tools/windows_installer/supertuxkart.nsi
+++ b/tools/windows_installer/supertuxkart.nsi
@@ -208,7 +208,7 @@ Section -Prerequisites
   ;SetOutPath $INSTDIR\Prerequisites
   ;MessageBox MB_YESNO "Install Microsoft VC++ runtime libraries?" /SD IDYES IDNO endVC
     File "prerequisites\vcredist_x86.exe"
-    ExecWait "$INSTDIR\prerequisites\vcredist_x86.exe /q"
+    ExecWait "$INSTDIR\prerequisites\vcredist_x86.exe /q /norestart"
     ;Goto endVC
   ;endVC:
 SectionEnd


nado:
What if someone already installed it ?

qwertychouskie:
Should be fine, different years can be installed at once, and (I think) if the same year is installed twice, if the second version is newer than the first, it is upgraded, otherwise nothing happens.

qwertychouskie:
Also, most people don't know if they have it already or not, so it isn't making things worse.

qwertychouskie:
Also, every Windows installer I have ever seen just installs it...

qwertychouskie:
This should probably be looked at for 0.9.3.  Otherwise it is very easy for the following situation to happen:

1. User runs STK installer
2. Dialog pops up: "Install Microsoft VC++ runtime libraries?"
3. User: "NO! I just want my application, not a bunch of bundled stuff! I thought FOSS was better than that..." and clicks "No"
4. User goes to launch STK and it fails, giving an error about a missing DLL.
5. User: "@$#%@%"
6. User uninstalls STK and makes hate tweet, e.g. https://twitter.com/crapintosh123/status/883080880855502849

qwertychouskie:
[This](https://msdn.microsoft.com/en-us/library/windows/desktop/ee416805(v=vs.85).aspx), although for the DirectX Redistributable, seems applicable here:

## Installation of DirectX by the Game's Installer 

The following are the best practices for adding installation of DirectX to a game's installer: 

Term | Description
-- | --
Install the redistributable components every time. | A game's installation process should install the DirectX  redistributable components during every single installation without  allowing users to opt-out of it. If you allow opting-out, then some  users will guess that they don't need it, and if they actually do, the  game will not run.
Let the DirectX installer check for optional components. | Do not assume that the latest optional components are already  installed on a system, because Windows Update and Service Packs do not  provide any of DirectX's optional components. You must install the  DirectX runtime either by running dxsetup.exe directly or calling  DirectSetup.
Set up silently. | Launch setup in silent mode so that users do not accidentally skip  updating the DirectX runtime. You can do this by launching dxsetup.exe  with the following command:                                                                                                                  path-to-redistributable\dxsetup.exe /silent                                or by calling DirectSetup and not showing any UI.
Combine EULA acceptances. | If you prompt the user to accept a EULA, then combine that with  prompting for acceptance of the DirectX EULA when installing in silent  mode so that prompting for acceptance of EULAs happens just once.  Prompting should happen before you install anything so that if user  doesn't accept, you don't end up with a failed and partial installation.
Just run dxsetup or call DirectSetup. | Because the DirectX version number doesn't  refer to anything except the core DirectX components, do not check an  installed version before running dxsetup.exe or calling DirectSetup.  Also, do not check for a file's existence to test if an optional  component is already installed, since this usually will not correctly  determine when a component exists but needs updating. However, the  DirectX setup package will quickly determine this, and perform the right  action.

auriamg:
What does DirectX have to do with any of this? STK does not use directX

qwertychouskie:
I know, most of the points apply to any redist though.

qwertychouskie:
There might be a bigger issue here, from <https://www.gnu.org/licenses/gpl-faq.en.html#WindowsRuntimeAndGPL>:

>You may link your program to these libraries, and distribute the compiled program to others. When you do this, the runtime libraries are “System Libraries” as GPLv3 defines them. That means that you don't need to worry about including their source code with the program's Corresponding Source. GPLv2 provides a similar exception in section 3.
>
>**You may not distribute these libraries in compiled DLL form with the program. To prevent unscrupulous distributors from trying to use the System Library exception as a loophole, the GPL says that libraries can only qualify as System Libraries as long as they're not distributed with the program itself. If you distribute the DLLs with the program, they won't be eligible for this exception anymore; then the only way to comply with the GPL would be to provide their source code, which you are unable to do.**

Maybe we need to switch to MinGW builds for our installers?

deveee:
I'm not sure what about distributing msvc. It's not your program, but an external installer that's just included in the package. And a lot of applications do it. I'm not a lawyer though.

But some libraries (libvpx, openal) seem statically linked with msvc (there is no dynamic dependency on msvcr) to avoid dependency on particular msvc version. And this is probably not legal...

deveee:
At least openal is correct now:
https://github.com/supertuxkart/dependencies/commit/8c4657b8e089389880d148c7dfd86cfaae28897b

deveee:
I looked at some other applications and it looks that nobody cares :P For example Blender, Freecad, Krita, Mercurial, Firefox, Audacity have msvc runtime or d3d compiler INSIDE their directories. And at least some of them are GPL.

qwertychouskie:
The thing is though, that STK still distributing the proprietary DLLs with the program, just inside of a (also proprietary) EXE.  Maybe it would be a good idea to contact the FSF for clarification?

Blender, Krita, Mercurial, and Audacity are GPLv2+ (which may have slightly different restrictions), FreeCAD is LGPLv2+ ([definitely fewer restrictions](https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License)), and Firefox is MPLv2.0.

qwertychouskie:
Another idea, maybe we could use Wine's msvcrt replacements, which are under the LGPL?

Also see this: https://news.ycombinator.com/item?id=11964589

deveee:
I updated libvpx for mingw too:
https://github.com/supertuxkart/dependencies/commit/7fa2aeb799de0256c676e854e111121cae49ad19

Maybe it could be used for windows VS build too, so that we won't statically link with msvc. @Benau are you ok with it? And also do you think that compile libvpx with default options is ok? It looks that it automatically detects CPU features (it was working on Celeron 1.7ghz when I was testing it last time), so there is probably no need to disable SSE3/4.

Benau:
use libvpx."dll" allow people to reuse the dll independently, because vpx dll available on the internet has some *gcc*.dll deps.....

And can you re-work Allow to have multiple dependencies directories for 32bit/64bit builds for master in libopenglrecorder (for turbojpeg pkgconfig)?

Benau:
btw can you make the "pause button" in android clickable after network game is finished for a player himself (but not all players finished yet)

deveee:
Ok, so I will go with even more off topic. Indeed it was working only with default kart controller. Now it behaves like "escape" key instead of player PA_PAUSE_RACE action.

For libvpx unfortunately they don't allow to compile it as shared library with mingw. Maybe it's possible to hack their build system to do it. But anyway openglrecorder will fail with different libvpx.dll, because it will detect wrong ABI version.

The pull request is already updated for turbojpeg.

And btw. I had a crash when creating a lan server. It was with two STK windows on single computer. I started a race, finished one kart and then pressed exit the race. Then I tried to create the lan server second time and it hanged and then crashed with segfault. Unfortunately no backtrace, only stdout.log
[crash.log](https://github.com/supertuxkart/stk-code/files/2164238/crash.log)
And when I did it once again, it totally hanged whole operating system. I can try to reproduce it with backtrace tomorrow.

Benau:
no such file or directory... what do you get from getCurrentExecutableLocation() second time?

deveee:
Now I can't reproduce it...

qwertychouskie:
Closing as a different GPL-compliant solution is needed.

Issue #2577, opened 2016-07-22
Unclear attribution and missing source for Boom_boom_boom.ogg


Vincent-C:
( Forwarded from Debian BTS [#832062](https://bugs.debian.org/832062) )

STK's source tarball contains the following file:
```
data/music/Boom_boom_boom.ogg
```
for which attribution is apparently unclear (the credits mention Matt Thomas as the creator, but the bug reporter seems to have credible evidence that Matt Thomas may not actually have been the creator of Boom_boom_boom.ogg?). I'll just quote the bug reporter directly:

```
Supertuxkart contains a song called "Boom boom boom" in Ogg Vorbis
format. According to the credits and debian/copyright, this song is
created by Matt Thomas, and is licensed under the GNU GPL v3 or later.
The source of the .ogg file is missing. Although what consists the
source of media assets can be unclear, I would like to point out that
there is also a module tracker version of this song, which can
reasonably considered the source, as it contains the score, the samples
used and can be editted in any module editor:

https://modarchive.org/index.php?request=view_by_moduleid&query=38184

This module contains the following text:

> -kbx128-  march 1993
> music for a crap amos
> game written by my
> brother.(only joking)

There is also some doubt that Matt Thomas and kbx128 are the same
person:

http://forum.freegamedev.net/viewtopic.php?t=6562&p=66000
```

auriamg:
This song was added to the project quite a long time ago so it's pretty unclear, but agreed we should probably just replace it

qwertychouskie:
http://forum.freegamedev.net/viewtopic.php?f=18&t=7062&p=70586#p70586

akien-mga:
> http://forum.freegamedev.net/viewtopic.php?f=18&t=7062&p=70586#p70586

Uh-oh. Given that many tracks were inspired from this tune, I think it would be worth trying to get in touch with the original author and ask if they would relicense under a STK-compatible license. Given the impact their little tune had on the game, I'm pretty sure they would agree provided we can find a proper contact address.

stephenjust:
It might be easiest to contact the author through his website or twitter: https://vimster.bandcamp.com/, @vimster

Legimet:
You can find the Boom_boom_boom.mod file in the source of the original TuxKart: http://tuxkart.sourceforge.net/downloads.html

SuicSoft:
@Legiment That page about the original TuxKart is quite interesting.

deveee:
[...]

PR #2650, opened 2016-10-08
Fix credits for Boom-boom-boom song

Legimet:
Changed to credit the actual author, Keith Baylis aka Vim. See #2577 and http://forum.freegamedev.net/viewtopic.php?f=17&t=6562.


deveee:
Merged, thanks

Issue #2649, opened 2016-10-06
Using appveyor for windows CI


nado:
Hi,

I’m playing with appveyor config to build stk on it, it is currently working for x86 VisualStudio compiler.

Would stk-team be interested in using this service to continually test windows compilation and/or provide windows regular builds ?

You can see here the discussion I am having with their twitter CM about cache configuration https://twitter.com/n4d0/status/783707739516203008
The dependencies and the assets should be cached but it doesnt work atm, we can avoid the assets download but it would be nice for providing official builds to still have them : if the caching works, its ~5 minutes the first time and should be far less after.

[My appveyor history](https://ci.appveyor.com/project/nado/stk-code)
[Appveyor.yml](https://github.com/nado/stk-code/blob/test/appveyor/.appveyor.yml)

qwertychouskie:
Now that it is working, will we see a PR soon?

nado:
It’s not working 100%, caching of the assets doesnt work because its size is too high.
I haven’t tested anything much more than appveyor itself, I guess I’ll make a PR within one week or two.

hiker:
TBH, I am not entirely sure that I would even want this. Reason is that because we are such a small team (4 coders) we are using a somewhat 'agile' model - i.e. apply updates to master before they are fully mature (so we are all testing all the time).
As a result of this our master is sometimes unstable, so if we would have many people trying to help us by testing daily builds, we would just get many more bug reports for things that we might be aware of (which means we need time to answer them, ...). And we might give a rather bad impression of the state STK is in.
Yes, if we had a group of people testing branches, another group to do a code review before merging the changes in, ... but we don't.

Being able to trigger a full installer build automatically might be useful (to allow us to easily do alpha releases), though the overhead of a release is more the testing, ticket updates, ... - building is not that much overhead.

qwertychouskie:
But what if, for example, @samuncle wanted to easily get the latest master (with walldriving) to work on the "newton" track?

Benau:
I think everyone who wants to contribute to STK should know how to compile STK themselves

nado:
Alright, and what about Win compilation testing for linux/mac users ? No releasing config, only compile and report errors/warning. Would it be interesting ?

Benau:
We don't even have time/pto maintain a coverity build

qwertychouskie:
[...]

PR #2652, opened 2016-10-14
Adding appveyor config file

nado:
Only x86 VS 2015
No notification, probably best for the first uses.

All you need to do to start using the appveyor CI is to create an account and link the GH repo to it, and select branches or PR to be built.

See #2649 for the rest of the discussion

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,57 @@
+environment:
+    DEPS_DIR: c:\\projects\dependencies
+    ASSETS_DIR: c:\\projects\stk-assets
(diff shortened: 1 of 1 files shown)


leyyin:
Can someone please review this? @hiker @auriamg

auriamg:
Hiker also created an official STK account on appveyor

hiker:
First build successful - great!! 

I wonder if it is possible to get irc notifications?  Also it appears that notifications are sent to the email address set for the account (in this case our official stk email address). Is it possible to send the notifications to the committer  instead?

hiker:
OOops, I should have added this to the ticket #2649, not to the pull request ... doing this now ;)

PR #2654, opened 2016-10-17
Remove duplicate include statements from kart.cpp

LoadingPleaseWait:
A few of the include statements in `kart.cpp` were written twice. It looks like `kart.cpp` is the only file with obvious redundant include statements like this.


deveee:
Looks fine. We are using cmath include for std::isnan, so the math.h indeed looks unnecessary. And we certainly shouldn't use both of them at the same time.

PR #2655, opened 2016-10-18
Improve SSAO performance by using swizzle mask

SuicSoft:
Just using swizzle mask instead of creating a new **vec3** in the `ssao.frag` shader.  Swizzle masks make code faster according to https://www.opengl.org/wiki/GLSL_Optimizations.

Improves FPS (with SSAO and high shadows) by +3 FPS on HP Pavilion g7  2269wm with A8-4500M and 7640G. 

Tested on Nessie's Pond track at 1600x900 (FPS is 15 - 22)

scotland-2016 10 18_12 34 29
### SSAO shader time (Cocoa Temple)
- With swizzle mask : 0.375ms
- Old : 0.487ms

### DIFF
--- a/data/shaders/ssao.frag
+++ b/data/shaders/ssao.frag
@@ -44,10 +44,8 @@ void main(void)
     vec2 offset = vec2(cos(invSamples), sin(invSamples));
 
     for(int i = 0; i < SAMPLES; ++i) {
-        float alpha = (float(i) + .5) * invSamples;
-        rotations = vec2(rotations.x * offset.x - rotations.y * offset.y, rotations.x * offset.y + rotations.y * offset.x);
-        float h = r * alpha;
-        vec2 localoffset = h * rotations;
+        rotations.xy = rotations.xy * offset.xy - rotations.yx * offset.yx;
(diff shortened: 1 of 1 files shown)


deveee:
Could you attach the reports from our profiler with and without your optimization, so that we can easily compare the difference in performance?

Also note that 
```vec2(rotations.x * offset.x - rotations.y * offset.y, rotations.x * offset.y + rotations.y * offset.x);```
is not equal to
```rotations.xy * offset.xy - rotations.yx * offset.yx;```
so it should be fixed first.

SuicSoft:
@devee I'll add the reports from the profiler. 

Also it looks the same in SSAO viz compared to the unmodified shader so do I still need to fix the equation

SuicSoft:
@deveee  Any updates ?

nado:
Any report and or fix that was asked on your side ?

qwertychouskie:
> Also it looks the same in SSAO viz compared to the unmodified shader so do I still need to fix the equation

You should probably fix it anyways. (and update the profiler time afterwards)

PR #2660, opened 2016-10-28
[WIP] Add Android compiling support

SuicSoft:
Working on getting STK to run on Android without dirty hacks (like other ports disabling audio). STK should still compile for other platforms

To compile for Android download CMake and NDK bundle using SDK manager and run something like the following command : `cmake -DCMAKE_TOOLCHAIN_FILE=[SDKPATH]/ndk-bundle/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi ..`

Currently work in progress.

Should work if you add all the dependencies to SYSROOT of the toolchain and add a native activity (i will add later). I was unable to find an Android port of libvorbisfile so I cannot test it (probably the reason other ports used dirty hacks).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ bld*/
 build*/
 cmake_build/
 cmake_build*/
+android_build/
+android_build*/
 dependencies/
 CMakeFiles/
 stk-editor/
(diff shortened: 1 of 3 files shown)


auriamg:
@deveee is already working on an android port and is fairly advanced (  http://forum.freegamedev.net/viewtopic.php?f=17&t=7122&sid=eb2a5b64b5926c5115d071726b40462c&p=71266#p70576 ), this pull request is unecessary duplication of effort and not anywhere as advanced

SuicSoft:
@auriamg Where is the source code for @devee 's port. If it is working we should consider adding it to our git

qwertychouskie:
@SuicSoft According to the forum post, it will be merged soon.

PR #2723, opened 2017-01-04
Convert CREDITS to UTF-8

jcowgill:
This converts the CREDITS file to UTF-8 and adjusts the code accordingly (each line is converted early on back into UTF-16). It's just a lot easier to manage the CREDITS file this way and various tools can understand it without any special config options. This was motivated by a request to patch the CREDITS file in Debian which isn't very easy at the moment.


auriamg:
Merged, thanks!

PR #2731, opened 2017-01-08
Update CIrrDeviceLinux.cpp

amoibos:
Fix compile error under NetBSD


auriamg:
@amoibos Maybe you could combine all your NetBSD patches into a single pull request instead of opening a lot of small ones?

PR #2732, opened 2017-01-11
Change "Geometry detail" to "Level of Detail", small tweaks

qwertychouskie:
Changed for clarity as it controls whether extra objects are shown or not.  Also changed a message to make it accurate regarding current behavior.

### DIFF
--- a/data/gui/custom_video_settings.stkgui
+++ b/data/gui/custom_video_settings.stkgui
@@ -18,7 +18,7 @@
         <div layout="horizontal-row" width="100%" proportion="1">
             <checkbox id="dynamiclight"/>
             <spacer width="10" height="10"/>
-            <label text="Advanced pipeline (lights, etc.)" I18N="Video settings"/>
+            <label text="Advanced pipeline (lights, etc.)*" I18N="Video settings"/>
         </div>
 
         <spacer height="4" width="10" />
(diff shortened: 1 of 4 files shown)


Benau:
Please don't change our copyright header, your PR will be reject if you do so

MTres19:
@qwertychouskie Geometry detail level and LOD aren't the same thing. LOD changes the geometry based on distance from the camera; geometry detail level restricts it to a certain level based on the user's config. They're completely different things and we don't want confusion.

samuncle:
Hi, as one of the dev who had the idea of Geometry Detail I specifficaly choose a different name because they are two different concepts:

1) Level of detail = An object that's displayed depending on the distance to the player
2) Geometry detail = An object is load or not depending on the user setting

It avoid confusion when you code and also when you create tracks

qwertychouskie:
When I think of "Geometry detail", I think of how low poly/high poly.  Maybe something like "Scenery level"?

MTres19:
That IS more descriptive.

qwertychouskie:
New commit pushed, should be fixed.

auriamg:
"Level of scenery" is not clear at all, "Geometry detail" seems much clearer to me. In any case this is all buried in advanced options. I don't think this PR is an improvement. Next time, please check with us before creating a pull request that we actually _want_ the terminology changed.

PR #2736, opened 2017-01-14
Fix clang compilation

nado:
Hi, this is some very minor change but it fixes a compilation bug when using this configuration :
```
CC=clang CXX=clang++ \
CFLAGS="${CFLAGS} -flto=thin" \
CXXFLAGS="${CXXFLAGS} -flto=thin"\
LDFLAGS="${LDFLAGS} -flto=thin" \
cmake -D_CMAKE_TOOLCHAIN_PREFIX=llvm- \
-DUSE_FRIBIDI=0 -DUSE_WIIUSE=0 -DCMAKE_BUILD_TYPE=Debug ..
```

The linker used was ld.gold and I had to increase my tmpfs used by clang from ~750MB to ~1GB for all the objects to be stored during link time.

### DIFF
--- a/src/utils/time.hpp
+++ b/src/utils/time.hpp
@@ -26,13 +26,13 @@
 #ifdef WIN32
 #  define WIN32_LEAN_AND_MEAN
 #  include <windows.h>
-#  include <time.h>
 #else
 #  include <stdint.h>
 #  include <sys/time.h>
 #  include <unistd.h>
 #endif
 
+#include <time.h>
 #include <string>
 #include <stdio.h>
 


deveee:
Is this on linux? And what is the error that you get?

I don't really understand this fix, because now you declare both time.h and sys/time.h for non-windows operating systems. And at the beginning of sys/time.h header there is #include <time.h>, so it should be included anyway. 

Include both sys/time.h and time.h will probably work, but I'm not 100% sure that it's available on all non-windows platforms.

nado:
time.h is part of the C library. sys/time.h defines some struct/functions used in this file.
When using clang -flto, it could not find `difftime` because time.h wasnt included.

I guess this comes from the use of libcxx instead of gcc’s libc. I saw no harm to always include time.h and it built fine with gcc.

deveee:
Ok, indeed time.h should be always included, because it provides the difftime function that we use.

PR #2759, opened 2017-01-29
remove dt dependencie on camera position

rogue-spectre:
Hi, 

When i play with all graphics settings to maximum, the camera distance to the kart is not stable. It seems like a loss of frame rate, but it's not. In fact the timestep is fluctuating and since the position of the camera is depending on it. So i made a modification in order to remove the dt dependency and only use the speed.

### DIFF
--- a/src/graphics/camera_normal.cpp
+++ b/src/graphics/camera_normal.cpp
@@ -91,8 +91,8 @@ void CameraNormal::smoothMoveCamera(float dt)
     float ratio = (current_speed - max_speed_without_zipper) / max_increase_with_zipper;
     ratio = ratio > -0.12f ? ratio : -0.12f;
 
-    // distance of camera from kart in x and z plane
-    float camera_distance = -3 * (0.5f + ratio);
+    // distance of camera from kart in x and z plane    
+    float camera_distance = -3.0f * (0.5f + ratio) - fabsf(current_speed)/7.0f;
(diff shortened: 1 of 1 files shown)


auriamg:
it probably does remove the shaking of the camera, though the camera is a bit less smooth and is more likely to be affected, for instance, by little kart movements like polygon transitions that aren't fully smooth, or when hitting an obstacle. we need to discuss this before merging... ( @hiker ) I've also had a few cases where the camera shakes when skidding (if the physical body of the kart is shaking as happens sometimes, then your camera will also shake)

the camera is a recurring issue but I'm not sure we can go with this kind of solution because it has other issues

auriamg:
I just created a new branch called NewSmoothCanera ( https://github.com/supertuxkart/stk-code/tree/NewSmoothCamera ), can you try that? It should hopefully fix issues, as I finally underdstood (I hope) what was wrong with our smoothing code

auriamg:
The branch has been merged, everyone reports it's smooth now, this supercedes this PR. Thanks for bringing this to our attention though, your PR ultimately gave me the idea

qwertychouskie:
Now the camera has a rotational gitter.  Also drifting feels a bit weird.

rogue-spectre:
Cool, it seems smoother for me.

deveee:
@auriamg I just checked it under cygwin with software rendering to make sure that fps is non-stable enough and it looks that the camera is super-smooth now :)

Issue #1699, opened 2014-11-12
Improving GP points distribution


Alayan-stk:
The current version define how much difference in points there is between positions in a given range (e.g. ; with the actual stk_config, first and 2nd have 3 points more than the next ; 3rd and 4th 2 points more than the next, and then 1)
That system is really limited. It gives ugly numbers with very few karts, but more importantly there is no way to tweak the values to have something that works well with all kart numbers. Atm, with 20 karts, doing 5th and 20th is rewarded the same way than 12th and 13th which is wrong (only the very few better place have a decent extra). However something working with 20 would likely be bad with 10.

So the deal is more or less to limit the tweaking possibilities in stk_config to be able to have a good default system adapted for all values. Having raw data for each possible number of karts would be overkill imo.

I did a table with the goal of promoting good performance. e.g. ; 5th and 20th rewards one point more than 10th + 11th and 50% more than 12th + 13th so the 5th place has a true bonus. 1st/2nd seems to be a bit less helped than before but that is mostly a contrast effect.
The table is adapted for all kart numbers and could even be expanded up to 30 karts without issue. Simple formulas allow to describe it.
![image](https://cloud.githubusercontent.com/assets/9522948/5019457/124da7e2-6ac0-11e4-8a0e-484cd9748373.png)

Alayan-stk-2:
Fixed with #2765

PR #2765, opened 2017-02-04
New GP point system

Alayan-stk-2:
New GP point system as discussed on the IRC with @auriamg & @Arthur-D.

Compiled & tested.

It adds a compilation warning about a comparison between an unsigned int and a signed int. It shouldn't have any consequence, and I'm not sure how to solve the warning. For info, it's on the line 227 of the .cpp.

Tested, and everything works. The point are attributed correctly according to what is in the config, it also tolerates the case where there is more karts than points nodes (it defaults to a zero difference for the supernumerary karts, but the config supports up to 30 before defaulting to it).

I used my own formula for the scoring, but the new system is more powerful - even if one would prefer to tweak the specifics differently, he would have more options.

Commit fixing #1699

### DIFF
--- a/src/config/stk_config.cpp
+++ b/src/config/stk_config.cpp
@@ -103,7 +103,7 @@ void STKConfig::load(const std::string &filename)
                    strA,filename.c_str());              \
     }
 
-    if(m_score_increase.size()==0 || (int)m_score_increase.size()!=m_max_karts)
+    if(m_score_increase.size()==0)
     {
         Log::fatal("StkConfig", "Not or not enough scores defined in stk_config");
     }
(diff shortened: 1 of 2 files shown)


src/config/stk_config.cpp
```diff
@@ -391,15 +394,27 @@ void STKConfig::getAllData(const XMLNode * root)
...
+
+    std::sort (sorted_score_increase.begin(), sorted_score_increase.end());
+
+    (*all_scores)[num_karts-1] = sorted_score_increase[0];  // last position score
 
     // Must be signed, in case that num_karts==1
     for(int i=num_karts-2; i>=0; i--)
     {
-        (*all_scores)[i] = (*all_scores)[i+1] + m_score_increase[i];
+        (*all_scores)[i] = (*all_scores)[i+1] + sorted_score_increase[num_karts-i];
```
auriamg: I'm a bit confused about this part, why make "(*all_scores)[i]" equal to "(*all_scores)[i+1]" when it's going to be overwritten in the next iteration of the for loop?

Alayan-stk-2: It's a decreasing loop ; i-- and not i++. As you can see, I kept the original code there.

In a decreasing loop, it won't be overwritten.

auriamg:
@hiker I'd like your feedback on this one, agree to merge? One thing I like about this PR is that it moves all GP points to a config file, so even if we eventually find we can improve the point distribution, it becomes easier to tweak. I'd be in favor of merging

PR #2768, opened 2017-02-05
Parachute improvements

Alayan-stk-2:
Create two new characteristics - durationRankMult and durationSpeedMult.

The first one is used when the parachutes is activated by an item. It makes the duration higher depending on the rank : max effect on the 1st, slightly less on the 2nd, etc. and no at all on the last kart affected. The value I use in kart_characteristics.xml is of 1.35 ; so the feature is used but keep a moderate effect (a MK-thunder-like value would have been something like 5 or 10 with a lower base duration)

The second one is used in all cases. It makes parachutes less frustrating at low speed and more penalizing at a high speed.

Also, the tweaking of the config done to integrate the new features makes a bit harder to lose the parachute by braking - it may be worth it at a low speed or for a kart ranked 1st taking the parachute at full speed ; but it's not a no-brainer. Especially, except in accidents, it recquires a conscious braking decision to have this factor play. So, these who loved the "strategic decision" should love it because it truly exists now.

The rebalancing in the xml may be further improved, however.

Compiled and test, the new effect works as intended and no new issue has been spotted.

Note this branch also contains the GP modification submited in #2765, because I did them on the master of my fork (I'm learning github...).

### DIFF
--- a/data/stk_config.xml
+++ b/data/stk_config.xml
@@ -16,51 +16,12 @@
 
   <!-- Scores are the number of points given when the race ends. -->
   <grand-prix>
-  <!-- Establish the distribution of points in GP.
-
-       For a race of N karts ; the N-first point values are taken.
-       Then, they are sorted. E.g. ; 0 1 2 1 3 2 becomes 0 1 1 2 2 3.
-       Then these numbers are used to establish the DIFFERENCE of points
-       between consecutive karts.
-
-       The smaller of the numbers is used to establish the score for the
-       last kart and not the difference between 2 karts.
-
-       In the above example, the last kart will have 0 point, the one before
-       before 1 (0+1) ; the one before 2 (0+1+1), the one before 4 (0+1+1+2),
-       etc. until the 1st which have 0 (0+1+1+2+2+3)
-
-       There shall be at least as much points nodes as max-numbers kart -->
-    <points points="0" /> <!-- added with 1 kart, score for the last kart -->
-    <points points="1" /> <!-- added with 2 karts -->
-    <points points="1" /> <!-- added with 3 karts -->
-    <points points="2" /> <!-- added with 4 karts -->
-    <points points="2" /> <!-- added with 5 karts -->
-    <points points="1" /> <!-- added with 6 karts -->
-    <points points="3" /> <!-- added with 7 karts -->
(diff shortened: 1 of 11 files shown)


auriamg:
Hi, your pull request appears to mix together parachute and GP point commits. Do you think you could develop them in separate branches and create separate pull requests for them, for the sake of clarity and ease of code review?

src/config/stk_config.cpp
```diff
@@ -391,15 +394,27 @@ void STKConfig::getAllData(const XMLNode * root)
...
+
+    std::sort (sorted_score_increase.begin(), sorted_score_increase.end());
+
+    (*all_scores)[num_karts-1] = sorted_score_increase[0];  // last position score
 
     // Must be signed, in case that num_karts==1
     for(int i=num_karts-2; i>=0; i--)
     {
-        (*all_scores)[i] = (*all_scores)[i+1] + m_score_increase[i];
+        (*all_scores)[i] = (*all_scores)[i+1] + sorted_score_increase[num_karts-i];
```
auriamg: I'm a bit confused about this part, why make "(*all_scores)[i]" equal to "(*all_scores)[i+1]"m when it's going to be overwritten in the next iteration of the for loop?

data/stk_config.xml
```diff
@@ -16,12 +16,51 @@
...
+    <points points="5" /> <!-- added with 21 karts -->
+    <points points="2" /> <!-- added with 22 karts -->
+    <points points="1" /> <!-- added with 23 karts -->
+    <points points="3" /> <!-- added with 24 karts -->
+    <points points="4" /> <!-- added with 25 karts -->
+    <points points="1" /> <!-- added with 26 karts -->
+    <points points="2" /> <!-- added with 27 karts -->
+    <points points="1" /> <!-- added with 28 karts -->
+    <points points="3" /> <!-- added with 29 karts -->
+    <points points="5" /> <!-- added with 30 karts -->
```
auriamg: This code about GP points should not be part of this PR

src/config/stk_config.cpp
auriamg: Oh, and this is also part of the wrong PR, this has nothing to do with parachutes right?

PR #2771, opened 2017-02-05
Add feedback for skid bonus

rogue-spectre:


src/karts/skidding.cpp
```diff
@@ -443,6 +443,9 @@ void Skidding::update(float dt, bool is_on_ground,
                                              bonus_speed, bonus_speed,
                                              bonus_force, bonus_time,
                                              /*fade-out-time*/ 1.0f);
+                    // Play custom character sound (weee!)
+                    m_kart->playCustomSFX(SFXManager::CUSTOM_SKID);
+                    m_kart->m_controller->handleZipper(true);
```
auriamg: If the goal of this PR is only to play a sound, then "handleZipper" should definitely not be called, a that may have side effects (such as notifying the server that you used a zipper, in multiplayer mode, which is definitely not intended)

rogue-spectre: The goal is to play the sound + have the blur effect. If i get it right i must do a copy of 
void LocalPlayerController::handleZipper(bool play_sound)  in src/karts/controller/local_player_controller.cpp
void PlayerController::handleZipper(bool play_sound) in src/karts/controller/player_controller.cpp
and customise them ? in order to have a more independent effect

auriamg: There's multiple ways to do this. One way could be to have a separate method showZipperVisualEffects() that would be called both by handleZipper and your new method, for instance, that would handle the graphical effects only. There are probably other ways, this is just an idea

PR #2799, opened 2017-02-26
Remove #ifdef Explicit_Attrib_Location_Usable in vertex shaders

Elderme:
I have simplified vertex shaders. Instead of for example:
```
#ifdef Explicit_Attrib_Location_Usable
layout(location = 0) in vec3 Position;
#else
in vec3 Position;
#endif
```

We can just write
`layout(location = 0) in vec3 Position;`

And when GL_ARB_explicit_attrib_location extension is not supported, the line is modified (remove "layout(location = 0)") when the shader file is loaded (in ShaderFilesManager class).

### DIFF
--- a/data/shaders/billboard.vert
+++ b/data/shaders/billboard.vert
@@ -2,13 +2,8 @@ uniform mat4 ModelViewMatrix;
 uniform vec3 Position;
 uniform vec2 Size;
 
-#ifdef Explicit_Attrib_Location_Usable
 layout(location = 0) in vec2 Corner;
 layout(location = 3) in vec2 Texcoord;
-#else
-in vec2 Corner;
-in vec2 Texcoord;
-#endif
 
 out vec2 uv;
 
--- a/data/shaders/coloredquad.vert
+++ b/data/shaders/coloredquad.vert
@@ -1,12 +1,7 @@
 uniform vec2 center;
 uniform vec2 size;
 
-#ifdef Explicit_Attrib_Location_Usable
 layout(location = 0) in vec2 Position;
-#else
-in vec2 Position;
-#endif
-
 
 void main()
 {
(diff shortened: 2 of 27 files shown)


Benau:
Regex seems only available to gcc 4.9 (needs further checking), though if these changes make your further refactoring easier, i'd say we will do a buildbot update.

@deveee what do you think

Benau:
(confirmed 4.9)

... Or can you think a way without using regex?

Elderme:
My gcc version is too recent, I missed that point!

Yes, these changes would make the refactoring a little easier. I think I can do the refactoring without this branch, but shaders will have imbricated "#ifdef" and that makes shaders difficult to write and to read.

I used a regular expression because of the whitespaces. For example "layout(location = 0) in vec3 Position;" or "layout (location=0) in vec3 Position;" are both valid glsl code. So we need to replace both by "in vec3 Position" when GL_ARB_explicit_attrib_location is not supported. I can try to replace the regex with string comparison (like we do for "#stk_include").

deveee:
Using layout(location = x) without #ifdef is not clear because it suggests that explicit attrib location is always used and that the code requires this extension. I don't care that much about it because I know how it works, but it may be not clear for other people.

Much more important for me is gcc 4.7 support. I use it for creating linux package and I prefer not doing any revolution in the build system because I'm busy with other things recently.

Btw. in some shaders eg. pointlight.frag there are layout(location = x) for out variables. They are needed for GLES renderer. Maybe it could be simplified in the same way.

Benau:
@Elderme we will move to non-template shader soon and required explict attribute location
Example:
https://github.com/supertuxkart/stk-code/blob/fix_2885/src/graphics/sp/sp_base.cpp#L332

If you have time, this should rather easy to make the loadShaders() into an xml version.

Issue #2735, opened 2017-01-12
Basketball doesn't hit immediately although it catches up


tobbi:
I have experienced a weird bug where the basketball doesn't hit as soon as he catches up with you but takes quite a while to hit you.

I have screencasted it. Relevant time frame: https://youtu.be/Anm0uo9pu3E?t=1m45s

Looks and feels quite weird.

qwertychouskie:
Been meaning to report this... you beat me to it. :)  Also, sometimes I can avoid the ball long enough it gives up.

P.S. You should try SimpleScreenRecorder's OpenGL capture.  FPS similar to normal play.

qwertychouskie:
In that video, the effect of the swatter was also delayed, another bug?

deveee:
Just note that it happens for me quite often

deveee:
One strange basketball behaviour is reproducible in this place:
![zrzut ekranu z 2017-08-17 22-01-28](https://user-images.githubusercontent.com/5108558/29431350-ebcad4e8-8397-11e7-8082-b702dd50c8bf.png)
though the reason may be different.

deveee:
Basketball in hacienda does very strange things. It for example sometimes goes forwards, then backwards for a few seconds, and then forwards again. Or goes somewhere outside of the track (something like on the screenshot).

I noticed that ```ball_distance = getDistanceFromStart()``` sometimes gives wrong values (something around 600 at the beginning of track, so it's probably (total_track_length - real_ball_position). In this case target is lost and this may be the reason that basket ball doesn't hit the kart.

Though I don't know yet why basketball changes direction.

All other values seem to look reasonable...

hiker:
Strange things are indeed happening, I am able to reproduce most of those issues, so I'll have a look.

hiker:
I've seen various problems:
- in sandtrack the ball would sometimes be under the drivelines (which were too high there). Then the find track sector function would find a potentially quite distant different sector (which fulfills the height condition), and the ball would move towards that point - quite possibly backwards or sidewards. After one frame (since the ball is jumping) it would find the original track sector again and then move in the right direction again. Auria fixed the drivelines which solved that problem. 
[...]

PR #2804, opened 2017-03-03
Speed up basketball

### LINKED ISSUE #2783: The basket ball is too slow
Currently, the basket ball is often too slow to catch up to the first kart, at least in a timely fashion.

The set speed is of 35. The max_speed of most karts is at 25. They may go faster with drifting and nitro, but let's assume the average speed is at 25. (NB : in Super Tux mode. I think the basket ball is not slower in easier difficulties while the karts are.)

This may seem like a noticeable difference, except it means that to catch up the basket ball need 3.5x the time separating the user of the ball and the first kart (1/(1-25/35)).

So when the user is 10 seconds behind the first, it takes 35 second for the basket ball to hit.

**When the user of the ball is 20 seconds behind, it takes 70 seconds for the basket ball to hit.**

(Side issue : when there are a lot of karts, those in bad positions shouldn't get a basket ball, it's useless for them and is ineffective even to slowdown the first)

If the first is able to drift correctly and get some nitro ; and he has no other accident ; he may well do half a race hunted by a basket ball without being hit. E.g. ; **on the lightouse track, I'm able to go faster than the basket ball.**

Increasing the basketball speed seems therefore a good idea - 40 would mean 2.7x the time at 25 and 4x at 30 ; 45 would mean 2.25x the time at 25 and 3x the time at 30. Both would still let some room to get to the finish lap in the end of a race or to get a bubblegum shield.

Another possible solution  would be to change the speed of the basketball according to its distance to the 1st kart - fast at the start but slower when it's closer. This would keep some fun trying to outrun the basket ball but making it harder.

qwertychouskie:
As proposed in #2783.  Closes #2783, #2735.

### DIFF
--- a/data/powerup.xml
+++ b/data/powerup.xml
@@ -56,7 +56,7 @@
                     tunneling.
    -->
   <item name="rubber-ball"    icon="rubber_ball-icon.png"
-        model="rubber_ball.b3d" speed="35.0" 
+        model="rubber_ball.b3d" speed="45.0" 
         scale="1 1 1"         interval="1"
         max-height="4.0"      min-height="0"
         fast-ping-distance="50" 


qwertychouskie:
Closing as this really needs to be set per difficulty level.

PR #2820, opened 2017-03-22
Add feedback for skid bonus (Bis)

rogue-spectre:


src/karts/kart.cpp
```diff
@@ -1858,8 +1858,19 @@ void Kart::handleMaterialGFX()
...
 
 //-----------------------------------------------------------------------------
+/** Display visual effect and play sound if needed for zipper or skid feedback
+* \param play_sound If true this will cause a sfx to be played even if the
+*                  terrain hasn't changed. It is used by the zipper powerup.
+*/
+void Kart::showZipperVisualEffects(bool play_sound)
+{
+    playCustomSFX(SFXManager::CUSTOM_ZIPPER);
+    m_controller->handleZipper(play_sound);
```
auriamg: Still the same issue as last PR, showZipperVisualEffects should call handleZipper. Rather, the parts of handleZippers that are visual effects should be moved into showZipperVisualEffects, and then handleZipper should call showZipperVisualEffects.

Alayan-stk-2:
No answer or change to make it work as asked in the review in more than a year, closing.

Issue #2302, opened 2015-08-30
Make FPS unlimited on title screen


MarkkusCraft123:
In my opinion, the STK menus look better when the fps is not capped.  If you want to see what I mean, you can do it by turning on the artist debug mode and on the menu screen you can select do not limit fps.

deveee:
In my opinion it doesn't matter at all. What's the difference that you see?

auriamg:
Also there is little use in having the game use 100% CPU+GPU at a ludicrius framerate (menus will easily go to 200-300 FPS without cap), when out of the game, or when paused. I see little use

MarkkusCraft123:
Lock at 60? I am a frame rate junky and the I can't stand the current fps.

deveee:
I think that human's eye can't see 60 fps and everything is in your mind :-P

The problem isn't that fps is too low, but that it doesn't have constant value. The counter shows you that averaged fps is 29.9, but in fact you have for example 30.0, 30.0, 30.0, 1.0, 30.0 etc... And the game is not smooth because of this.

And still 30 fps is certainly enough for gui with few buttons and simple animations. It's not a real game where you need higher responsivity.

I'm closing this as invalid.

deveee:
I prepared for you an example:

```

MarkkusCraft123:
How about this:

<code>const double wanted_fps = 60.0;
const int peak_frequency = 100; // once per 100 frames
const int peak_duration = 0; // 0 ms</code>

leyyin:
@MarkkusCraft123 I am curios, what is your monitor refresh rate?

MarkkusCraft123:
60 Hz

deveee:
Just to clarify, if I use nvidia graphics card, then indeed it looks a bit
smoother. But it's because nvidia driver automatically locks it at
precisely 60 fps (adaptive vsync?). And this is advantage of vsync
technology, not of higher fps. With disabled vsync I see no difference.

MarkkusCraft123:
I use an Intel HD 3000 now
, I got an i3.

qwertychouskie:
On my system (Intel HD 3000, laptop's built-in screen at 60 Hz), when I use the Artist Debug Menu to un-limit the FPS, the menu looks way smoother.  The menu, when not limited, runs at 60 FPS.

[...]

PR #2823, opened 2017-03-27
Set menu FPS to 60

qwertychouskie:
Smoother, see https://github.com/supertuxkart/stk-code/issues/2302.


MarkkusCraft:
Thanks dude.

SuicSoft:
What about setting it to the monitor refresh rate since some are not 60hz (75,120,etc).Or maybe just always enable vsync on the menu

PR #2824, opened 2017-03-27
Update string

qwertychouskie:
It hasn't been called stkaddons since the 0.8 days.


hiker:
I don't think we should put the URL in there. If we ever should decide to change it, we need to change the code again. Better leave it the way it was originally, or at least something neutral like "SuperTuxKart Addon Server"  or so

cap44:
@hiker what do you think of this string: "Connection failed, unable to contact SuperTuxKart add-ons server" for both files?

qwertychouskie:
Updated, now says "Failed to connect to the SuperTuxKart add-ons server."

PR #2833, opened 2017-04-07
Update .appveyor.yml to correct recent build failures

cap44:
Attempting to correct recent AppVeyor build failures, error: Get-ChildItem : Cannot find path 'C:\projects\dependencies' because it does not exist.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,5 @@
 environment:
-    DEPS_DIR: c:\\projects\dependencies
+    DEPS_DIR: c:\\projects\stk-code\dependencies
     ASSETS_DIR: c:\\projects\stk-assets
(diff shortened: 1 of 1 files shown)


auriamg:
It doesn't seem to be working:

 Cannot find path 'C:\projects\stk-code\dependencies' because it does not exist

nado:
You changed the variable which indicates where to clone the dependencies repository,.
The correct subdirectory (when i wrote the appveyor file, I did it only for one architecture but) is then copied in stk repository under the name of 'dependencies'.

The reason it fails seems to be because the branch fetched doesnt exist anymore in the dependencies repository. See https://github.com/cap44/stk-code/blob/7157211aa5475bca10ebe599ff31b6b379918462/.appveyor.yml#L30

Benau:
It was fixed

PR #2843, opened 2017-04-22
Changed "macOS X" to "macOS"

Property404:
No such system stylized as "macOS X."

Left "OS X Install" because I don't think Apple uses install DVD's anymore since renaming as macOS?


cap44:
I would keep the original string because of legacy support (10.9, 10.10 and
10.11)

auriamg:
I honestly don't really care... but alright

PR #2854, opened 2017-06-09
Fix a few compiler warnings

qwertychouskie:


Benau:
Assert is removed in release build, so please find a way to preserve those line in release build

qwertychouskie:
Those vars are not used except for the assert, so the compilation gives an unused variable warning when compiling in release mode.  This PR fixes those warnings by consolidating it into the assert, so it behaves the same for a debug mode build, but doesn't have the unused var for the release build.

deveee:
It looks a bit more ugly, but #ifdef DEBUG will look worse. And we most probably don't need to check framebuffer status in release mode if we don't use the result (I didn't test, but it should work without glCheckFramebufferStatus). @Benau can we merge?

auriamg:
As far as I can tell, glCheckFramebufferStatus has no side effect other than returning a value, so if indeed in release mode we don't use the return value then this should be a safe change

PR #2857, opened 2017-06-15
Tell users to report issues on issue tracker

qwertychouskie:
Bug reports in the forum tend to get buried and lost.


auriamg:
Things are just fine the way they are, we monitor the forum and we'll open a ticket if appropriate.

PR #2858, opened 2017-06-16
Reorganization of graphics options

qwertychouskie:
- Allow manual control of advanced settings in advanced settings dialog
 - Put option for scale_rtts above resolution switcher
 - Clarification of "Geometry detail"
 - More tweaks


auriamg:
The current settings are just good enough IMO, we don't need EVERYTHING to be configurable in the UI, the current options are just flexible enough IMO. power users who want to tweak every single setting (< 1% of ours users) can use the config file, no need to annoy everyone else with advanced settings they don't understand. Your popup with more options is also probably not going to look good on lower resolutions.
In the future, please check with us whether we even _want_ some change before working on it, there's a reason why we combined many settings in a simpler "image quality" setting .

PR #2861, opened 2017-06-17
Fix the --cutscene=<cutscene> commandline flag

MTres19:
If the menu is shown, there's this ugly purple background (the OpenGL default, I think) and the cutscene won't play. It doesn't make sense to show the menu anyway if the user wants to immediately play a cutscene.


auriamg:
Thanks!

PR #2863, opened 2017-06-20
Tweak graphics options

qwertychouskie:
Changes based on feedback on https://github.com/supertuxkart/stk-code/pull/2858.

 - Put option for scale_rtts above resolution switcher, as it is a "replacement" for changing the resolution, especially on Android
 - Clarification of "Geometry detail"
 - Change header of advanced graphics settings dialog from "Graphics Settings" to "Advanced Graphics Settings"
 - Add * to Texture Compression checkbox to indicate changes to it will not take effect until restarting STK (Took me a long time to figure this out myself)

### DIFF
--- a/data/gui/custom_video_settings.stkgui
+++ b/data/gui/custom_video_settings.stkgui
@@ -152,25 +152,25 @@
         <spacer height="20" width="10" />
 
         <div layout="horizontal-row" width="100%" proportion="1">
-            <label text="Animated Characters" I18N="Video settings" width="40%"/>
+            <label text="Animated Characters" I18N="Video settings" width="50%"/>
             <spacer width="10" height="10"/>
-            <gauge id="steering_animations" min_value="0" max_value="2" width="50%" />
+            <gauge id="steering_animations" min_value="0" max_value="2" width="45%" />
         </div>
 
         <spacer height="4" width="10" />
 
         <div layout="horizontal-row" width="100%" proportion="1">
-            <label text="Rendered image quality" I18N="Video settings" width="40%"/>
+            <label text="Rendered image quality" I18N="Video settings" width="50%"/>
             <spacer width="10" height="10"/>
-            <gauge id="image_quality" min_value="0" max_value="3" width="50%" />
+            <gauge id="image_quality" min_value="0" max_value="3" width="45%" />
         </div>
 
         <spacer height="4" width="10" />
 
         <div layout="horizontal-row" width="100%" proportion="1">
-            <label text="Amount of scenery (Geometry detail)" I18N="Video settings" width="40%"/>
+            <label text="Amount of scenery (Geometry detail)" I18N="Video settings" width="50%"/>
             <spacer width="10" height="10"/>
-            <gauge id="geometry_detail" min_value="0" max_value="2" width="50%" />
+            <gauge id="geometry_detail" min_value="0" max_value="2" width="45%" />
         </div>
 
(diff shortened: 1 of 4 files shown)


auriamg:
Rendered image scale has no business being there since the vast majority of our users should have it at 100%. If it's useful on android (devee should confirm) then maybe it could be on some android-specific screen but has IMO nothing to do in the PC version. We already got presets that cover that

deveee:
The problem with scale_rtts is that it works only when rtts are used (for advanced lightihg). It's possible to use rtts with non-advanced lighting, but in this case it's only faster for very low scale_rtts values. Most android devices probably need disabled advanced lighting for smooth playing, so IMO it doesn't have a sense...

But... I recently noticed that it should be possible to scale the window with ANativeWindow_setBuffersGeometry. But I didn't test it yet.

deveee:
A quick hack that allows to scale resolution from 1920x1080 to 1280x720:
```
diff --git a/lib/irrlicht/source/Irrlicht/CContextEGL.cpp b/lib/irrlicht/source/Irrlicht/CContextEGL.cpp
index 42783dfd05..f615eb6971 100644
--- a/lib/irrlicht/source/Irrlicht/CContextEGL.cpp
+++ b/lib/irrlicht/source/Irrlicht/CContextEGL.cpp
@@ -232,7 +232,7 @@ bool ContextManagerEGL::chooseConfig()
     EGLint format = 0;
     eglGetConfigAttrib(m_egl_display, m_egl_config, EGL_NATIVE_VISUAL_ID,
                        &format);
-    ANativeWindow_setBuffersGeometry(m_egl_window, 0, 0, format);
+    ANativeWindow_setBuffersGeometry(m_egl_window, 1280, 720, format);
 #endif
 
     return true;
diff --git a/lib/irrlicht/source/Irrlicht/CIrrDeviceAndroid.cpp b/lib/irrlicht/source/Irrlicht/CIrrDeviceAndroid.cpp
index c5e9d00a98..2b5f89a297 100644
--- a/lib/irrlicht/source/Irrlicht/CIrrDeviceAndroid.cpp
+++ b/lib/irrlicht/source/Irrlicht/CIrrDeviceAndroid.cpp
@@ -113,8 +113,8 @@ void CIrrDeviceAndroid::createVideoModeList()
 	
 	if (width > 0 && height > 0)
 	{
-		CreationParams.WindowSize.Width = width;
-		CreationParams.WindowSize.Height = height;
+		CreationParams.WindowSize.Width = 1280;
+		CreationParams.WindowSize.Height = 720;
 	}
 
 	core::dimension2d<u32> size = core::dimension2d<u32>(
@@ -440,8 +440,8 @@ s32 CIrrDeviceAndroid::handleInput(android_app* app, AInputEvent* androidEvent)
 			for (s32 i = 0; i < count; ++i)
 			{
 				event.TouchInput.ID = AMotionEvent_getPointerId(androidEvent, i + idx);
-				event.TouchInput.X = AMotionEvent_getX(androidEvent, i + idx);
-				event.TouchInput.Y = AMotionEvent_getY(androidEvent, i + idx);
+				event.TouchInput.X = AMotionEvent_getX(androidEvent, i + idx) * 1280 / 1920;
+				event.TouchInput.Y = AMotionEvent_getY(androidEvent, i + idx) * 720 / 1080;
 				
 				device->postEventFromUser(event);
 				
```

Benau:
duplicated

PR #2864, opened 2017-06-21
Small tweaks in graphics options

qwertychouskie:
Changes based on feedback on https://github.com/supertuxkart/stk-code/pull/2863.

Scale_rtts was removed from Image Quality because:

1. All the other options in this slider affect textures, so scale_rtts does not really fit here
2. It causes inconsistent/confusing behavior, as it only makes a difference when advanced lighting is enabled, so it will never be used in the presets anyways (1 uses "Very Low" but uses non-advanced lighting, 2-6 don't use "Very Low").

### DIFF
--- a/data/gui/custom_video_settings.stkgui
+++ b/data/gui/custom_video_settings.stkgui
@@ -5,16 +5,6 @@
 
         <spacer height="20" width="10" />
 
-        <!--
-        <div layout="horizontal-row" width="100%" height="fit">
-            <checkbox id="pixelshaders"/>
-            <spacer width="10" height="10"/>
-            <label text="Pixel Shaders" I18N="Video settings"/>
-        </div>
-
-        <spacer height="4" width="10" />
-        -->
-
         <div layout="horizontal-row" width="100%" proportion="1">
             <checkbox id="dynamiclight"/>
             <spacer width="10" height="10"/>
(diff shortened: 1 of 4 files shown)


auriamg:
We never asked for changes to scale_rtts_factor. We think it is just fine where it is. Stop flooding us with pull requests changing things we don't want changed and wasting both your and our time. Next time you contribute something, please make sure it's a change we actually want and not pointlessly moving around obscure options.

auriamg:
Okay, I spoke a bit too fast, you do have a point that scaling doesn't work without advanced lighting... we need to rethink about this

auriamg:
Okay so indeed on preset 1 scaling is not going to work, but what you did by ignoring "m_scale_rtts_factor" is not right, we should definitely ensure to set the proper scaling on higher qualities. I suggest to juste use scaling=1 on all preset levels and that should be good enough.

So you could change that from the PR.

Also, label changes like "Texture quality" and "Amount of scenery" should IMO be rollbacked, I don't think they're much clearer and certainly not worth breaking all existing translations over.

qwertychouskie:
> we should definitely ensure to set the proper scaling on higher qualities

STK defaults to a scale of 1.0, and currently this can only be changed by a setting in `config.xml`.  If someone sets a custom option using `config.xml`, the current code will undo the change as soon as the dialog is opened and closed, even if no options are changed.  This behavior is confusing and unwanted in most situations, as if someone specifically modified `config.xml`, they probably want those changes to not just mysteriously disappear.

qwertychouskie:
One thing I have learned (by reading a lot of comment threads :) is that people don't like programs undoing changes they specifically made, and generally get unhappy if this happens.

Benau:
Fixed all by me, beacuse you ignore what auria saod

PR #2865, opened 2017-06-21
Add Coverity badge

qwertychouskie:


auriamg:
We are not actively using coverity very much so that doesn't make sense. Let's not add badges of tools we don't really use.

PR #2867, opened 2017-06-22
Handle missing libopenglrecorder gracefully

qwertychouskie:


auriamg:
After discussion, the preferred way to do this would be :

set(PROJECT_VERSION "0.9.1") -> that could say "git", and the default of the recorder could depend on the version (disabled by default for git, enabled by default for release)

qwertychouskie:
The problem is version number is easily forgotten (it was still at 0.9.1!).

hiker:
That 'number' should always be 'git' on any dev branch.

Gee, just went back through the history, looks like it has been not-git ... forever :(
So as part of this, set the version in cmake to 'git' - our code is expecting this

qwertychouskie:
@hiker Requested changes implemented.

auriamg:
That seems good, thanks

Issue #2870, opened 2017-06-22
RPi - Compilation error when building with GLES


qwertychouskie:
```
In file included from /usr/include/GL/gl.h:2055:0,
                 from /tmp/stk-code-master/src/guiengine/widgets/../../../lib/irrlicht/source/Irrlicht/CIrrDeviceLinux.h:21,
                 from /tmp/stk-code-master/src/guiengine/widgets/CGUIEditBox.cpp:24:
/usr/include/GL/glext.h:468:19: error: conflicting declaration ‘typedef ptrdiff_t GLsizeiptr’
 typedef ptrdiff_t GLsizeiptr;
                   ^
In file included from /tmp/stk-code-master/src/graphics/gl_headers.hpp:47:0,
                 from /tmp/stk-code-master/src/graphics/2dutils.hpp:21,
                 from /tmp/stk-code-master/src/guiengine/widgets/CGUIEditBox.cpp:16:
/usr/include/GLES3/gl3.h:73:25: note: previous declaration as ‘typedef khronos_ssize_t GLsizeiptr’
 typedef khronos_ssize_t GLsizeiptr;
                         ^
In file included from /usr/include/GL/gl.h:2055:0,
                 from /tmp/stk-code-master/src/guiengine/widgets/../../../lib/irrlicht/source/Irrlicht/CIrrDeviceLinux.h:21,
                 from /tmp/stk-code-master/src/guiengine/widgets/CGUIEditBox.cpp:24:
/usr/include/GL/glext.h:469:19: error: conflicting declaration ‘typedef ptrdiff_t GLintptr’
 typedef ptrdiff_t GLintptr;
                   ^
In file included from /tmp/stk-code-master/src/graphics/gl_headers.hpp:47:0,
                 from /tmp/stk-code-master/src/graphics/2dutils.hpp:21,
                 from /tmp/stk-code-master/src/guiengine/widgets/CGUIEditBox.cpp:16:
/usr/include/GLES3/gl3.h:74:26: note: previous declaration as ‘typedef khronos_intptr_t GLintptr’
 typedef khronos_intptr_t GLintptr;
                          ^
CMakeFiles/supertuxkart.dir/build.make:2846: recipe for target 'CMakeFiles/supertuxkart.dir/src/guiengine/widgets/CGUIEditBox.cpp.o' failed
make[2]: *** [CMakeFiles/supertuxkart.dir/src/guiengine/widgets/CGUIEditBox.cpp.o] Error 1
[...]

PR #2879, opened 2017-07-10
Fix GLES compilation on RPi

qwertychouskie:


deveee:
I can't reproduce it, but as I said in the bug report https://github.com/supertuxkart/stk-code/issues/2870 the conflict between GL and GLES headers shouldn't be a problem anymore. If it's still a problem, then please paste the error message and I will try to find better solution.

PR #2882, opened 2017-07-20
Use track version 7 for walldriving tracks

qwertychouskie:
Any track/arena that depends on walldriving (e.g. icecave) should use version 7.


auriamg:
There is a _lot_ more to adding a new version than what you did in this PR. The exporter needs to export the new version number, for one.

Benau:
Not necessary after spm

PR #2883, opened 2017-07-22
Fix MinGW cross-compilation

qwertychouskie:
Auto-detect version


auriamg:
I'm not sure how reliable this "grep prosix" is, but I like the idea of dehardcoding the folder name, so I will merge it, thanks

PR #2890, opened 2017-08-01
Cleanup old stuff

qwertychouskie:
- Remove unused "topbar" widget
- Remove unused Irrrlicht files
- Update GL headers


hiker:
I am happy to remote the irrlicht test files which we don't use.
While top_bar appears to be unused atm, it is still supported by the gui, so this should stay.
Updating header files should be a separate issue with a reason for the update (just saying "it's newer" it's not good enough, we risk potentially breaking things that are working fine)

PR #2893, opened 2017-08-02
Save and use usernames in replay files (#2754)

### LINKED ISSUE #2754: Add player name to ghost replays
Currently there's no way to tell who made a replay. I suggest adding a playername field to the replay files and display it in the list of replays as another thing to sort by. Additionally display it in the race result instead of just kart name.

GEOFBOT:
Fixes #2754.

* Store usernames of each racer in recorded replay files
* Display those usernames in a column of the replay selection UI
  and in race result dialogs
  * If an older replay file is used that does not have usernames, kart name is displayed in race result dialogs (which is the current behavior)
* RaceResultGUI::getKartDisplayName functionality moved into
  Controller::getName

### DIFF
--- a/src/karts/controller/controller.cpp
+++ b/src/karts/controller/controller.cpp
@@ -34,3 +34,7 @@ Controller::Controller(AbstractKart *kart)
     m_kart     = kart;
     setControllerName("Controller");
 }   // Controller
+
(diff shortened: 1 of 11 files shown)


auriamg:
Thanks for the contribution, I've merged it! :)

Issue #2889, opened 2017-08-01
Simplify getLightFactor for non-advanced lighting


deveee:
This is just a reminder for me.

For non-advanced lighting specular and diffuse textures are single color, so it should be enough to just use something like:
```
return diffuseMatColor * 0.73;
```

For GLES renderer in Hacienda it increases fps from 24 to 28 on my device.

deveee:
@Benau Do you think that it's always true that specMapValue = 0.0 when advanced lighting is disabled?

Benau:
yes

deveee:
Ok, done in https://github.com/supertuxkart/stk-code/commit/bc6ff38b22e8b7457c21df7e438267d7fe935163

Issue #2894, opened 2017-08-02
Eventually move away from Irrlicht


qwertychouskie:
Back in 2010, when STK 0.7 was released, Irrlicht was modern, frequently updated, etc.  It was great, it provided everything STK could need in an engine, reducing required developer manpower.

Now it's 2017.  Things have changed.  Irrlicht is all but abandoned, STK uses it's own renderer, and is getting it's own scene manager and mesh loader...  It might finally be time for STK to dump Irrlicht and use something more maintained.

Currently, STK still uses Irrlicht for:

- Scene management (SP will replace this soon though)
- Fallback rendering for older versions of OpenGL (also will be removed with SP)
- Particle system (only supports billboard particles, resulting in ugly skidsparks)
- Window+OGL context creation and management
- Input management (which is causing headaches like #2872)
- GUI (ish, it's kind of a hybrid system)

The first two won't be an issue once SP is merged.  For the rest there are a few options:

Godot 3 (currently in alpha, should be out soon):
Handles everything STK could ever need it to, in active development, large user base, supports all platforms (Windows, OS X, Linux, *BSD, Haiku, iOS, Android, BlackBerry OS, HTML5), etc.  Since Godot was open-sourced in 2014, development has happened at a breakneck pace, with no sign of slowing down.
Handles:
- Everything including the kitchen sink and VR

Potential downsides:
- Big, might take a while to integrate into STK

My opinion is this is the best option.

SDL2:
Well known, well used, supported on pretty much all platforms (including Android and Wayland).
Handles:
- Window+OGL context creation and management
- Input management
Potential downsides:
- Doesn't handle GUI or particles, additional libraries would be required for this.

GLFW doesn't support Android, so probably a no-go.

SFML is similar to SDL2, though Android support is "soon".

Obviously eliminating Irrlicht won't happen immediately, but it is probably a good idea to start discussing options now.

deveee:
Qwerty I will eat you

2017-08-02 19:32 GMT+02:00 QwertyChouskie <notifications@github.com>:

auriamg:
@qwertychouskie Please. Stop these kinds of post. You are a newcomer to the project, you are in a good position to make small bugfixes, etc. It is up to lead developers to decide whether an engine change is appropriate, **not** up to the newcomer. We are the ones that can evaluate a proper migration strategy, ad just posting "let's switch engines!" falls way short is not helping us.

[...]

PR #2895, opened 2017-08-05
Simplify getLightFactor for non-advanced lighting

qwertychouskie:
See https://github.com/supertuxkart/stk-code/issues/2889.


auriamg:
Quoting deve from #2894 

> In #2895 you did it wrong too and you probably have no idea why there is the 0.73 value. And I don't know why you opened this pull request, because in the first sentence I assigned it to myself.

PR #2905, opened 2017-08-24
Update CHANGELOG.md

qwertychouskie:
Mainly formatting stuff.


auriamg:
oh come on there is no point to rewrite the entire history, what has already been released is in the past and done and over with

PR #2913, opened 2017-08-29
Fix memory leak

qwertychouskie:
As reported by Coverity.


auriamg:
It's already cleaned up in void HQMipmapGenerator::cleanThreadedLoader()

did you check there actuall is a leak? you can't just trust coverity. Your code would not work, you can't delete memory that was malloced anyway.

PR #2915, opened 2017-08-29
Spelling: Wiimote, OK, BlueZ, HTTPS

comradekingu:
BlueZ as per http://www.bluez.org/

### DIFF
--- a/src/input/wiimote_manager.cpp
+++ b/src/input/wiimote_manager.cpp
@@ -390,7 +390,7 @@ int WiimoteManager::askUserToConnectWiimotes()
 #else
         _("Press the 1 and 2 buttons on your Wiimote simultaneously to put "
           "it in discovery mode, then click on OK. "
-                  "Detailed instructions at https://supertuxkart.net/Wiimote"),
+                  "Detailed instructions at http://supertuxkart.net/Wiimote"),
 #endif
         MessageDialog::MESSAGE_DIALOG_OK_CANCEL,
         new WiimoteDialogListener(), true);
--- a/src/input/wiimote_manager.cpp
+++ b/src/input/wiimote_manager.cpp
@@ -386,7 +386,7 @@ int WiimoteManager::askUserToConnectWiimotes()
     new MessageDialog(
 #ifdef WIN32
         _("Connect your Wiimote to the Bluetooth manager, then click on OK. "
-                  "Detailed instructions at https://supertuxkart.net/Wiimote"),
+                  "Detailed instructions at supertuxkart.net/Wiimote"),
 #else
         _("Press the 1 and 2 buttons on your Wiimote simultaneously to put "
           "it in discovery mode, then click on OK. "
--- a/src/input/wiimote_manager.cpp
+++ b/src/input/wiimote_manager.cpp
@@ -39,7 +39,7 @@ WiimoteManager*  wiimote_manager;
 
 bool WiimoteManager::m_enabled      = false;
 
-/** Irrlicht device IDs for the wiimotes start at this value */
+/** Irrlicht device IDs for the Wiimotes start at this value */
 static const int    WIIMOTE_START_IRR_ID   = 32;
 
 WiimoteManager::WiimoteManager()
@@ -64,9 +64,9 @@ WiimoteManager::~WiimoteManager()
  */
 void WiimoteManager::launchDetection(int timeout)
 {
-  // It's only needed on systems with bluez, because wiiuse_find does not find alredy connected wiimotes
+  // It's only needed on systems with BlueZ, because wiiuse_find does not find alredy connected Wiimotes
 #ifdef WIIUSE_BLUEZ
-    //Cleans up the config and the disconnected wiimotes
+    //Cleans up the config and the disconnected Wiimotes
     int number_previous_wiimotes = 0;
     wiimote_t** previous_wiimotes = (wiimote_t**) malloc(sizeof(struct wiimote_t*) * MAX_WIIMOTES);
     memset(previous_wiimotes,0,sizeof(struct wiimote_t*) * MAX_WIIMOTES);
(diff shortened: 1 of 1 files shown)


comradekingu:
@Arthur-D I think for preventing MITM attacks, on what i assume to be a preloaded signature, i think it is different. Don't know if it even serves HTTP, or if the redirect is automatic, but still change is neither one way or the other.

auriamg:
Maybe, but this is so minor that I'm not sure it's worth breaking all translations... ( @hiker  ? )

comradekingu:
@auriamg That only happens with Transifex, every other platform handles fuzzy strings correctly. That being said, translators do like to see change for the better, and do correct things given the chance. On transifex strings are stored as suggestions.

Arthur-D:
The issue is that we don't only ship 100% translated languages, and most of those which aren't complete do not have someone maintaining that translation.

Here's an example with made up numbers (I don't care to check right away the numbers and we don't know for sure until release anyway), but say we ship 20 fully complete translations and 30 partially complete translations. Breaking a string means that there's a good chance a lot of translations get one less translated string, while the fully complete ones get only marginally better.

Of course this doesn't mean improvements don't need to be done at times but there's a cost attached and that cost might be larger than the benefits (fewer translations having a string translated than before).

So I'm personally very torn in this case, but luckily I'm not the one calling the shots. ;)

nado:
Bluez is only used in comments, so we dont care about it concerning translations.
The only changes affecting translations concern : 
wiimote -> Wiimote (why not)
Ok -> OK (not sure we even care, do you have any link explaining those spellings for my personal interest ?)
Your change line 391 (https://github.com/supertuxkart/stk-code/pull/2915/files#diff-9542354c9019980f0c7493c60310b19cR391) about the buttons, which I find less clear than previously.
Do we need to break all translations for those 3 changes ?

comradekingu:
@nado The change is not what breaks the translations, the nature of the translation platform is. 
Its not as if reverting to "Ok" has any profound effect. Then again a few of the translations may have gotten it right.
It is the same on every dialog there is, so if it looks different, it feels out of place.
http://grammarist.com/spelling/ok-okay/

hiker:
Re the http(s): the proper solution is to provide the url from STK, and not making it part of the translation. So I would prefer not breaking the translations for that string unless we do it properly, i.e. using %s instead of the URL, and then providing the real URL from stk (maybe from a config file).

src/input/wiimote_manager.cpp
```diff
@@ -385,12 +385,12 @@ int WiimoteManager::askUserToConnectWiimotes()
...
-        _("Connect your wiimote to the Bluetooth manager, then click on Ok. "
+        _("Connect your Wiimote to the Bluetooth manager, then click on OK. "
                   "Detailed instructions at supertuxkart.net/Wiimote"),
 #else
-        _("Press the buttons 1+2 simultaneously on your wiimote to put "
-          "it in discovery mode, then click on Ok. "
-                  "Detailed instructions at supertuxkart.net/Wiimote"),
+        _("Press the 1 and 2 buttons on your Wiimote simultaneously to put "
+          "it in discovery mode, then click on OK. "
+                  "Detailed instructions at https://supertuxkart.net/Wiimote"),
```
qwertychouskie: This still has the `https://`.

comradekingu:
@hiker Changed to HTTP now :)

deveee:
As I understand hiker wants to save the http://... link in config file and use %s variable in the translation. There are already some links in stk_config.xml: https://github.com/supertuxkart/stk-code/blob/master/data/stk_config.xml#L159 so you can look how it's done.

src/input/wiimote_manager.cpp
```diff
@@ -385,12 +385,12 @@ int WiimoteManager::askUserToConnectWiimotes()
...
-        _("Connect your wiimote to the Bluetooth manager, then click on Ok. "
+        _("Connect your Wiimote to the Bluetooth manager, then click on OK. "
                   "Detailed instructions at supertuxkart.net/Wiimote"),
 #else
-        _("Press the buttons 1+2 simultaneously on your wiimote to put "
-          "it in discovery mode, then click on Ok. "
-                  "Detailed instructions at supertuxkart.net/Wiimote"),
+        _("Press the 1 and 2 buttons on your Wiimote simultaneously to put "
+          "it in discovery mode, then click on OK. "
+                  "Detailed instructions at http://supertuxkart.net/Wiimote"),
```
qwertychouskie: I think the point here was to remove the prefix to make it the same as the line 389.

hiker:
I am happy to merge this in if the URL issue is fixed (ideally using a function that would create the URL from a domain and path component:  https://%ssupertuxkart.net%s).

UnMe0w:
bluez is not for windows so wiiuse is better

src/input/wiimote_manager.cpp
```diff
@@ -385,12 +385,12 @@ int WiimoteManager::askUserToConnectWiimotes()
 {
     new MessageDialog(
 #ifdef WIN32
-        _("Connect your wiimote to the Bluetooth manager, then click on Ok. "
-                  "Detailed instructions at supertuxkart.net/Wiimote"),
+        _("Connect your Wiimote to the Bluetooth manager, then click on OK. "
```
qwertychouskie: Not sure if this is even valid anymore, will test soon.

Chujinda:
....ไม่พูดดีกว่า

james165527:
i cant understand can u please translate it !

*Yours sincerely,*
Mark James Agripo.

james165527:
import java.util.*;
public class CalBmi
{
public static void main (String []args)
{
double bmi , weight , height;
Scanner keyboard = new Scanner(System.in) ;

System.out.println("Welcome to BMI") ;
System.out.println("Enter weight in kg") ;
weight = keyboard.nextDouble() ;
System.out.println("Enter height in m") ;
height = keyboard.nextDouble() ;
bmi = weight/ (height*height) ;
System.out.print("BMI is") ;
System.out.println(bmi);
}
}
*Yours sincerely,*
Mark James Agripo.

james165527:
public class BMI
{
public static void main ( String [] args)
{
double weight = 43;
double height = 1.43;
double num n m = ( weight) ( height x weight);
Systtem.out.println("Body Mass Index is:" + sum);
}
}

*Yours sincerely,*
Mark James Agripo.

BO41:
Is something blocking this PR? What is missing?
Or is it not relevant anymore an can be closed?

PR #2918, opened 2017-08-29
Update and rename README.ANDROID to README.ANDROID.md

qwertychouskie:
Make it look nicer.  I tried to make it still look nice when viewed as regular text too.

### DIFF
--- a/android/README.ANDROID.md
+++ b/android/README.ANDROID.md
@@ -1,13 +1,9 @@
 ================================================================================
-
- SUPERTUXKART
-
+# SuperTuxKart for Android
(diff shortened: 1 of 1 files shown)


leyyin:
This does not display properly when looking at https://github.com/qwertychouskie/stk-code/blob/patch-12/android/README.ANDROID.md

Or when visiting the root directory of the Android subfolder https://github.com/qwertychouskie/stk-code/tree/patch-12/android

qwertychouskie:
Not really needed and I have other things I need to be spending my time on.

PR #2855, opened 2017-06-12
Alert the user if an update is available

qwertychouskie:
Closes https://github.com/supertuxkart/stk-code/issues/1171.

PR #2924, opened 2017-09-01
Update notification (that actually works :)

### LINKED ISSUE #1171: Make STK 'updateable' in game
**Author: hikerstk**

It would be a nice feature to be able to update stk (or parts of stk) in game. Some thoughts:

1) data files:
they could be updated by making the file manager a bit more intelligent: if it tries to open a file X in the data package, it would first check if there is a file in '.../addons/update/X', and if so transparently load that newer file. Some work would be required to enable this feature for textures loaded from b3d models.
2) Executable: STK should be started from a script. This script will check for a new binary in '.../addons/update/bin' - and if so start that binary instead of the installed one. If STK exits with a certain return code, that script would automatically restart (which is used to download a new binary in game, then restart STK with the new binary). Security would be crucial, i.e. it should get a hash sum from a different server to make sure it is really an official binary.

The current addon server can then be used to install new updates in '.../addons'.

Some care must be taken of inconsistencies (e.g. some data is loaded from the installed directory, then an update is installed --> there might be some incosnsitencies now between updated and already loaded files). Also there must be a way of removing files (not from the install location, but from a previous update) to avoid taking up too much space on disk.

This is more of a long term goal, or a potential GSOC project.

Migrated-From: https://sourceforge.net/apps/trac/supertuxkart/ticket/1115

qwertychouskie:
Closes #1171.  Since the last PR:

 - I fixed the segfaulting (Yay! :)
 - The actual check was updated to use `HTTPRequest()` and now has proper logging
 - More tweaks and fixes

The dialog pops up when it detects an update.  If "Yes" is pressed, it opens <https://supertuxkart.net/Download>.  It also has a don't show again button to disable update checking, and checking can be enabled and disabled in the settings.

I was going to also add an auto-update feature for Windows, but @samuncle said that it was probably not needed, since Steam is coming soon.

Note that checking is disabled on Android, as the app store provides updates, and thus this would be redundant.

Screenshots:

 

Contents of CurrVer.php:
```
<?php
/* Version number of latest STK, must be in form "w.x.y.z". */
print "0.9.3.0";
?>
```

Tested to work on both Linux and Wine (and thus Windows).  If someone (@auriamg?) with a mac could also confirm opening the page works on macOS (just call `LinkHelper::OpenURL("https://supertuxkart.net/Download")` somewhere), it would be appreciated.


hiker:
Could you stop wasting our time? Still the same statement as your pull request 2855:
"Not sure if you noticed my message: we already have this feature (in a much more general way, with proper version comparison etc), so unless you tell us what is missing I will just close this ticket - it's a waste of time."

qwertychouskie:
1. The current system **brakes as soon as another blog post is made**.  For example, 0.9.1 currently shows this (just tested):

This doesn't tell them that 0.9.2 is available.
2. The current system shows notifications as a tiny strip of barely noticeable text on the bottom that most people won't notice.
3. The current system relies on the user knowing what version they have.  Most users usually have no idea.

On the other hand, this dialog:
 - Gives the user full control whether or not update notifications are shown
 - Has a button to open the webpage to download it immediately, instead of "put it off a while, finally open a browser to download it, forget why you opened the browser, etc."

Also see <https://github.com/supertuxkart/stk-code/pull/2855#issuecomment-317149456>.

Alayan-stk-2:
This would definitely be an improvement.

PR #2925, opened 2017-09-03
Skidding fixes

qwertychouskie:
Mainly for when the kart is in the air.

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -2301,9 +2301,9 @@ void Kart::updatePhysics(float dt)
     m_skidding->update(dt, isOnGround(), m_controls.getSteer(),
                        m_controls.getSkidControl());
     m_vehicle->setVisualRotation(m_skidding->getVisualSkidRotation());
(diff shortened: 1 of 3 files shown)


Arthur-D:
Not sure we want to fix this - it's come up before and then it was decided to keep it as a feature. While skidding in the air is not realistic it allows for more control and you can keep accumulating skid bonus even if you're slightly in the air at some point.

qwertychouskie:
If you want, I could separate this PR into two parts: the visual and auditory fixes; and the physics changes.

Arthur-D:
Maybe you could begin with describing it. "Skidding fixes - Mainly for when the kart is in the air." What is this pull request fixing, and why is it needed?

qwertychouskie:
The changes to effects were separated into https://github.com/supertuxkart/stk-code/pull/2941.

qwertychouskie:
Closing as the physics changes don't seem to be wanted (I have to admit I do use skidding in the air :).  See https://github.com/supertuxkart/stk-code/pull/2941 for effects changes.

PR #2932, opened 2017-09-06
Put screenshots in ~/.local/share/supertuxkart/screenshots

qwertychouskie:
Instead of `~/.cache/supertuxkart/screenshots`.

### DIFF
--- a/src/io/file_manager.cpp
+++ b/src/io/file_manager.cpp
@@ -960,7 +960,8 @@ void FileManager::checkAndCreateScreenshotDir()
     m_screenshot_dir  = getenv("HOME");
     m_screenshot_dir += "/Library/Application Support/SuperTuxKart/Screenshots/";
 #else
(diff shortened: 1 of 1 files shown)


src/io/file_manager.cpp
```diff
@@ -960,7 +960,8 @@ void FileManager::checkAndCreateScreenshotDir()
     m_screenshot_dir  = getenv("HOME");
     m_screenshot_dir += "/Library/Application Support/SuperTuxKart/Screenshots/";
 #else
-    m_screenshot_dir = checkAndCreateLinuxDir("XDG_CACHE_HOME", "supertuxkart", ".cache/", ".");
+    m_addons_dir = checkAndCreateLinuxDir("XDG_DATA_HOME", "supertuxkart",
+                                          ".local/share", ".stkaddons");
```
Flakebi: That should be `m_screenshot_dir` not `m_addons_dir`.
Also, `.stkaddons` as the last argument is wrong.

qwertychouskie: Fixed, thanks.

PR #2941, opened 2017-09-20
Don't show/play skidding effects when kart is in the air

qwertychouskie:
This is the visual/auditory only version of https://github.com/supertuxkart/stk-code/pull/2925 and does not change the physics behavior, it just disables sound/visual effects when the kart is in the air.


deveee:
I was complaining about it before STK 0.8 with skidding was even released and someone told me that karts should be able to skid in the air and this is a correct behavior :P

And if we assume that we should be able to use skidding in the air, then visual effect is a kind of feedback and it's IMO useful.

hiker:
We general all agree that skidding in air (while being unrealistic - but any kind of skidding that gives you a boost is unrealistic in the first place) improves game play, making it more fun, and we do need feedback for the user to be aware of this.

PR #2953, opened 2017-10-08
Compile on GCC 7

MuriloStrike:
There are some negative numbers being inserted into the variable of type "unsigned char", causing error when compiling in GCC 7.2

The following error appears several times:
narrowing conversion of '-X' from 'int' to 'unsigned char' inside {} [-Wnarrowing]

The "X" is any number because it contains several.


deveee:
I don't see any negative numbers there and actually I fixed similar error with gcc 6 some time ago (by converting all these negative numbers):
https://github.com/MuriloStrike/stk-code/commit/58846002ebf1137df372435140c1d91a5d455b25#diff-c01d69de71dda909b94f038997e7f4f8

I will check gcc 7 later.

MuriloStrike:
You're right ... I'm using the version that is in Gentoo and it contains the negative numbers, I did not notice the version. Ignore that.

Issue #2888, opened 2017-08-01
Make crash sound volume relative to impact


Arthur-D:
As the title says, it would be really nice if the crash sounds were played according to the relative physical impact, i.e. if you crash slowly you don't get full volume, but if you crash fast you do. Right now crashing sounds repetitive because it's always played at full volume as defined in sfx.xml. Maybe some formula could be used which takes relative velocity into account and auto-scales between 0.01 to 0.70 for example, but if the 0.7 value in sfx.xml changes it would scale between 0.01 and the new volume.

auriamg:
Fixed by pulled request

PR #2964, opened 2017-10-24
Crash sounds based on speed and direction

Fantasmos:
This pull request modifies the volume of the crash sound effect based on the speed of the kart/s being involved in the crash (and the direction as well). 

Physics aren't too much my strong point, but from a game-play perspective the sounds feel appropriate.


Fantasmos:
Based on this issue: https://github.com/supertuxkart/stk-code/issues/2888
The kart on kart collision could be simplified by someone a lot more comfortable with physics

hiker:
Quick comment: since we are currently preparing the 0.9.3 RC (and the release), it might take a while before we can get to this. It certainly sounds interesting, looking forward to try it ;)

src/karts/kart.cpp
```diff
@@ -2175,17 +2177,70 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
     {
         if (m_body->getLinearVelocity().length()> 0.555f)
         {
+			
+			float SpeedForMaxVolume = 15; //The speed at which the sound plays at maximum volume
```
hiker: Please read our coding style: variables are lower case with _

src/karts/kart.cpp
```diff
@@ -2175,17 +2177,70 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
...
+			else
+			{
+				btVector3 ThisKartRotation = getVelocity();
+				btVector3 OtherKartRotation = k->getVelocity();
+				
+				
+				btScalar ScaledRotationDifference = ThisKartRotation.angle(OtherKartRotation) / 3.14159; //Difference in angles of karts in collision, Scaled by PI
+				
+				
+				btScalar SpeedDifference = getSpeed() - k->getSpeed();
```
hiker: Don't use the bullet variables - just use plain float (and probably even Vec3 instead of btVector3)

src/karts/kart.cpp
```diff
@@ -2175,17 +2177,70 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
...
+				getCurrentEmitter()->play(getXYZ(), m_boing_sound);
             }
             else
             {
                 int idx = rand() % CRASH_SOUND_COUNT;
+				
                 SFXBuffer* buffer = m_crash_sounds[idx];
-                getNextEmitter()->play(getXYZ(), buffer);
+				
+				getCurrentEmitter()->play(getXYZ(), buffer);
```
hiker: Looks like whitespace errors ;)

src/karts/kart.cpp
```diff
@@ -2175,17 +2177,70 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
...
+			float Volume;
+			float MaxVolume = 1; //The maximum volume a sound is played at 
+
+			if (k == NULL) //Collision with wall
+			{
+				Volume = (m_speed / SpeedForMaxVolume);
+			}
+			else
+			{
+				btVector3 ThisKartRotation = getVelocity();
```
hiker: rotation = velocity?

Fantasmos:
Consider using linearvelocity and calculating the difference between karts. Check it uses a vector relative to the world not the kart.

Fantasmos:
Re implementation of collision sound: 

Utilise: 
(my_velocity  - other_velocity).length()

Make sure to ensure its accurate at all angles

hiker:
The vectors are all world vectors. Only exception is: Moveable::getVelocityLC() 'LC' = local coordinates, so that's in kart coordinates.

src/karts/kart.cpp
```diff
@@ -2173,19 +2173,46 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
     // karts from bouncing back, they will instead stuck towards the obstable).
     if(m_bounce_back_time<=0.0f)
     {
-        if (m_body->getLinearVelocity().length()> 0.555f)
-        {
-            // In case that the sfx is longer than 0.5 seconds, only play it if
-            // it's not already playing.
-            if (isShielded() || (k != NULL && k->isShielded()))
+        if (getVelocity().length()> 0.555f)
+		{
```
auriamg: Watch out for 'tab' characters, our coding conventions require spaces

src/karts/kart.cpp
```diff
@@ -2173,19 +2173,46 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
...
+                const float LengthOfDifference = VelocityDifference.length();
+            
+                volume = LengthOfDifference / speed_for_max_volume;
+            }
+            
+            if (volume > max_volume) { volume = max_volume; }
+            if (volume < min_volume) { volume = min_volume; }
+            
+            SFXBase* crash_sound_emitter = getNextEmitter();
+            crash_sound_emitter->reallySetVolume(volume);
```
auriamg: Do no use reallySetVolume, please use regular 'setVolume'. Sound effects must be processed on a secondary thread, not in the main gameplay thread, to avoid gameplay pauses while OpenAL processes stuff

src/karts/kart.cpp
```diff
@@ -2173,19 +2173,46 @@ void Kart::playCrashSFX(const Material* m, AbstractKart *k)
...
+            
+            // In case that the sfx is longer than 0.5 seconds, only play it if
+            // it's not already playing.
+            if (isShielded() || (k != NULL && k->isShielded()))
+            {
+                crash_sound_emitter->play(getXYZ(), m_boing_sound);
+			}
```
auriamg: more tab characters

auriamg:
the collision sound is often very soft now, so soft it's often hard to hear. maybe it could be tweaked to be a bit louder?

hiker:
Re "the collision sound is often very soft now, so soft it's often hard to hear. maybe it could be tweaked to be a bit louder?"
It might be worth trying to change the shape of the curve. Atm it's linearly (double speed => double volume). If the volume is too soft at low speeds, you might want to try:
```
volume = sqrt(volume)
```
Since volume is in [0,1], this will make the volume more high for slower speed. Other functions could be tested, too (log(1.0f+x)+0.3f ??) Needs some testing to see which works best.

auriamg:
@Fantasmos Agreed about 11 and 0.4, I think you can commit those and things are good as far as I'm concerned. @hiker  ?

Fantasmos:
@auriamg Just for clarification I haven't implemented 11 and 0.4 I instead did:  sqrt(m_speed / speed_for_max_volume).

This makes a majority of lower values higher and distinct from each other, but louder sounds don't receive much of an increase and are less distinct.

hiker:
Sure, go ahead

PR #2966, opened 2017-10-25
Fix 2nd-gen Celeron on Windows

qwertychouskie:
Also the Advanced Pipeline is disabled due to the card's slowness.

### DIFF
--- a/data/graphical_restrictions.xml
+++ b/data/graphical_restrictions.xml
@@ -1,10 +1,10 @@
 <?xml version="1.0"?>
 <graphical-restrictions>
-  <card is="Intel(R) HD Graphics"      os="windows" version="<9.0" disable="ForceLegacyDevice"/>
-  <card is="Intel(R) HD Graphics"      os="windows"                disable="HighDefinitionTextures"/>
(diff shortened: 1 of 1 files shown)


deveee:
Not possible in this way, because it affects all generations :/ For ironlake generation I added version="<9.0" that should be generally safe (it's very old driver and I don't think that intel will do any updates for it). But no idea how to detect 2nd-gen celeron...

qwertychouskie:
@deveee The last driver update for 2nd-gen processors was in 2015, so I set the version to "<10.0", which is probably the best option.

deveee:
For STK 0.9.4 you can try to read cpu name and add it to graphical restrictions, so that intel generations will be properly detected.

deveee:
This pull request is invalid after Benau's work. And I eventually added ForceLegacyDevice for version < 10.

If someone with newer generation really uses that old drivers, then he will get a message about outdated drivers and he can upgrade it. And we can't guarantee that shader-based pipeline with work for this driver because noone can test it, so that force legacy device is safer.

qwertychouskie:
Sounds good, thanks!

PR #2968, opened 2017-10-27
Improve windows installer script

leyyin:
This improves the NSIS installer in the following ways:
- updates the list of uninstalled files
- install STK with specific version, this allows us to have multiple versions of STK installed
- add STK to the system add/remove programs list
- update the icon of the installer
- change compression method to default zlib as it is much faster and the lzma compression only saves 20 MB anyways


Benau:
bad english by me:
I should have said: missing an uninstaller "in windows add/remove programs list", anyways the installer uploaded to sourceforge includes the changes in this pr

hiker:
Hi leyyin,

can you confirm that you used this installer, and the (rather old) redistributable that was in our repo for the 0.9.3-rc windows installer?

Thanks!
   Joerg

leyyin:
Yes, I used this installer but I used the redist for VS 2017 (https://www.visualstudio.com/downloads/ - "Other Tools and Frameworks")  as I was compiling with that.

But VS 2017 and VS 2015 binaries are generally compatible so it should not matter much. https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017

hiker:
Thanks!

Issue #2770, opened 2017-02-05
Remember number of AI karts per mode


Alayan-stk-2:
Right now, the value is shared across all mode. So say you prefer 11 AI karts in normal races and you do a 3-strike battle/football game ; your preferred value will have been forgotten for normal races.

Normal/time-trial/follow the leader may share a common value (they are quite similar modes) ; but it would be a small improvement to have different memorized value for 3-strike/football.

cpdef:
i wanna go maybe to do this, but if not i found out that the value is also forgotten if u start an ghost replay (set to 1), so this should be also changed.

Fantasmos:
Should we consider closing this issue, or leaving it until Grand Prix does it PER setup? I think it's fine to close up

PR #2969, opened 2017-10-28
Created Map<T1,T2> config data type, and save karts per gamemode.

Fantasmos:
Solving issue: https://github.com/supertuxkart/stk-code/issues/2770

I've created a new config data type that's very similar to the list one. As shown; it's a map with a definable type for both the key and value. 

To demonstrate its usefulness, I've utilized it to save the number of karts per gamemode. The key values are the minor game mode type, and the values are the number of karts (an int to int pair).
(Due to this I've renamed m_num_karts to m_default_num_karts) 

This template could later be reused to save the kart per gamemode as well!

The main issues i've become concerned about are: 
- A lot of the code is repeated in ListUserConfigParam
- The usage of "typedef std::pair<T, U> PairType;" in "src/config/user_config.cpp"
- --numkartsis in the command line


Fantasmos:
The Travis CI build is failing due to the aforementioned issue: 
- The usage of "typedef std::pair<T, U> PairType;" in "src/config/user_config.cpp"
- "error: second argument to 'va_arg' is of non-POD type 'pair_type' (aka 'pair<int, int>') [-Wnon-pod-varargs]"

On my setup it's compiling and running fine, indicating that there's a way to rewrite these lines. This is my first time using templates however so i'm unsure how to approach this and any help is appreciated

src/states_screens/gp_info_screen.cpp
```diff
@@ -235,7 +235,7 @@ void GPInfoScreen::init()
 
         // Avoid negative numbers (which can happen if e.g. the number of karts
         // in a previous race was lower than the number of players now.
-        int num_ai = UserConfigParams::m_num_karts - race_manager->getNumLocalPlayers();
+        int num_ai = UserConfigParams::m_default_num_karts - race_manager->getNumLocalPlayers();
```
Fantasmos: Need to change this to use the dictionary, but ensure that there are no identical enum values (because one is for minor mod whilst the other is for major mode).

src/main.cpp
```diff
@@ -1189,16 +1189,16 @@ int handleCmdLine()
...
+        UserConfigParams::m_default_num_karts = n;
+        if(UserConfigParams::m_default_num_karts > stk_config->m_max_karts)
         {
             Log::warn("main", "Number of karts reset to maximum number %d.",
                       stk_config->m_max_karts);
-            UserConfigParams::m_num_karts = stk_config->m_max_karts;
+            UserConfigParams::m_default_num_karts = stk_config->m_max_karts;
         }
-        race_manager->setNumKarts( UserConfigParams::m_num_karts );
+        race_manager->setNumKarts( UserConfigParams::m_default_num_karts );
```
Fantasmos: Perhaps remove the saving to userconfigparams::m_default_num_karts, and just call setnumkarts (after ensuring its valid)

src/config/user_config.cpp
```diff
@@ -321,7 +321,154 @@ core::stringc ListUserConfigParam<T, U>::toString() const
...
+    all_params.push_back(this);
+    if (comment != NULL) m_comment = comment;
+
+    // add the default list
+    va_list arguments;
+    va_start(arguments, nb_elements);
+   
+    struct pair_type { T key; U value; };
+
+    for (int i = 0; i < nb_elements; i++) {
```
auriamg: Braces go on a newline, here and in many other places

Fantasmos: I'll fix it in other places, however when defining the struct which is equivalent to an std::pair<T,U> is it acceptable to keep it all in one line considering that's how an std::pair would be? (Also it would take up significantly less space).

auriamg: You mean something like "struct X { int a; int b; }"? yes that would be OK

src/states_screens/gp_info_screen.cpp
```diff
@@ -235,7 +235,7 @@ void GPInfoScreen::init()
 
         // Avoid negative numbers (which can happen if e.g. the number of karts
         // in a previous race was lower than the number of players now.
-        int num_ai = UserConfigParams::m_num_karts - race_manager->getNumLocalPlayers();
+        int num_ai = UserConfigParams::m_karts_per_gamemode[RaceManager::MAJOR_MODE_GRAND_PRIX] - race_manager->getNumLocalPlayers();
```
Fantasmos: it could be worth creating a new dictionary that stores per grand-prix, however that may be overkill.

src/main.cpp
Fantasmos: Could be worth also appending the minormode to the end of the major mode, thoughts?

auriamg:
@Fantasmos sorry for the little delay, I think we'll check this after the release, at this stage of the release process we are focused on bugfixes. We aren't forgetting you! (don't hestitate to give us a little reminder once the release is out, though :) )

auriamg:
Hi, sorry for the delay. As the release is going smoothly, I've had some time to look at this. One question I have it, why is there a "size" attribute? Can't you simply infer the number of items from how many attributes there are?

```
    <num_karts_per_gamemode
        Size="3"
        1002="0"
        1100="4"
        2000="6"
    >
```

to me it seems like the "size=3" part is redundent and unecessary.
Apart from that, everything seems to work well, nice work

Fantasmos:
@auriamg I wrote i  a while ago but from what i recall, it was necessary/included for the other structure that used a list. I'm unsure if I kept it to ensure compatibility or a 'just in case' scenario however.

auriamg:
@Fantasmos With the release done, I've merged your pull request, thanks!

PR #3007, opened 2017-11-02
Android: Run at a reduced resolution

SpotComms:
Running at a reduced resolution dramatically increases performance while also reducing battery drain.

Based off of https://github.com/supertuxkart/stk-code/pull/2863#issuecomment-309864131

This also enables immersive/expanded mode which hides the title bar and navigation bar, which is necessary to get access to the screen's entire real estate.

Edit: This doesn't do anything, height and width is probably mixed

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -8,7 +8,7 @@
     <application android:label="@string/app_name"
                  android:icon="@drawable/icon"
                  android:hasCode="false"
-                 android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
+                 android:theme="@style/AppTheme.NoActionBar"
                  android:hardwareAccelerated="true">
 
         <activity android:name="android.app.NativeActivity"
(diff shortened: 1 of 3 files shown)


deveee:
I like the idea, but it's a bit more complicated.

The main problem is that irrlicht device may be initialized multiple times (when extracting assets, null device, main GLES 3.0 device, fallback to GLES2 if needed). And IIRC when you set width/height using ANativeWindow_setBuffersGeometry then ANativeWindow_getWidth/Height will return incorrect values next time.

Maybe it could be possible to use ```ANativeWindow_setBuffersGeometry(m_egl_window, 0, 0, format);``` in destructor to restore original values, but I'm not sure if it will always work.

Or maybe just store original width/height as the static variables and read them only once on startup, when window is created.

You also need to scale touch/mouse events when resolution is scaled.

I would also prefer scale factor (i.e. 0.8, 0.7, 0.5) rather than hardcoded resolutions.

And second problem is that I'm not sure if it will always work and I don't have possibilities to properly test it myself. Maybe after 0.9.3 we can create "supertuxkart-unstable" project and then push it as alpha release for testing.

PR #3008, opened 2017-11-02
Fixing selection issue on end-race screen

Fantasmos:
Here's the fix for gamepad controllers. May be worth making this part of the setvisible function.


qwertychouskie:
Yeah, it would probably make sense to make `setFocusable()` part of `setVisible()`.

auriamg:
Sorry for the late feedback - this seems like a good fix, though as discussed on IRC it may make sense to make this behavior automatic for hidden items.

Also, you handled one specific buttons, but other buttons may be shown or hidden even on that one specific screen, it would make sense to support all of them

Fantasmos:
@auriamg I've had a go at making it a global fix, however it doesn't look like simple. I tried editing the code that checks if a button is focusable, but didn't seem to fix it (however I was rushing a bit when doing it). I'll have another go in a few days hopefully.

auriamg:
Okay, I've merged the workaround, a more complete fix can always come later, thanks

PR #3013, opened 2017-11-03
README: Correct building cmake command sentence

urtzai:
cmake doesn't find CMakeLists.txt according to the README. Needs to specify stk-code folder


nado:
Thats because you're supposed to execute these commands *in* stk-code directory (therefore ../ = stk-code).

urtzai:
So the README is not clear ... it says the opposite. That's what I fixed

Benau:
Not necessary anymore

PR #3014, opened 2017-11-03
All language names capitalized

urtzai:
In order to keep a correct order in a language selection menu, the language names need to be capitalized


urtzai:
For example, the basque language was at the end of the list before it was capitalized:

2017-11-03 10-32-51-ko-pantaila-argazkia

Now has a correct order. I've only fixed names with Latin letters... I suppose the others are already correct.

PR #3015, opened 2017-11-04
Mark "fortmagma" as locked on the minimap until the door can open

LeMagnesium:
Keep the "fortmagma" challenge marked as locked on the GUI minimap even if enough points (180+) have been gathered to unlocked it, when the door is not open yet.
This tiny change make things way less confusing than having the final track "unlocked" and arriving at the door to be told that it is, in fact, not available yet.

### DIFF
--- a/src/states_screens/race_gui_overworld.cpp
+++ b/src/states_screens/race_gui_overworld.cpp
@@ -443,6 +443,19 @@ void RaceGUIOverworld::drawGlobalMiniMap()
         const ChallengeData* challenge = unlock_manager->getChallengeData(challenges[n].m_challenge_id);
         const unsigned int val = challenge->getNumTrophies();
         bool unlocked = (PlayerManager::getCurrentPlayer()->getPoints() >= val);
+        if (challenges[n].m_challenge_id == "fortmagma")
+        {
+            // For each track, check whether any difficulty has been completed ; fortmagma will not affect our decision (`n == m`) ; tutorial is ignored because it has no completion level
+            for (unsigned int m = 0; unlocked && m < challenges.size(); m++)
+            {
+                if (challenges[m].m_challenge_id == "tutorial") continue;
+                    unlocked = unlocked &&
+                        (PlayerManager::getCurrentPlayer()
+                            ->getChallengeStatus(challenges[m].m_challenge_id)
+                            ->isSolvedAtAnyDifficulty() || n == m);
+            }
+        }
+
         int state = (unlocked ? OPEN : LOCKED);
         
         if (UserConfigParams::m_everything_unlocked)


src/states_screens/race_gui_overworld.cpp
```diff
@@ -443,6 +443,15 @@ void RaceGUIOverworld::drawGlobalMiniMap()
         const ChallengeData* challenge = unlock_manager->getChallengeData(challenges[n].m_challenge_id);
         const unsigned int val = challenge->getNumTrophies();
         bool unlocked = (PlayerManager::getCurrentPlayer()->getPoints() >= val);
+	if (challenges[n].m_challenge_id == "fortmagma") {
+		// For each track, check whether the easy difficulty has been completed ; fortmagma will not affect our decision (`n == m`) ; tutorial is ignored because it has no completion level
+		for (unsigned int m = 0; unlocked && m < challenges.size(); m++) {
+			if (challenges[m].m_challenge_id == "tutorial") continue;
+			unlocked = unlocked &
+				(PlayerManager::getCurrentPlayer()->getChallengeStatus(challenges[m].m_challenge_id)->isSolved(RaceManager::DIFFICULTY_EASY) || n == m);
```
nado: Does that also work when you solved a challenge on higher difficulty ?
I haven't checked where the isSolved method is implemented.

LeMagnesium: It did (and actually, I solved all challenges in difficult first, which is what made me notice the discrepancy I'm fixing here). I checked both low and high difficulty, and saw other uses of `isSolved` in the already existing code. If `RaceManager::DIFFICULTY_HIGH` or `RaceManager::DIFFICULTY_MEDIUM` are completed for a track, `isSolved` will also return `true` if we check the lower difficulty (which makes sense since both medium and easy levels are marked as complete if you do the expert mode first).

I noticed though that I could use `isSolvedAtAnyDifficulty`, which, perhaps, is less confusing. Should I change it?

nado: Neat, I suppose it is not less confusing for someone used to the base code, but I find it easier to read without knowing much from the rest of the code.

LeMagnesium: Then it shall be pushed in a minute or so (I had it coded and tested since I posted my answer).
Look out for the new diff right after.

Benau:
Use 4 spaces than tab, { from if goes next line, also your characters per line is too long, < 80 chars best

LeMagnesium:
That's done.

src/states_screens/race_gui_overworld.cpp
```diff
@@ -443,6 +443,17 @@ void RaceGUIOverworld::drawGlobalMiniMap()
         const ChallengeData* challenge = unlock_manager->getChallengeData(challenges[n].m_challenge_id);
         const unsigned int val = challenge->getNumTrophies();
         bool unlocked = (PlayerManager::getCurrentPlayer()->getPoints() >= val);
+        if (challenges[n].m_challenge_id == "fortmagma") {
+            // For each track, check whether the easy difficulty has been completed ; fortmagma will not affect our decision (`n == m`) ; tutorial is ignored because it has no completion level
```
qwertychouskie: Should say "check whether any difficulty"...

LeMagnesium: Correct. I forgot to change the comment.

src/states_screens/race_gui_overworld.cpp
```diff
@@ -443,6 +443,17 @@ void RaceGUIOverworld::drawGlobalMiniMap()
         const ChallengeData* challenge = unlock_manager->getChallengeData(challenges[n].m_challenge_id);
         const unsigned int val = challenge->getNumTrophies();
         bool unlocked = (PlayerManager::getCurrentPlayer()->getPoints() >= val);
+        if (challenges[n].m_challenge_id == "fortmagma") {
```
auriamg: Bracket goes on a newline.

Also, I'm not a huge fan of hardcoding "fortmagma" in the code :( Though the alternatives I can think of require a lot more code and may not be worth it... Maybe someone has an idea

src/states_screens/race_gui_overworld.cpp
```diff
@@ -443,6 +443,17 @@ void RaceGUIOverworld::drawGlobalMiniMap()
...
+            // For each track, check whether any difficulty has been completed ; fortmagma will not affect our decision (`n == m`) ; tutorial is ignored because it has no completion level
+            for (unsigned int m = 0; unlocked && m < challenges.size(); m++) {
```
auriamg: Bracket goes on a newline

src/states_screens/race_gui_overworld.cpp
LeMagnesium: I'm not really a fan of it either... but there is no marker for the final track, and adding it would require adding stuff which may/may not go along the rest of the code's architecture...

auriamg:
Okay so as discussed I'm not in love with that fix, but until we can re-think story mode on stronger foundations this patch will do. Thanks

PR #3020, opened 2017-11-05
Fix dependencies folder

qwertychouskie:
I checked all the Cmake files and these seem to be the only ones not updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,8 +139,8 @@ else()
 endif()
 
 if (BUILD_RECORDER)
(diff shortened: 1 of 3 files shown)


deveee:
Are you sure that it's correct? I didn't check it yet, but it's possible that toolchain file is executed before CMakeLists.txt, so dependencies variable may be not known yet.

deveee:
I checked and ${DEPENDENCIES} doesn't work in toolchain file. I just used hardcoded "dependencies-64bit" and already made a commit: https://github.com/supertuxkart/stk-code/commit/08be02af8f069215c7bf9a61d4cf78bbec9a8008

Thanks

qwertychouskie:
You're welcome, and thank you for realizing my stupid mistake ;)

PR #3022, opened 2017-11-06
Reset FeatureAvailable array to a sane default

austriancoder:
The FeatureAvailable array elements never get reseted to false. This
causes random problems when using the fixed pipeline rendering path.
This problem was found during testing of the etnaviv gallium driver
found in recent versions of mesa.

Signed-off-by: Christian Gmeiner <christian.gmeiner@gmail.com>

### DIFF
--- a/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp
+++ b/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp
@@ -329,6 +329,8 @@ void COpenGLExtensionHandler::dumpFramebufferFormats() const
 
 void COpenGLExtensionHandler::initExtensions(bool stencilBuffer, bool useCoreContext)
 {
+	memset(FeatureAvailable, 0, sizeof(FeatureAvailable));
+
 	const f32 ogl_ver = core::fast_atof(reinterpret_cast<const c8*>(glGetString(GL_VERSION)));
 	Version = static_cast<u16>(core::floor32(ogl_ver)*100+core::round32(core::fract(ogl_ver)*10.0f));
 	if ( Version >= 102)


deveee:
It has been changed/broken in this commit 271a26eaf0b3e05e5445f446a80f107633674feb, so it's now always true.

This may be also the reason why STK crashes on old intel 965gm on windows (it has been broken after 0.8.1 version).

deveee:
In this way we will keep it true for core profile:
```
diff --git a/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp b/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp
index 8b7afb0ad4..9af150c0b7 100644
--- a/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp
+++ b/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.cpp
@@ -336,15 +336,16 @@ void COpenGLExtensionHandler::initExtensions(bool stencilBuffer, bool useCoreCon
 	else
 		os::Printer::log("OpenGL driver version is not 1.2 or better.", ELL_WARNING);
 
+	if (!useCoreContext)
 	{
+		for (u32 i = 0; i < IRR_OpenGL_Feature_Count; i++)
+			FeatureAvailable[i] = false;
+			
 		const char* t = NULL;
 		size_t len = 0;
 		c8 *str = 0;
 		
-		if (!useCoreContext)
-		{
-			t = reinterpret_cast<const char*>(glGetString(GL_EXTENSIONS));
-		}
+		t = reinterpret_cast<const char*>(glGetString(GL_EXTENSIONS));
 		
 		if (t)
 		{
diff --git a/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.h b/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.h
index 597c71b6e1..319b1329a0 100644
--- a/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.h
+++ b/lib/irrlicht/source/Irrlicht/COpenGLExtensionHandler.h
@@ -932,8 +932,6 @@ class COpenGLExtensionHandler
 	//! queries the features of the driver, returns true if feature is available
 	bool queryOpenGLFeature(EOpenGLFeatures feature) const
 	{
-	  if (COpenGLExtensionHandler::IRR_EXT_packed_depth_stencil)
-		 return true;
 		return FeatureAvailable[feature];
 	}
 
```

Is this ok for you?

austriancoder:
Looks good to me.

Tested-by: Christian Gmeiner <christian.gmeiner@gmail.com>

deveee:
Ok, I already pushed it to the repo:
https://github.com/supertuxkart/stk-code/commit/3baefcd6a7563bf9a5ca0ed3c7583f00dd167957

Thanks a lot!

chealy:
Is this a bug that should be pulled into the final 0.9.3 release?  At least in the case of etnaviv on the i.MX6, having this bug fixed prevents the application from crashing.

qwertychouskie:
@chealy It has been backported to the 0.9.3 branch, see https://github.com/supertuxkart/stk-code/commit/2a5baac3317edfa7aa284505cd9b538ca67fc9ce

chealy:
Excellent, thanks for letting me know!

PR #3073, opened 2017-12-12
Update README.ANDROID

RiderExMachina:
Changed "he" to "they" for a more gender neutral tone. (I hate being this guy, but I do think using "they" is better)


deveee:
At least userS have to choose if they need... And anyway these issues should be probably placed on github with "Android" label.

1, 2 are invalid, 3, 5 are already on github and for 4 it's said in readme that user should use NDK r12.

deveee:
Fixed in https://github.com/supertuxkart/stk-code/commit/2b8737a75bdb6706b71d67be2ff9edb7426af3f4

Issue #3034, opened 2017-11-10
Badly adjusted playername for replays


Arthur-D:
The vertical adjustment of playername for replays is not very good. See attached screenshot. From looking at the screenshot, which of the replays would be Kitoko's?
![replay_username_badly_adjusted](https://user-images.githubusercontent.com/5515261/32681453-e0fce522-c66f-11e7-91a9-a15862a9a8c0.jpg)

kitokofox:
Mine is the STK Enterprise replay with 122.967s

PR #3081, opened 2017-12-23
Ghost replay selection fix

Fantasmos:
To fix: https://github.com/supertuxkart/stk-code/issues/3034

Example: https://i.imgur.com/WRniypS.png 

A simple solution to this issue, alternative solutions are to make it " " instead of "Unknown."


Fantasmos:
Changed it to be language neutral and instead have no number: https://i.imgur.com/WEwZD4m.png

auriamg:
Okay, but why must it be a whitespace character? Couldn't it just be an empty string "" ?

auriamg:
What I was trying to say in my previous comment is : I guess I'd prefer to understand the root of the problem, rather than patching it locally by adding spaces. If you can't find the root of the issue I guess that's still better than nothing though

Fantasmos:
@auriamg  The issue is that there are three vectors that are inserted values for each replay, however when a username doesn't exist it doesn't insert a value into the column. Therefore when it comes to display it, since there are less values than other columns fucky stuff happens. 

Therefore having strings be inserted in place of actual values is a solution, and it doesn't disrupt existing setups or add more unecessary load. The only other fix I can think of is the parameter that's "false" may have other inputs or settings for this scenario.

auriamg:
Merged, thanks

PR #3090, opened 2018-01-11
Reduce number of places version number needs to be changed on release

qwertychouskie:
Just greasing the wheels of the releasing process a little bit. :)  ~Note that std::regex_replace is cpp11~ (no longer needed).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 2.8.4)
 # root CMakeLists for the SuperTuxKart project
 project(SuperTuxKart)
 set(PROJECT_VERSION "git")
+add_definitions( -DSUPERTUXKART_VERSION="${PROJECT_VERSION}" )
 
 if(NOT (CMAKE_MAJOR_VERSION VERSION_LESS 3))
   cmake_policy(SET CMP0043 OLD)
--- a/data/CREDITS
+++ b/data/CREDITS
@@ -1,6 +1,6 @@
 = SuperTuxKart =
 A Kart Racing Game Featuring Tux & Friends
-- Version git
+- Version $STKVERSION$
 - Visit us at supertuxkart.net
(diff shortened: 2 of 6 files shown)


deveee:
Just note that I'm ok with upgrade to gcc 4.9.

STK 0.9.4 linux package probably will be based on Debian Jessie (gcc 4.9.2) or Ubuntu 16.04 (gcc 5.3.1). Or maybe I will compile whole package using Android script and dependencies.

For Android I prefer gcc <= 4.9.2, but in worst case I can hack android script to make it working with clang.

leyyin:
An alternative would be just to use a `std::string` replace.

Add this to the `namespace StringUtils` (see `string_utils.hpp`).
```cpp
std::string findAndReplace(const std::string& source, const std::string& find, const std::string& replace)
{
    std::string destination = source;
    std::string::size_type found_position = 0;

    // Replace until we can't find anymore the find string
    while ((found_position = destination.find(find, found_position)) != std::string::npos)
    {
        destination.replace(found_position, find.length(), replace);
        // Advanced pass the replaced string
        found_position += replace.length();
    }
    
    return destination;
}
```

And use it like this:
```cpp
// To be honest I would use something else than "STKVERSION" to be the key to search
// "$STKVERSION$" or maybe "__STKVERSION__"
line = StringUtils::findAndReplace(line, "$STKVERSION$", STK_VERSION);
```

hiker:
+2 for everything @leyyin just wrote :)

qwertychouskie:
I just tested @leyyin's code and it seems to work, so I force-pushed the commit (removing the Travis-changing commit).  My idea was to make things easier (especially if a "0.9.3.1" is released to fix the add-ons manager crash with Russan and the gamepad crash with gamepads that report too many buttons), not more complicated. :)

auriamg:
@hiker I'd tend to merge this, agreed?

qwertychouskie:
Thanks!

Issue #1675, opened 2014-11-07
Rescued karts appears in middle of road without any warning


Alayan-stk:
This causes karts behind to bump them and lose all their speed. There is no way to avoid that except knowing where rescued tracks may be released and avoid that part of road (not practically doable, and won't stop AI to be tricked with that)
I took a look on MK to see how it is managed, and there the rescue animation happens over the road with a little time before the rescued kart is released; In STK, the rescue animation is done where thunderbird rescue the kart, and then it appears instantly on the road.

auriamg:
A fix for this was merged

PR #3106, opened 2018-01-26
Better rescue

Fantasmos:
Solution to: https://github.com/supertuxkart/stk-code/issues/1675

This animation is a major improvement over the current one, as the current one is much more disruptive to the other players. Improvements could be made, including to add an additional non-collision period after being placed down (however this current one is in-line with other games, making it sufficient, and this is also a design choice). 

Video preview at: https://youtu.be/qSTPWhoKqn8


Fantasmos:
if peopel are unhappy with how the animation is, setting the split time to 0.5 may be more appropriate.

hiker:
TBH, I find the jump in the middle more ... 'disturbing' or noticeable then what we currently have. To me it feels like the camera is 'jerking' twice now instead of once.

Fantasmos:
Could we perhaps include a fade in and out?

Benau:
transparency fade-in / out?

auriamg:
The proposed solution seems interesting, but I agree the camera shouldn't cut twice. I'm pretty sure this can be done using a single camera cut

Fantasmos:
Latest commit has made it into 1 camera cut

auriamg:
Is there a data file needed to try this? I get the following error, and it crashes

```
[error  ] Irrlicht: Could not load mesh, because file could not be opened: : ../data/../../stk-assets/models/reset-button.spm

>	supertuxkart.exe!AttachmentManager::loadModels() Line 95	C++
 	supertuxkart.exe!main(int argc, char * * argv) Line 1615	C++

```

Fantasmos:
I've not modified that file or anything related, strange. I'll look into it later today

hiker:
I seem to remember that Benau recently asked about removing reset-button.spm recently

Benau:
git merge origin/master

auriamg:
Much better! Once thing that's still slightly annoying, is that for maybe 1 second just after the cut, the camera quickly moves back into place, would it be possible to remove that? It's a little dizzying

auriamg:
In case my previous comment wasn't clear enough, here's a capture, especially for the third rescue there is some rapid camera motion snapping into place very quickly, that's not smooth

[SuperTuxKart 2018-01-29 7_03_25 PM.zip](https://github.com/supertuxkart/stk-code/files/1675737/SuperTuxKart.2018-01-29.7_03_25.PM.zip)

Benau:
auria you forgot libopenglrecorder

Fantasmos:
Ok this should be done, demonstration at: https://youtu.be/rPgnu3BJRKk

deveee:
Indeed it seems to have more sense than our current rescue (I just watched video, didn't look at the code).

Actually I wonder if we can for example tweak lightning shader and add fading to black and then appearance effect when "teleport" happens during rescue. We can pass full color to lightning.frag as uniform variable and change the shader name to something more generic. But it may be a separate task for later.

src/graphics/camera_normal.cpp
```diff
@@ -117,22 +117,23 @@ void CameraNormal::smoothMoveCamera(float dt)
...
-    m_camera_offset += (wanted_camera_offset - m_camera_offset) * delta;
-
+    
     float delta2 = dt * 8.0f;
     if (delta2 < 0)
         delta2 = 0;
     else if (delta2 > 1)
         delta2 = 1;
 
+    m_camera_offset += (wanted_camera_offset - m_camera_offset) * delta;
```
auriamg: you edited the camera.cpp, and I'm not sure why
this line was moved, but it seems like it shouldn't matter?

src/graphics/camera_normal.cpp
```diff
@@ -68,8 +68,8 @@ CameraNormal::CameraNormal(Camera::CameraType type,  int camera_index,
 //-----------------------------------------------------------------------------
 /** Moves the camera smoothly from the current camera position (and target)
  *  to the new position and target.
- *  \param wanted_position The position the camera wanted to reach.
- *  \param wanted_target The point the camera wants to point to.
+ *  \param dt Delta time, 
+ *  It's expected in SnapToPosition() that setting DT to 1 snaps to the destination
```
auriamg: that assumption seems weird to me, "dt" is the number of seconds elapsed since last frame. So 1 means "1 second", not necessarily "snap to end".  Calling the code with a large dt will simulate that a lot of time happened and skip the animation, so the logic is okay, the comment isn't quite true though

src/graphics/camera_normal.cpp
```diff
@@ -160,7 +161,10 @@ void CameraNormal::smoothMoveCamera(float dt)
     assert(!std::isnan(m_camera->getPosition().Z));
 
 }   // smoothMoveCamera
-
+void CameraNormal::SnapToPosition()
```
auriamg: our convention is that method/function names should be camelCase

src/karts/rescue_animation.cpp
```diff
@@ -37,6 +37,7 @@
 
 #include <algorithm>
 
+#include "graphics\camera_normal.hpp"
```
auriamg: Should be a forward-slash for better portability, only windows uses backslashes in paths

src/karts/rescue_animation.cpp
```diff
@@ -54,20 +55,8 @@ RescueAnimation::RescueAnimation(AbstractKart *kart, bool is_auto_rescue)
...
-    if (tm.castRay(m_xyz, to, &hit, &m, NULL/*normal*/, /*interpolate*/true))
-    {
-        hit_dest = (hit - m_xyz).length();
-        hit_dest -= Referee::getHeight();
-        if (hit_dest < 1.0f)
-        {
-            hit_dest = 1.0f;
-        }
-    }
+    float hit_dest = MaximumHeight();
```
auriamg: Method names should be camelCase

src/karts/rescue_animation.cpp
```diff
@@ -101,50 +90,86 @@ RescueAnimation::RescueAnimation(AbstractKart *kart, bool is_auto_rescue)
...
+    m_kart->getBody()->setLinearVelocity(btVector3(0, 0, 0));
+    m_kart->getBody()->setAngularVelocity(btVector3(0, 0, 0));
     m_kart->getNode()->removeChild(m_referee->getSceneNode());
     delete m_referee;
     m_referee = NULL;
-    if(m_timer < 0)
+}   // ~RescueAnimation
+
+// Determine maximum rescue height with up-raycast
+float RescueAnimation::MaximumHeight() {
```
auriamg: Brackets go on a newline, you know that by now...

src/karts/rescue_animation.cpp
```diff
@@ -101,50 +90,86 @@ RescueAnimation::RescueAnimation(AbstractKart *kart, bool is_auto_rescue)
...
+    {
+        if (kart_on_track == false) 
+        {
+            kart_on_track = true;
+
+            World::getWorld()->moveKartAfterRescue(m_kart);
+
+            for (unsigned int i = 0; i < Camera::getNumCameras(); i++)
+            {
+				CameraNormal* camera = dynamic_cast<CameraNormal*>(Camera::getCamera(i));
```
auriamg: Tabs

src/karts/rescue_animation.cpp
```diff
@@ -101,50 +90,86 @@ RescueAnimation::RescueAnimation(AbstractKart *kart, bool is_auto_rescue)
...
+                if (camera && camera->getKart() == m_kart &&
+                    dynamic_cast<CameraNormal*>(camera)) 
+                {
+					camera->setMode(Camera::CM_NORMAL);
```
auriamg: Tabs

auriamg:
@Fantasmos merged, thanks!

Issue #3115, opened 2018-02-07
Regressions related to unlimited multiplayer


auriamg:
@Fantasmos this is related to your recent PR, it would be appreciated if you could take a look

* "Wrong way" message not centered correctly
![35842578-fb3916be-0b3d-11e8-8948-a87967b56010](https://user-images.githubusercontent.com/4296971/35894232-b35538f4-0b7f-11e8-98ab-da7961ee8a9f.jpg)

* spinner focus color missing
![no_color](https://user-images.githubusercontent.com/4296971/35894210-9fc03f0a-0b7f-11e8-84e0-dc9ac07248a2.jpg)

* powerups are off-center
![scotland-2018 02 06_20 50 37](https://user-images.githubusercontent.com/4296971/35894189-89d3662c-0b7f-11e8-830f-3054433522ba.jpg)

Fantasmos:
I'll try to take a look when I can, question however. What's wrong with the "wrong way" message. To me it looks fine, but I may not be remembering it previously correctly.

Benau:
previously it's centered vertically, now it's always at the top

auriamg:
@Fantasmos  Actually, the "wrong direction" and other similar messages used to be vertically centered, now they are at the top

PR #3116, opened 2018-02-07
Split screen regression fixes

Fantasmos:
https://github.com/supertuxkart/stk-code/issues/3115 Should tackle everything listed here.


src/guiengine/skin.cpp
```diff
@@ -1276,6 +1270,10 @@ void Skin::drawSpinnerBody(const core::recti &rect, Widget* widget,
     {
         params=&SkinConfig::m_render_params["spinner::neutral"];
     }
+    widget->m_skin_r = color_rgb.r * 255.0F;
+    widget->m_skin_g = color_rgb.g * 255.0F;
+    widget->m_skin_b = color_rgb.b * 255.0F;
```
deveee: I never really checked if uppercase "F" will work, but it just looks strange. We use lowercase "f" everywhere.

auriamg:
Thank you, everything seems fixed!

laizigouweijifen:
小白路过

Issue #3126, opened 2018-02-24
"Final lap!" message on one-lap races


ethanmuz:
## Description

When starting a one-lap race, there is a message when you cross the starting line "Final lap!". This is unnecessary because the player knows there is the only lap in the race.

## Steps to reproduce

Start a normal race with one lap only.

## Pull request

The change I proposed can be seen in pull request #3127. I made the change that only lets the "Final lap!" message appear when the race has more than one lap.

PR #3127, opened 2018-02-24
Remove "Final lap!" message in one-lap races

ethanmuz:
This change fixes the issue #3126 that I opened. From a player standpoint, I disagree with having the message "Final lap!" appear when starting a one-lap race. This change makes the "Final lap!" message only appear in races that have more than one lap.


auriamg:
Thanks, looks good!

PR #3130, opened 2018-03-10
Add unit testing to Travis - V2

qwertychouskie:
More readable, doesn't clone stk-assets if not needed, only ~one~two commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ matrix:
 
 env:
   matrix:
-    - BUILD_TYPE="Debug" SERVER_ONLY="OFF"
-    - BUILD_TYPE="Debug" SERVER_ONLY="ON"
-    - BUILD_TYPE="Release" SERVER_ONLY="OFF"
-    - BUILD_TYPE="Release" SERVER_ONLY="ON"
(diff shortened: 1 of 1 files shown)


auriamg:
I'm not sure travis-ci would allow us to download/use 2GB of space on their systems... even if they allow it, that's likely to make builds extremely slow, no?

qwertychouskie:
@auriamg There doesn't seem to be any limit that I could find for disk space in Travis.  A fresh clone of the assets repo currently clocks in at ~1.2GB.  In theory space should not be an issue, whole OSes are built on Travis (e.g. https://travis-ci.org/reactos/reactos).  The SVN clone adds ~5-6 minuntes to the build time (1)(2), so not great, but not unworkable either.

1: https://travis-ci.org/supertuxkart/stk-code/jobs/351849815 330.45 seconds
2: https://travis-ci.org/supertuxkart/stk-code/jobs/351849819 315.77 seconds

auriamg:
even so , it seems to me that unit tests should ideally be runnable without assets? why are the assets required?

qwertychouskie:
Closed in favor of https://github.com/supertuxkart/stk-code/pull/3328.

Issue #2777, opened 2017-02-08
Preferred GP AI winner


Alayan-stk-2:
Following a discussion on IRC with @hiker.

hiker: Alayan: you mean MK makes sure that the same two karts are always the first 2 ai karts?
hiker: Interesting ... I wonder how they do that :)
hiker: That would indeed be a nice tweak to make GPs more challenging

There are several possible implementations. The goal should be to do it only by affecting AI behavior, without cheats.

Ideas : 
- Don't select only one preferred AI kart, but two or three, so that the challenge is more consistent against random events.
- Make the other AI attack less the preferred karts
- Make the other AI dumber so that the preferred AI tend to be better ranked. To not make the SuperTux difficulty races easier than currently, new configurable AI improvements would be useful. In easier difficulty modes, the preferred AI could borrow some improvements of the next-tier of difficulty.

Another possibility would be to select more often than not the preferred AI among the better performing karts, but it's not wanted that there are better karts than other in the fist place.

Possible issues : 
- Making some AI karts dumber only in GP would create an unwanted-gameplay difference between normal races and GP races.

stkAnon:
I always figured individual AI strengths were already slightly randomized for every event...

For the possible issue, you could also apply these changes to normal races. The only difference in grand prixes would then be that the handicaps/bonuses are persistently given to the same karts for every race in the event.

Arthur-D:
I don't see how these suggestions are any less of "cheating" than doing it more directly, and it seems more complicated than it might have to be.

I suggest to select one AI randomly at start of GP which gets double the nitro when picking up canisters - not particularly visible what's going on, but it should do better on average than the other AI. Another possibility is to always give it 3 weapons when picking up items, maybe in combination with the nitro cheat above. But just increasing nitro might be enough.

Of course this would need to work consistently when saving/loading GP as well.

Alayan-stk-2:
There are now between 1 to 4 "preferred AI karts", depending on the total of karts in the race. At the beginning of each GP race, the N best ranked AI karts receive the boosted status, which improve their item usage, nitro usage, and don't make rubber-banding slow them down more when another AI kart overtake them.

So, technically, this PR is implemented.

However, the boosted AI status is not that strong currently for expert/supertux (it doesn't affect pathing/skidding), so I'll leave this issue open until this advantage is

Longer races would also probably help boosted AIs to be more reliably on top - and by longer, I mainly mean "3 minutes" instead of "<2 minutes".

[...]

Issue #2790, opened 2017-02-12
"About to pick an item" function


Alayan-stk-2:
Several AI improvements depend of the ability to know if the kart is about to pick an item : using an item before collecting a new one ; using a bubbleshield before driving on a bubblegum ; not using a switch before collecting nitro/item box (though could benefit of even more knowledge) ; not using nitro/zipper if going into a banana/bubblegum ; and probably some others I can't think of right now.

This is code I don't really know how to do.

Benau:
Sorry this is not bug but suggestion only

Issue #2789, opened 2017-02-12
Two (or more ?) levels of AI-item usage


Alayan-stk-2:
I started improving the AI item usage today. It struck me that it's the same AI used from intermediate to supertux.

The intermediate, and even expert, AI shouldn't master things perfectly (and the jump from novice where it is random would be harsh too).

I started by adding a new bool property to allow advanced mode for supertux (could be added to expert too if the lead devs so wish) ; however this may not be sufficient, especially depending on how we decide to tackle #2777.

So : 
- There should be at least two levels of AI-item usage.
- But it may not be sufficient. This is something I can not decide myself.

Benau:
Sorry this is not bug but suggestion only

PR #3143, opened 2018-03-15
AI item handling improvements

### LINKED ISSUE #3139: Inform the AI that it can't pass bomb back
When an AI kart has a bomb, he tries to give it to another kart by bumping it.

I think it's good gameplay that we can't give it back to the one who gave it to us (to allow back and forth with small consecutive bumps).

But the AI Currently isn't aware of this rule.

Therefore, it often engages in sharp turn to try to give it back immediately. Even if it touches back, it keeps the bomb. Worse, it often lose time because of the bad trajectory provoked by the sharp turn (often, I pass the bomb while going faster and hitting from the side, and the kart expecting to hit me back, instead of bumping on me, simply turn towards the edge of the road).

Alayan-stk-2:
A general summary of the changes in the patch : 

- As discussed on the IRC, replace the random item usage bool by an int for several modes (from 1 to 5 plus 0 for "do nothing"), to make each difficulty have its own item AI (with some shared parts).
- Add a boost_ai characteristic for karts ; allowing some karts to have a different AI in the same race. This is a preparation for a future fix to #2777 : make GP results less random and more challenging.
- Several AI improvements in bubbleshield, swatter and throwed items usage. Main changes summarized at the end.
- Collecting a different type of item resets the time of last use to remove unnecessary waiting.
- Different tiers of item usage ability set in place for some items
- Tiers of nitro usage revised
- Vast improvements of nitro usage
- HandleItems call placed inside HandleSteering to allow making it aware of items nearby
- Fixes #3139 : the AI won't try to pass back a bomb to the kart it got it from.
- Adds a small engine boost to nitro for balance reasons (making it useful at speeds other than max speed)

I've compiled and tested the patch.

<del>The shield activation near boxes/bad items needs improvement, but the overall behavior remains more effective.</del> Fixed with latest commit. There should be few false positives.

The items who have not yet been updated for the new system will be used like before AI-wise. So there should be no regression.

Main AI improvements : 
- Throwed items are used even if the shield is on if restricted mode is disabled ; 
- Shield and swatter are used to remove bad attachments ; 
- Shield is used before hitting a bad item or an item box
- The Switch is not used blindly anymore. The best AI now evaluates close items to make a decision.
- Nitro is used by bursts by checking on the remaining time of nitro effect
- Nitro is used when close to max speed or slow (to make use of the max speed or of the acceleration bonus), not when slightly below max speed (avoiding the AI using nitro when catched with a plunger)
- Several minor improvements


src/karts/controller/skidding_ai.cpp
```diff
@@ -1172,80 +1182,228 @@ void SkiddingAI::handleItems(const float dt)
...
+               m_controls->setLookBack(false);
+               break;
+           }
+           
+           // Check if a flyable (cake, ...) is close. If so, use bubblegum
+           // as shield
+           if(ai_skill == 3) //don't protect against cakes
+           {
+              if( !m_kart->isShielded() && projectile_is_close
+                 && (projectile_types % 1000)/100 == 0)
```
auriamg: Instead of all these weird decimal additions, wouldn't a bitmask be simpler??

src/karts/kart.cpp
```diff
@@ -470,6 +471,17 @@ float Kart::getSpeedIncreaseTimeLeft(unsigned int category) const
     return m_max_speed->getSpeedIncreaseTimeLeft(category);
 }   // getSpeedIncreaseTimeLeft
 
+// -----------------------------------------------------------------------------
+void Kart::setBoostAI(bool boosted)
```
auriamg: This method doesn't seem to ever be called? What is it for?

src/items/projectile_manager.cpp
```diff
@@ -153,3 +153,44 @@ bool ProjectileManager::projectileIsClose(const AbstractKart * const kart,
     }
     return false;
 }   // projectileIsClose
+
+// -----------------------------------------------------------------------------
+/** Returns ABCD with D, C, B and A the numbers of bowling balls, plungers, cakes
```
auriamg: This weird decimal addition is weird, wouldn't a bitmask be much simpler? Or an array like byte[4] ?

src/karts/controller/skidding_ai.cpp
```diff
@@ -1172,80 +1182,228 @@ void SkiddingAI::handleItems(const float dt)
...
+                                                      m_successor_index[node]);
+        node = m_next_node_index[node];
+        // Stop when we have reached the last quad
+        if(node==last_node) break;
+    }   // while (distance < max_item_lookahead_distance)
+
+    //items_to_avoid and items_to_collect now contain the closest item information needed after
+    //What matters is (a) if the lists are void ; (b) if they are not, what kind of item it is
+   
+    switch( m_kart->getPowerup()->getType() )
```
auriamg: Since this code has grown a lot, it may make sense to split the handling of each powerup type in its own method to avoid this method growing too big

src/karts/controller/skidding_ai.cpp
Alayan-stk-2: I wanted to pass only one int arguments. The decimal trick was meant to allow this. I agree it's rather rough.

A bitmask would certainly work ; I simply have no experience using it.

src/karts/kart.cpp
Alayan-stk-2: Currently, it is indeed not called anywhere.

It is destined to set a given kart as using an improved AI. The expected use of this feature is in GPs : setting some karts to have a slightly better AI will allow them to have better and more consistent results compared to other AI karts, making harder to win a GP (#2777)

It is present in a patch because it is needed to make boosted_ai complete, which is introduced to determine the level of the AI item handling.

src/items/projectile_manager.cpp
Alayan-stk-2: I've a personal dislike for arrays making me avoid them when I can. The bitmask would certainly be less convoluted.

src/karts/controller/skidding_ai.cpp
Alayan-stk-2: Indeed, separate methods for each powerup could be a good idea. Especially as I expect it to grow even more.

src/karts/controller/skidding_ai.cpp
Alayan-stk-2: I looked up bitmasks.

I've not found clear examples and I think highly likely I would introduce some defect if I tried to do it.

However, if it is done, and some of my modulo/division tests are replaced by bitmasks, I think I should be able to imitate them for other cases.

Alayan-stk-2:
Suggestions #2789 and #2790 (which had been closed for being suggestions and not bugs) are in-effect solved by this pull request.

auriamg:
I got the following crash :
```
>	supertuxkart.exe!LinearWorld::getEstimatedFinishTime(const int kart_id) Line 422	C++
 	supertuxkart.exe!SkiddingAI::handleNitroAndZipper() Line 2059	C++
 	supertuxkart.exe!SkiddingAI::update(float dt) Line 367	C++
 	supertuxkart.exe!Kart::update(float dt) Line 1317	C++
 	supertuxkart.exe!World::update(float dt) Line 1000	C++
 	supertuxkart.exe!LinearWorld::update(float dt) Line 161	C++
 	supertuxkart.exe!World::updateWorld(float dt) Line 871	C++
 	supertuxkart.exe!MainLoop::updateRace(float dt) Line 155	C++
 	supertuxkart.exe!MainLoop::run() Line 277	C++
 	supertuxkart.exe!main(int argc, char * * argv) Line 1828	C++
```

```
    assert(m_kart_info[kart_id].m_race_lap == race_manager->getNumLaps()-1);
```
m_race_lap  is 0 , but getnUmLaps() is 3

auriamg:
I'd like to get @hiker 's opinion if possible, as far as I'm concerned I think this isnow mergeable

Alayan-stk-2:
I did test watching the AI race.

The nitro reserve code seems to not work as expected : the AI is indeed building a reserve ; it is indeed using immediately any amount exceeding the reserve... but it is currently not using the reserve at the end of the race.

I think I know why : I decrease a variable as if it was a global variable, while in fact it is reseted at each call.

Alayan-stk-2:
There seems to be an issue with the AI not using the parachute item. It puzzles me a bit as I didn't change much that part of the code. I'll have a look at it.

May be related to the overall waiting time issues which were put in place to prevent using twice the same item in a row but who also affect items immediately after collection.

Alayan-stk-2:
The parachute issue I noticed was unrelated to the fact that the AI had a tendency to wait between usage of items of different type. See #3150

The latest changes to remove unnecessary waiting remain however useful  to improve the AI.

src/items/projectile_manager.cpp
```diff
@@ -153,3 +153,44 @@ bool ProjectileManager::projectileIsClose(const AbstractKart * const kart,
...
+
+    for(Projectiles::iterator i  = m_active_projectiles.begin();
+                              i != m_active_projectiles.end();   i++)
+    {
+        float dist2 = (*i)->getXYZ().distance2(kart->getXYZ());
+        if(dist2<r2)
+        {
+            if ((*i)->getType() == PowerupManager::POWERUP_BOWLING)
+            {
+                counter = counter + 1;
```
auriamg: I still think this should be converted to a bitmask or an array. You stated that you have a 'personal' dislike for arrays, that doesn't seem warranted as the array version would IMO be much cleaner than this

src/karts/controller/ai_properties.hpp
```diff
@@ -111,11 +111,11 @@ class AIProperties
...
     bool m_handle_bomb;
 
-    /** True if items should be used better (i.e. non random). */
-    bool m_item_usage_non_random;
+    /** Determines the strategies used by the AI for items. 0 is worst, 5 is best. */
+    int m_item_usage_skill;
 
     /** How the AI uses nitro. */
-    enum {NITRO_NONE, NITRO_SOME, NITRO_ALL} m_nitro_usage;
+    int m_nitro_usage;
```
auriamg: If you remove the enum then please at least explain what the various levels mean in a comment

src/karts/kart.cpp
```diff
@@ -508,6 +523,15 @@ void Kart::setPowerup(PowerupManager::PowerupType t, int n)
     m_powerup->set(t, n);
 }   // setPowerup
 
+// ----------------------------------------------------------------------------
+/** Sets the powerup this kart has last used. Number is always 1.
+ *  \param t Type of the powerup.
+ */
+void Kart::setLastUsedPowerup(PowerupManager::PowerupType t)
+{
+    m_last_used_powerup->set(t, 1);
```
auriamg: You appear to use a 'full' Powerup object but only use it to store the TYPE of powerup last used, why? Storing the type might be simpler

hiker:
Thanks a lot for your work so far, and my apologies that I am rather busy atm.
A few comments:
1) Before we merge this, we should do some regression tests, by running a set of races of old AI against new AI. We have the special TestAI (which should be identical to the current AI), so we can compare. I don't expect a significant improvement (or make this a requirement for this patch to be accepted) by this work on items (though am happy to be surprised and proven wrong :)  ), I expect this to be more subtle. But I want to make sure we don't have any regressions. There us a script in tools/ai_test, which might tell us if we have some problems (and the statistics might even hint at the reason for this).
2) I noticed that now and again (often??) you have several commits with the same commit message. Are you aware that you can do a 'git commit --amend' to modify the last local commit, and maybe push less frequently? This would allow you to keep all the commits that belong together in one commit with one commit message, which makes it much easier to look at the commits.
3) I also agree with auria re the enum - they are very useful to understand code. I haven't had time to look at your patch in detail, but remember noticing a comparison with an enum being replaced with a comparison with 0 - unless there is a good reason not to use enums, I would prefer to leave them in.

src/items/projectile_manager.cpp
Alayan-stk-2: I'm not someone who code much. So I tend to stick to what I understand. I'd probably introduce bugs if I tried to use one of the two cleaner solutions.

Alayan-stk-2:
The latest changes, aside from fixing some errors, make the AI nitro code depend more on the kart-properties linked to nitro. It should work better across the different types of karts (light/medium/heavy) and be more resilient if the config files are tweaked.

From my tests, all the nitro features seem to finally work as intended. I don't expect to do more changes to it.

Alayan-stk-2:
@hiker, to answer to your comments ;

1. No problem, I understand the need of regression tests. I may try to run the tool on my own, but I guess you may wish to do so too.
2. This happens when I change several files to get a change done. I'm pushing things here directly, not via git, which is why it's not condensed in a single commit. Contrary to what I did at the start (where you can see all the "fix typo/crash/compilation" commits), I test the compilation and effect of the changes on my local copy to catch errors and fix more of them before pushing it there.
3. The issue is that it becomes hard to describe clearly what the level correspond to in an enum. Currently 0 is none, 1 is all immediately, while 2 to 4 varies in more subtle ways which can't be explained in a single word. If it's just saying "bad", "good", "best", it's not more informative than a number. If there is a need for code comments to explain what the enum options stands for, then I don't think the enum adds much if anything. However, the enum requires significantly more code/heavier code to be handled. Replacing "ai_skill >= 2" by "m_ai_properties->m_nitro_usage==AIProperties::NITRO_BAD || m_ai_properties->m_nitro_usage==AIProperties::NITRO_GOOD || m_ai_properties->m_nitro_usage==AIProperties::NITRO_BEST" would make things much harder to follow imho.

Alayan-stk-2:
Sorry for the fuss, I was tired and got confused with pointers.

Alayan-stk-2:
@hiker I had some difficulties to run the tool, but after adding ./supertuxkart before the parameter and launching it from the same directory as the bin, I got it to work.

I will note at first that the best comparison is of top AI to top AI. By default, the tool compares at difficulty 2. I ran it at difficulty 3 and I also edited the xml : by default, the new AI do not use the very best nitro and item handling (4 and 5), but those belows (3 and 4). This is intended to permit to introduce the preferred GP winners using superior AI. But to compare the ability of each AI, it seems better to run each at its very best.

I also ran it at only 8 laps per race, to reduce a bit the runtime.

You wrote you "don't expect a significant improvement".

I think you'll be pleasantly surprised then ! 

I think that while the item handling helped, the greater part of the improvement come from the nitro handling changes. For a casual player, it may not be evident how badly the AI used it before, but the test results are eloquent.

If I had to describe it in one word, it would be : bloodbath.

The city and farm track failed (because the tool is old and the tracks were replaced, I think ?). Otherwise, I got those results : 

stk_ai_1

(higher is better, +10 means that the new AI karts were ranked 1, 2, 3, 4 and 5 out of 9.)

stk_ai_2

stk_ai_3

2,5% time less is very significant in racing, as exemplified by the positions results. As an experimented player, I often manage 10-15% better than the old SuperTux AI, so there is quite a margin still, but it's a good leap.

For the tracks where the new AI underperformed, there is a clear pattern.

In the mines, there were 2,6 rescues in average  for the new AI (13 total) against 1,25 average (5 total). Maybe the additional speed backfired, maybe a random fluke. The AI has notoriously issues staying on track there.

Oliver Maths, Minigolf, Mansion, and XR591 are also notoriously tracks where turning well is important. The old AI already had significant issues there, and the additional speed seems to backfire there.

However, even in most of these worse cases, the new AI still matches the old one in average time. The exception in Mansion is due to one kart who has been rescued 107 (!) times in the race for a 390 seconds total. That's not a new AI issue.

deveee:
I also suggest to work on private/local repository and make a pull request with few commits (instead of 80), for example:
- fixed typos that were unrelated to my work
- fixed a bug that I found when I was looking at the code
- my AI work that improves xyz etc...

and make the pull request when you think that it's ready to review. Because I have no idea if it's final version, probably not ;)

And I fully agree with arrays/bitmask/enums.

Alayan-stk-2:
New AI testing with : 

- Updated track list in the test tool to also test Candela City, Cornfield Crossing and Volcano Island
- 10 laps per turn

Result : 

stk_ai_1b

stk_ai_3b

Alayan-stk-2:
Testing with items enabled is much more random than in time trial mode.

This is expected, especially as items vary according to rank : a failing AI will receive better items to compensate.

Still, if there are improvements, they should be visible on average.

I ran 4 tests of 10 laps and averaged the results. They are less impressive than the time trial ones, but still clear. Maybe more tests with a little less laps would yield more consistent results. Anyway : 

stk_ai_1c
stk_ai_2c
stk_ai_3c

The other stats from the tests give some insight as to what happen. In oliver math's, nitro is used but then the kart brake. There are 30%+ brake for the new AI in that track.

The tracks having the better and worst results are partially different from the time trial tests. Certainly a mix of randomness and a real difference in how tracks are played with or without items. Mansion, which was the worst track for the new AI due to the kart who got stuck and rescued 100 times, becomes the best in time % change...

src/items/projectile_manager.cpp
```diff
@@ -153,3 +153,32 @@ bool ProjectileManager::projectileIsClose(const AbstractKart * const kart,
...
 }   // projectileIsClose
+
+// -----------------------------------------------------------------------------
+/** Returns an int containing the numbers of a given flyable in a given radius
+ *  around the kart
+ *  \param kart The kart for which the test is done.
+ *  \param radius Distance within which the projectile must be.
+ *  \param type The type of projectile checked
+*/
+int ProjectileManager::projectileCloseType(const AbstractKart * const kart,
```
auriamg: Very minor, but 'projectileCloseType' doesn't sound like correct english to me. GetNearbyProjectilesCount maybe ?

src/karts/controller/skidding_ai.cpp
```diff
@@ -1132,24 +1140,36 @@ void SkiddingAI::evaluateItems(const Item *item, Vec3 kart_aim_direction,
 
 //-----------------------------------------------------------------------------
 /** Handle all items depending on the chosen strategy.
- *  Either (low level AI) just use an item after 10 seconds, or do a much
- *  better job on higher level AI - e.g. aiming at karts ahead/behind, wait an
- *  appropriate time before using multiple items etc.
+ *  Level 0 "AI" : do nothing (not used by default)
+ *  Level 1 "AI" : use items after a random time
+ *  Level 2 AI : strategy detailed before each item
```
auriamg: Maybe tweak the comments slightly to make it clearer that higher numbers = better AI

src/karts/controller/skidding_ai.cpp
```diff
@@ -1172,80 +1192,252 @@ void SkiddingAI::handleItems(const float dt)
...
+      {
+         ai_skill = 5;
+      }
+      else
+      {
+         ai_skill = m_ai_properties->m_item_usage_skill;
+      }
+   }
+   
+   if (m_kart->getBoostAI() == true)
```
auriamg: Very minor, could be written more concisely:

if (m_kart->getBoostAI() && ai_skill < 5) ai_skill++;

src/karts/controller/skidding_ai.cpp
```diff
@@ -1172,80 +1192,252 @@ void SkiddingAI::handleItems(const float dt)
...
+        return;
+    }
+
+    // Tactic 1: wait between 5 and 10 seconds, then use item
+    // ------------------------------------------------------
+    if(ai_skill == 1)
     {
-        if( m_time_since_last_shot > 10.0f )
+       int random_t = 0;
+       RandomGenerator g = RandomGenerator();
```
auriamg: Don't create a new one everytime, the way RandomGenerator is currently implemented this will not end well

src/karts/controller/skidding_ai.cpp
auriamg: My previous comment about splitting the by case-when into sub-methods still holds

src/karts/controller/skidding_ai.cpp
```diff
@@ -1368,14 +1568,131 @@ void SkiddingAI::handleItems(const float dt)
...
-            m_controls->setFire(true);
-        break;   // POWERUP_SWITCH
+        {
+            // It's extremely unlikely two switches are used close one after another
+            if(ai_skill == 2)
+            {
+                if (m_time_since_last_shot > 2.0f)
+                {
+                    m_controls->setFire(true);
+                   break;
```
auriamg: Minor, but please watch the indentation

src/karts/controller/skidding_ai.cpp
```diff
@@ -1368,14 +1568,131 @@ void SkiddingAI::handleItems(const float dt)
...
+               int good = 0;
+
+               //Good will store 1 for nitro, big or small, 2 for item box
+               //Big nitro are usually hard to take for the AI
+               for(i=items_to_collect.size()-1; i>=0; i--)
+               {
+                   if (items_to_collect[i]->getType() == Item::ITEM_BONUS_BOX)
+                   {
+                      good = 2;
+                     i = -1;
```
auriamg: wonky indentation

src/karts/controller/skidding_ai.cpp
```diff
@@ -1582,107 +1924,235 @@ void SkiddingAI::handleRescue(const float dt)
...
-    // If a parachute or anvil is attached, the nitro doesn't give much
+   
+    // Don't use nitro or zipper when the AI has a plunger in the face!
+    if(m_kart->getBlockedByPlungerTime()>0)
+    {
+        if ((nitro_skill < 4) && (ai_skill < 5))
+        {
+           return;
+        }
+       else if (nitro_skill < 4)
```
auriamg: wonky indentation

src/karts/controller/skidding_ai.cpp
```diff
@@ -1582,107 +1924,235 @@ void SkiddingAI::handleRescue(const float dt)
...
+        float burster;
+        
+        if ( nitro_time > 0)
+        {
+           burster = 2*nitro_max_time;
+        }
+       else
+       {
+            burster = 0;  
+       }
```
auriamg: indentation

src/karts/kart.cpp
```diff
@@ -118,6 +118,7 @@ Kart::Kart (const std::string& ident, unsigned int world_kart_id,
     m_max_speed            = new MaxSpeed(this);
     m_terrain_info         = new TerrainInfo();
     m_powerup              = new Powerup(this);
+    m_last_used_powerup    = new PowerupManager::PowerupType();
```
auriamg: You don't need to "new" an enum, or use a pointer to an enum, this has no reason to be on the heap

src/karts/kart.cpp
```diff
@@ -289,6 +291,7 @@ Kart::~Kart()
     delete m_max_speed;
     delete m_terrain_info;
     delete m_powerup;
+    delete m_last_used_powerup;
```
auriamg: Same as above, this has no reason to be on the heap

src/karts/kart.hpp
```diff
@@ -86,6 +86,9 @@ class Kart : public AbstractKart
 
     /** Handles the powerup of a kart. */
     Powerup *m_powerup;
+    
+    /** Remember the last **used** powerup type of a kart for AI purposes. */
+    PowerupManager::PowerupType *m_last_used_powerup;
```
auriamg: Same comment, no reason to be a pointer

src/karts/kart.cpp
Alayan-stk-2: Right, I changed the occurences everywhere when I converted it from a full Powerup type to a simple PowerupType without thinking about this. I'll remove it.

src/items/projectile_manager.cpp
Alayan-stk-2: Done

src/karts/controller/skidding_ai.cpp
Alayan-stk-2: Done

auriamg:
@hiker Since it doesn't seem like you've had the time to review it, I've just merged the PR to avoid letting it unattended for too long.

Issue #2592, opened 2016-08-03
Artist debug mode menu does not have enough camera/kart target entries


RQWorldblender:
Up to 20 karts, including the player ones, are supported. However, the debug menu lists only the first 8 karts, and when I start a race with more than 8, and switch to any of those targets, it can become impossible to switch back to other AI karts 9 up to 19, or even the player kart(s). I think that either the debug menu needs to have either 20 static menu entries for the 20 supported karts, or more optimally, show entries based on how many karts are in play, so that, for example, if there are only 4 karts, then only 4 entries should be listed, or if there are 16 karts, then 16 entries should be listed. The screen size should also be taken into consideration, so that lower entries can reached at all.

Benau:
Actually I added this function for debug (AI debug) mostly, not for watching real race for normal players.

hiker:
Those are debug functions, which assume a proper knowledge of what is happening, and when it is safe to use them. Unless you have an STK development need to do, we won't spend any time on this.

PR #3157, opened 2018-03-20
Allow to change camera view in debug mode to any of up to 20 karts

Alayan-stk-2:
Currently the limit is of eight.

This has two downsides : a)if there are 9 karts or more, it's currently impossible to go back to the view of the player kart and to drive with it. In some debugging/testing situations, it's helpful to be able to switch back and forth.
b)It's impossible to watch how the AI in bad positions acts if there are 9-10 karts or more. Not a huge downside for the moment, but it may help for some situations if what happens in races with a lot of karts is different to what happens when there are less of them.

The "change camera target" submenu is positioned higher than before. to make sure all its options are displayed in Xx768 resolutions or higher. It doesn't fully fit in 800x600 (it goes up to seventeen) ; but I don't think it's a huge issue.

### DIFF
--- a/src/utils/debug.cpp
+++ b/src/utils/debug.cpp
@@ -130,6 +130,18 @@ enum DebugMenuCommand
     DEBUG_VIEW_KART_SIX,
     DEBUG_VIEW_KART_SEVEN,
     DEBUG_VIEW_KART_EIGHT,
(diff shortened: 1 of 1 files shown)


Benau:
https://github.com/supertuxkart/stk-code/issues/2592

PR #3170, opened 2018-04-01
Makes nitro increase slightly acceleration

Alayan-stk-2:
New PR opened as requested, as the change was cut out being not AI-related.

Summary of the issue : in current STK, nitro increases max speed, and only max speed. Therefore, trying to use nitro to accelerate when at a slow speed is absolutely useless, it doesn't change at all the kart's speed.

This is extremely counter-intuitive (casual STK players are probably not aware of it and use nitro in situations where it is useless), and also makes the optimal utilization of nitro "use only when at max speed" without any sort of nuance.

The proposed patch gives a slight boost to acceleration for nitro. The speed gained when going already fast is small ; it makes the kart accelerate to the new max speed obtained faster but by a small amount. However, when at a slow speed, and trying to accelerate to max speed (an acceleration which takes more time), the boost is more significant.

The goal is to make it have some use when trying to reaccelerate (like after rescue) without otherwise upsetting the game-balance too much.

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -2460,6 +2460,10 @@ void Kart::updateEnginePowerAndBrakes(float dt)
     updateNitro(dt);
     float engine_power = getActualWheelForce();
 
+    // apply nitro boost if relevant
+    if(getSpeedIncreaseTimeLeft(MaxSpeed::MS_INCREASE_NITRO))
+    engine_power*=1.2f;
+   
     // apply parachute physics if relevant
     if(m_attachment->getType()==Attachment::ATTACH_PARACHUTE)
         engine_power*=0.2f;


hiker:
I admit I haven't tried it, but the idea is that nitro already adds engine force. Does this not work? If the effect is not strong enough, we can just tweak the settings in kart_characteristics.xml.

Alayan-stk-2:
If it is already supposed to exist, then it certainly does not work.

Alayan-stk-2:
The additional engine power boost may help to unlock the increased max speed, but its effect in normal acceleration is really small.

I've reworked code to have an engine-mult characteristic (instead of hardcoding the multiplicator) and fixed the comparison (in case the time left is negative). It's important to have a multiplication and not an addition for it to work well.

I've worked on my slipstreaming branch (as anyway the modified files here are the same) ; I'll submit the changed files when #3186 is merged.

Alayan-stk-2:
Closing this PR now, as the code change proposed is not the most clean way to do it and I've better code to submit later.

Issue #2339, opened 2015-10-07
Numeric values for the speedometer


RQWorldblender:
It's hard for me to tell precisely how fast I drive a kart (either in metric or imperial units), as the place where the numeric speed value used to be is replaced by the rank number, which I personally think could be moved next to each kart icon in the rank area. Next to the speedometer, there is the nitro meter, which I also have no way to tell the numeric value of. This makes it hard for me to tell whether I have collected enough nitro in the challenges that involve collecting them. To make it clear on what the values represent, I prefer that units accompany them in order to reduce ambiguity, and those should be placed in areas that are comfortable to place and read.
I may like to see showing/hiding them as an option so that these numeric values wouldn't cause for those who like the current speedometer setup.

![screenshot from 2015-10-06 20 47 46]
(https://cloud.githubusercontent.com/assets/13372387/10327330/a53f43b4-6c6b-11e5-941f-e2fedc615260.png)

deveee:
A lot of people complain about speedometer which shows rank number instead of speed value. I personally was thinking that it is simply broken and always shows me speed value "1" :P But it is actually minor thing and if some people decided that it should show rank number, it's not a big problem for me.

Adding an option to show/hide numeric values doesn't have sense for me - it's just a speedometer.

samuncle:
It would be better IMHO if it display the speed instead of the rank. It's very confusing I just learned now it display the rank. I thought it was broken :P

RQWorldblender:
Well, I gotta see why the rank number would be displayed if not the speedometer and/or the nitro value will be brought back. I will accept the defense if I think it will make sense.

hiker:
Not knowing exactly how fast you are is done on purpose: STK is not a realistic simulation, it focuses on fun and just driving - numbers are not really necessary for that. If you play with more karts, not all karts are shown in the rank area, so we can't display it there anyway.

Nitro display: similar reason as speed - we want to focus on 'just drive', not on optimising the last drop of nitro (together with air pressure, terrain, ...). There are games that focus on more realistic simulations, but stk is not one of them.

Challenges display a line at the target value, so that should be easy to see.

PS: Sorry, I forgot to publish this comment when closing the ticket

deveee:
@hiker you actually didn't close this ticket. Not sure if you wanted to do it or not.

hiker:
[...]

PR #3177, opened 2018-04-03
Improved speedometer and nitrometer

### LINKED ISSUE #2822: Show the number of nitro points you have, vs how much you need in a nitro challenge
Especially when playing XR591 in expert, it gets hard to see how many nitro points you have vs how many you need, especially as the bar gets closer...

It might make it a bit better to show a numerical counter of how many you have, vs how many you need.

### LINKED ISSUE #3171: Increase the max speed that the speedometer can display
Given #2339 was refused ; the only speed cue we have in race is the visual speedometer with the bars in an arc of circle.

3 bars correspond to 10 of speed (approximately : they are filled by computing an angle, so it doesn't jump immediately from a bar to another, and the bars themselves are not perfectly aligned) ; and the speedometer has 9 bars ; making the max speed it can display 30.

But in SuperTux, the base speed is 25. This means that when the kart gets a skid/nitro/zipper boost, the 9 bars are all filled. This makes the speedometer not useful to quickly gauge what's the kart speed, or if maybe one bonus is fading out.

The max speed attainable by combining nitro, red skidding and zipper is very high, and it may be excessive to ask for the speedometer to display it. However, situations where two out of three of these boosts are cumulated are common, especially nitro and skidding ; and should be better supported. I'm not sure what exact max displayed value would be best ; somewhere between 35 and 40 probably.

### LINKED ISSUE #3173: The nitro bar is not very informative
A kart can hold at max 20 nitros, and the nitro bar reflects this.

However, while it fills up, it's difficult to gauge really how much nitro a kart has.

- The nitro bar is not divided in segments like the speed bar. 5 segments of 4 would be simple and useful.
- The thickness of the bar vary. This is misleading, as one may think that the quantity of nitro is proportional to the area and not to the angle covered. A bar with a constant thickness would be less ambiguous.
- It fills itself very unevenly (probably related to a misplaced center point). It fills itself normally at the start, then seems to fill itself quickly from 9 to 13 nitros, and then very slowly up to 20. This is very misleading. See the video : https://www.youtube.com/watch?v=T36faaq8cGc

Alayan-stk-2:
Summary : 

- Fixes #3171 (new max displayed speed : 400. Previous 305. Base speed ranges from 150 to 250 ; bonuses from two sources are usually under 150 (unless one is a booster giving nearly this by itself) ; from three sources it can get near 250)
- Fixes #3173 
- Probably fixes #2822 ; as the demand of a numerical counter was motivated by the very slowly moving nitro bar when getting the last cans and by the overall inconsistencies of the meter, making it "hard to see how many nitro points you have vs how many you need" in nitro challenges (this is not the case anymore)
- Makes the speed progression of the speedometer significantly more regular (previously ; two identical angular progression could represent a speed delta up to 1.8x different. Now it is up to 1.25x. Doing better is possible but would require using a better method to avoid excessive code growth)
- The nitrometer inner circle is the same as the speedometer inner circle to make it look better.
- Center the rank display inside the meters inner circles (I'd make it bigger and prettier too, but this is not the point of this PR)

Video showcasing the changes : https://www.youtube.com/watch?v=IINFNcThKb4

Screenshoot showcasing the changes (because the video is not of the highest quality) : 

stk speedometer showcase

Known issues : 
- The advanced rendering engine saturates the speedometer texture, which makes the gradient look worse.


qwertychouskie:
IMO it looks too round/flat, especially the nitro part (fixed width).  The old one had a very subtle sense of being 3D, it would be nice if that could be kept.  Also, it may make sense to have something like this (quickly thrown together in Gimp):

speedometer concept grey background

This would probably solve https://github.com/supertuxkart/stk-code/issues/2339 (and some potential confusion).

Alayan-stk-2:
The fixed width for the nitro bar is a feature, not a bug. It's a gauge, and having a changing width is confusing : does the total amount scale with area ? With length ? By having the two proportional, this ambiguity is fully eliminated.

Is there any difference I'm supposed to see between the current master version and your mockup except the addition of the "/rank" to reduce confusion for rank ? (Which I entirely agree with, by the way, but this is not the topic of the PR. I've been told to not mix up too much things in a PR and I'll stick to it.)

Using non-circle ellipses makes much much more painful to have them filled accurately ; because a given angular change corresponds to a different length delta. Think of it as an orbit - it goes faster when closer to the center. As the drawing method rely on polygons, I don't even want to think the complications this would imply.

As for the flatness, there is absolutely no difference with the previous version on this side : a meter with a color gradient on a semi-transparent background. The background shadows may be made slightly larger if desired, though (very easy to change). Coloring changes (which may include fake 3D effects if desired) are significantly easier than change

I don't want to sound stubborn ; I already made significant changes from the first version and I may do others even if it's time consuming ; but proposed tweaks should keep in mind the purposes of the changes (clarity, accuracy), not defeat them.

auriamg:
It looks better than your previous attempt, but TBH it definitely looks like developer art, as opposed to something made by an artist... Even though your suggested one is probably better for the small percentage of our players that care about numbers and metrics, I expect the majority of our player base will find the current/previous one to look better. Other opinions welcome, though

Alayan-stk-2:
Some other "fun" racing games don't have any speed gauge. You basically accelerate to max speed in a blink, do funky things which gives boost, but it's all fuzzy. Imho, STK strikes a better balance.

I think there is truth in your "developer art" remark, but the opposite side of this medal is that, in the specific case of something like a meter, the artist's version gives very poor data because the appearance takes precedence over the function.

In my opinion, the whole point of a gauge/meter is to display (reasonably) accurate information. It should try to be nice and elegant, and it should integrate with the game's visual style ; but there is little point in having a gauge/meter which blatantly lie to the player. Maybe a majority would prefer the current look, but I think all the players frustrated in the nitro challenges would have much preferred my version. (Also, if you want more of my "developer thinking", raw numbers should be weighted by other factors. The satisfaction of someone playing an hour counts less than of someone playing ten hours.)

I'm open to any suggestion to improve the "look and feel" part as long as it saves the functionality.

From what I can think of : I could change where the nitro gauge starts and end. Currently it's a perfect quarter of circle, maybe if it went further left and slightly further up it would seem better, "less serious" ?

Alayan-stk-2:
The code for the poly drawing is very redundant (we have nearly identical versions at three places ; two for the nitrometer and one for the speedometer).

I'm moving this to a dedicated function. This will make the code shorter and clearer ; and also reusable.

It will also make it a lot easier to use a few more points for even more accuracy (the x1.25 of the speedometer is now low enough that it's hard to feel even when looking for it, especially as it moves up and down from x1 to x1,25 so on any significant delta it tends to average out ; but if it can be improved for cheap, why not ?).

Alayan-stk-2:
New dedicated function done. It should make the code much more readable and easy to tweak.

Alayan-stk-2:
New 3D effect for the speedometer : 

speedometer_3d

(I'll integrate the code changes later to don't have too much commits)

Alayan-stk-2:
The nitro bar has been completely redone to better match the speedometer and to look nicer.

new meter

Video showing the meters in action : https://www.youtube.com/watch?v=KcU0tticthI

psypherium:
Is there any way to pivot the axis? I can see the flatness, the red part should be more in the distance than the yellow part. Or maybe it could cast a shadow? Also I think it should be 1/2 instead of just 1. Maybe the nitro bar could get fatter at the bottom also, so they match up?

Alayan-stk-2:
@psypherium When you say "pivot the axis", you mean, creating some perspective effect as if the top was (slightly) farther than the bottom ? I guess in theory this is possible, as it would be applying a specific projection on the picture and the same to the coordinate used to precisely display the speed based on the circle. I may try this but I need some words from @auriamg or @hiker about if they wish me to do so before doing it, as it would be significant work.

Not sure about how you envision the shadow.

For the nitro bar, I want to keep the same thickness all along so that area is proportional to length and there is no ambiguity on the amount of nitro. I may add something just beneath it to complete the circle, I'll think about it.

If by "1/2", you're speaking about the rank, I agree that this need change, but this is not directly related to the speedo/nitrometer themselves, so I won't try to add a fix for this in this PR.

For the digits, I'd think rendered according to an IRC discussion I had recently, but I've not checked specifically to confirm.

deveee:
Just note that nitro meter is also used in android race gui, not sure if you checked it. It can be enabled in config.xml.

Alayan-stk-2:
@devee I've tested under various resolutions for the normal computer UI ; but I've indeed not conducted tests with an android device.

There is an option in config.xml to launch the UI "as if" it was an android device ?

Alayan-stk-2:
I don't know in which node to put this in.

PR #3178, opened 2018-04-03
Better GP points distribution

Alayan-stk-2:
Revisiting my contribution from last year, I've improved the GP points distribution.

The improvements are twofold : 

- Better reward victory and the places close after ; as the previous distribution wasn't aggressive enough, especially with a lot of karts.
- Smooth significantly the point rewards

This is with the old distribution : 

gp_old

This is with the new one : 

gp_new

The total of points distributed don't increase too much (especially with 20 karts or less : 10 karts : the 1st get from 16 to 18; 15 karts : 28->30 ; 20 karts : 39->50) ; as the additional high value nodes are compensated by less middle-value nodes ; so the player won't be confronted with excessive numbers.

### DIFF
--- a/data/stk_config.xml
+++ b/data/stk_config.xml
@@ -38,29 +38,29 @@
     <points points="2" /> <!-- added with 5 karts -->
     <points points="1" /> <!-- added with 6 karts -->
     <points points="3" /> <!-- added with 7 karts -->
-    <points points="2" /> <!-- added with 8 karts -->
+    <points points="1" /> <!-- added with 8 karts -->
     <points points="3" /> <!-- added with 9 karts -->
-    <points points="1" /> <!-- added with 10 karts -->
-    <points points="4" /> <!-- added with 11 karts -->
+    <points points="4" /> <!-- added with 10 karts -->
+    <points points="1" /> <!-- added with 11 karts -->
     <points points="2" /> <!-- added with 12 karts -->
-    <points points="1" /> <!-- added with 13 karts -->
-    <points points="3" /> <!-- added with 14 karts -->
-    <points points="2" /> <!-- added with 15 karts -->
-    <points points="1" /> <!-- added with 16 karts -->
+    <points points="5" /> <!-- added with 13 karts -->
+    <points points="1" /> <!-- added with 14 karts -->
+    <points points="3" /> <!-- added with 15 karts -->
+    <points points="6" /> <!-- added with 16 karts -->
     <points points="4" /> <!-- added with 17 karts -->
     <points points="2" /> <!-- added with 18 karts -->
-    <points points="3" /> <!-- added with 19 karts -->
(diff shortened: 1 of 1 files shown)


Mrxx99:
Does that fix that you can still win a GP if you skip tracks? I think that should be impossible. Maybe your newly introduced "preferred" winner will also help avoiding this.

Alayan-stk-2:
No, that don't change the possibility to win a GP if you skip tracks. If anything, it slightly increase it, as the GP skips rely on "win some race, skip others", and this make winning more relevant.

Skipping a race means finishing last and forfeiting all the possible points in the track, so if the player still manage to get over his opponents, it's that he's way above them. That said, I have my subjectivity on the topic - I got the fastest STK 0.9.3 full story mode completion ever by mastering those skips.

What however should, in my opinion, be changed is the current possibility to perfectly manipulate the AI's final place in the race by forfeiting it at a good moment (currently, the AI finishes the race exactly in the order they were when you forfeit/finish yourself the race). This can [be quite skillfull](https://youtu.be/6-Z5YlTk7qk?t=47m00s) but it's also completely exploitative and can in some circumstances make a GP easier to win (say you had bad luck/made mistakes and are going to lose the race ; but you only need to not lose to Konqi by more than 5 points to win the GP, and skip right at the moment he has lost time from a basket ball).

The "preferred winner" is not yet in place. There is a preparatory step that has been done, but not more.

Alayan-stk-2:
Here is the same table as above, but with the GP point distribution from STK 0.9.2 and older, for comparison : 

gp_old_old

You can see that it was much worse at differentiating the ranks from 5th onwards (the line 5 is for the difference between the 4th and the 5th) and far from following an exponential distribution ; but that it was slightly better than the current solution for the very first places. This is because a difference of 3 points is more significant when the first has 25 (it was 26 ; but the last had 1 ; so for comparison purposes you normalize it by subtracting 1) than a difference of 4 when he has 39.

The differences between the GP distribution for 0.9.3 and the one proposed in this PR are smaller than the one for 0.9.2 and the one for 0.9.3 ; but if e.g. with networking people play a serie of races with points counted after each (which functions as a GP), a better and fairer distribution is always useful.

Alayan-stk-2:
For reference ; real car racing sports have to tweak their formula to take into account that there will sometimes be mechanical breakdowns (that's why they don't award points to all racers and reward even more aggressively the 1st, so that someone having won a couple of races more is less likely to lose the championship due to a single breakdown) ; there is a fixed number of car getting points rather than all possibilities from 1 to 19 in STK (if we did something like 10th 1, 9th 2, 8th 4 ; the total points awarded would become much too big for higher number of karts)

auriamg:
Since hiker seemed to agree to merge this, let's just try this, we can further tweak later if we find room for improvement

PR #3181, opened 2018-04-08
Update README.md

psypherium:
- Add recompilation instructions
- Add build threading instructions
- Rearranged 1 sentence with passive voice to match the rest of the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ make -j4
 ```
 STK can then be run from the build directory with `bin/supertuxkart`
 
-### Recompiling with the most recent code
+#### Keeping your build up to date
 
 To recompile the latest code without redownloading the entire source, first run the ```svn up``` command inside the 'stk-assets' directory, then run the following commands inside the 'stk-code' directory:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,19 @@ make -j4
 ```
 STK can then be run from the build directory with `bin/supertuxkart`
 
+### Recompiling
+
+To recompile the latest code without redownloading the entire source, first run the ```svn up``` command inside the 'stk-assets' directory, then run the following commands inside the 'stk-code' directory:
+
+```git pull
+make clean
```
auriamg: Make clean shouldn't be required, also there is no makefile at the root so that wouldn't work?

psypherium: Ah, yes that makes sense. I must be getting that part confused with the 0ad rebuild process. I'll rebuild again after the next commit and make sure that make clean is definitely not required.

psypherium: The commands worked perfectly without make clean.

psypherium:
- Removed unnecessary "make clean" command

deveee:
Tbh. I don't see any reason why we should add information about recompilation.

psypherium:
"To recompile the latest code without redownloading the entire source"

The word recompilation is not the right one. the 0ad build instructions say "keeping up to date", i guess I just wanted to be more fancy.

The build threading instructions are for building the code faster.

The sentence rearrangement was for readability as every other sentence was written in the passive voice.

auriamg:
Agreed that "keeping up to date" sounds better, "recompile" seems to hint "from scratch" which is not the case

auriamg:
Thanks, merged

PR #3183, opened 2018-04-11
More item AI improvements

Alayan-stk-2:
The total of line changed may look impressive, but the patch is much smaller than this : most of it is due to moving line of codes from one place to another.

Changelog : 

- New function computeSkill (called to get the itemSkill/nitroSkill of the kart instead of doing the sanity checks and boost_ai checks directly in the function using the skill value ; can be extended easily for another skill type in the future)
- Move bubblegum, cake, bowling ball, swatter and switch handling in separate functions. This makes the greatest part of the apparent change. The behavior remains identical unless noted otherwise in this changelog. Item with a short specific code remain in the main function.
- Rename ai_skill to item_skill for improved clarity
- For the level 5 AI ; remove a bomb with swatter/shield only soon before it explodes (see lines 1427-1445 and 1748-1770)
- Use bubbleshield against plungers on the face (see lines 1447-1453)
- Check the angle of the kart behind for using the gum behind (code borrowed from the bowling ball check ; see lines 1488-1499)
- Bowling ball handling significantly rewritten (check for line of fire and invulnerability before selecting target, don't fire on a kart ahead with a shield for advanced AI)
- Cake handling significantly rewritten (use numerical counters to determine target as getting everything in a few condition checks would be too complicated ; check for invulnerability before selecting target, check that the angle is decent, take the swatter of other karts into account for target selection and fire decision, don't fire on a kart straight ahead with a shield)
- Remove anvil handling (obsolete, it's not given by item boxes anymore)
- Fixes a significant bug where the AI would use an item because it meets the zipper conditions when it has something else.


src/karts/controller/skidding_ai.cpp
```diff
@@ -1575,14 +1575,20 @@ void SkiddingAI::handleCake(int item_skill)
...
 
     // Don't fire on an invulnerable kart.
-    if (m_kart_behind->isInvulnerable())
+    if (m_kart_behind != NULL)
     {
-        fire_behind -= 100.0f;
+        if (m_kart_behind->isInvulnerable())
+        {
+            fire_behind -= 100.0f;
+        }
```
hiker: Don't nest those statements, use one if and &&. Also no need to use != NULL:
```
if(m_kart_behind && m_kart_behind->isInvulnerable())
```

src/karts/controller/skidding_ai.cpp
```diff
@@ -1774,6 +1354,611 @@ void SkiddingAI::handleItems(const float dt, const Vec3 *aim_point, int last_nod
...
+    if(item_skill >= 3)
+    {
+        if (kart_behind_is_slow)
+        {
+            fire_behind -= 50.0f;
+        }
+        else 
+        {
+            fire_behind += 25.0f;
+        }
```
hiker: More compact:
```
if(kart_behind_is_slow) fire_behind -= 50.0f;
else                    fire_behind += 25.0f;
```

src/karts/controller/skidding_ai.cpp
Alayan-stk-2: Ok. I wasn't sure if this was enough to avoid a segfault.

hiker: Yes, C++ guarantees that those expressions are evaluated in order

Alayan-stk-2:
Don't merge yet ; the test_AI is still the old one, but the new results are less clear. There seems to be an issue I haven't caught, I'll need to do deeper tests to understand what may be going on.

Alayan-stk-2:
I've done further tests. Despite the randomness, 4 times 10 lap races on all 21 tracks are enough to gather a trend.

I compared the AI from this PR to the one resulting from my previous PR (by updating testAI) ; and this one lost on average 1,15 places (the previous PR had a gain of 3,8 places). The time difference is small, at 0,35% more.

I suspect a slight regression in nitro handling.

Alayan-stk-2:
The nitro code which I suspected at first hasn't even changed apart from the computeSkill thing - which detailed tests confirm to work as expected.

Further tests indicate that when the difference is small, the randomness gets bigger. Even on a 10-laps race ; a few events can change the result significantly.

I think that racing with 5 normal AIs and 4 test AIs may be not enough. I'll conduct future tests with 8 normal AIs and 7 test AIs and see if it helps to tone down the randomness.

Still, -0,64 place on average on 5 times 10 lap races (tests ran on the latest version) ; seems to indicate that there is an issue I haven't caught.

Alayan-stk-2:
After some observation of a race in green valley, I noticed the following issues : 
- Bubble Shield activated when there doesn't seem to be any danger.
- Cake thrown ahead (unsuccessfully it seems) while there was a kart very close behind.
- Swatter activated without apparent reason (shield active, no other karts close)

(And some others of course, but not related to this patch)

Alayan-stk-2:
Further debugging : 
- The cake issue has been observed again. But the log output placed inside the "if(m_controls->getFire())" of handleCake has NOT been triggered in the observed instances of random throwing. This seems to indicate that "setFire" has been called somewhere else.
- The bubble and swatter issue seems similar
- The issue probably affects other items as well

Alayan-stk-2:
I've found the root cause : there was no check that the user indeed has a zipper before using the path which activates the zipper under certain conditions.

However, this was also the case with the previous version, so this doesn't explain the poor results afaik. Maybe should I thank the bad luck in my tests ?

Alayan-stk-2:
I ran a serie of tests running my improved testing tools, with 5 series of 10 laps races. The max difference in a race is 28, up from 10. There is no big consistency on most tracks, sometimes swinging wildly from +15 to -15. There are some exception though - the situation is clearly improved in STK enterprise while I got poor results in volcano island, for example.

The overall average is a loss of one place every 35 races ; which is extremely close to zero, much smaller than the error margin.

With the randomness, it's hard to tell what the average would be on more tests.

The improvements in this patch weren't expected to be very significant, as they were mostly either about improving the code quality or fixing tiny oversights or improving things in rare edge cases.

But the zipper misfire fix is significant, and I suspect it hides the effect of a regression somewhere else.

Alayan-stk-2:
To see if there was a regression in cake handling, I replaced the new code by the old. To do my tests quickly, I use parallel folders. I put the new executable in 1, 3 and 4 ; and then, I ran the same test track file in each, with seemingly inconclusive results... Or so I thought. I realized midway that what I thought ran in my folder 4 ran in my folder 5. With the previous executable (which had the new cake handling). When I checked the results there, it got an overall +71 (the results of my first 5 tests were between -27 and +36 and averaged -0,6). Including a +20 in volcano island.

This answers my own question : the slightly subpar results I got at first seem to have more to do with sheer randomness in close races than with an hidden regression I would have missed.

That testing phase being done, I think the branch is ready for merging.

auriamg:
@hiker I am in favor of merging this, agreed?

Alayan-stk-2:
The merging conflict is easy to resolve : 1)For the first one, the test involving the zipper time left is removed because it's the nitro/zipper handling function which is tasked to manage that sort of things. Keep the PR's version. 2)For the second one, the test version of the PR has been changed, and should be kept, except that the function checking for time left should be replaced by its ticks equivalent present in master.

PR #3186, opened 2018-04-13
Greatly improve slipstreaming

### LINKED ISSUE #3149: Revamp slipstreaming
Slipstreaming implies following another kart from close behind to get a speed boost.

This is currently an unbalanced feature ; for a few reasons : 
1)The speed bonus is significant and often unexpected. This can be destabilizing and push a player to go out of track (or have similar issues)
2)There is no visual indicator of slipstreaming progress. This plays a huge part for the destabilizing aspect, as the "is slipstreamed" cue doesn't tell us if we have slipstreamed for long enough.
3)It takes a long time to get the bonus (2 seconds minimum). It is hard enough to stay close enough to the AI kart (or more exactly, inside the right angle cone) to get it. This is especially an issue if there is any kind of turning, as even if we are in fact following the same trajectory, the direction in which the kart ahead is going will make the slipstream cone change.

slipstreamangle

4)The force of the bonus and the minimum speed required are independent of the difficulty level. It would work better if this was not the case (should help with karts getting it too easily right at the start).
5)It is risky enough to go close behind another karts with bowling balls/bubble gums in play.

After improvements, slipstreaming has the potential to be a skill-selecting feature - helping a better player to have a better place despite item randomness.

It could also help to reduce slightly the advantage of being first ahead by making the pack faster.

The AI already has code to use slipstream. The hardest part is already done.

Fixes : 
- [x] Add some kind of slipstream progress bar, with clear visual cue indicating when the player will get a bonus. The progress bar should be intersected by a minimum slipstream line and go further up to a maximum time at which no additional bonus is added. The progress bar may be displayed only when the kart is slipstreaming.
- [x] Make the bonus easier to get. Require less time behind for it to activate (1 second min should be enough) and allows a larger angle. Distance may be reduced slightly however.
- [x] Change the way the angle is calculated to compensate for kart steering.
- [x] Scale the bonus force with the time behind (scaling limited by caps). The bonus should overall be slightly smaller than before, making it less destabilizing ; the ease of activation more than compensating it.
- [x] Make the speed at which slipstream activate related to the usual max speed of the difficulty level, either by a fixed offset or by a relative percentage cap [OK with zone size scaling with speed]
- [x] Scale the slipstreaming zone and bonus with speed. This would not only be logical, but make it matter more in higher difficulties and less in lower. The speed at which it activates and all other characteristics could then be uniform without issues.
- [x] Make the AI understand slipstream better (make it use the bonus willingly) [The bonus scales so it is not necessary to use the bonus as soon as possible ; except if there is a danger. Further refinements are not in the scope of this issue.]

I think I can rework the non-GUI parts.

Here is a rough outline of what the GUI could look like : 
slipstreammockup2
slipstreammockup

Alayan-stk-2:
There are so many changes here it's hard to keep track of them. Not quoting the side changes to secondary files to make some things possible : 
- Dynamically update Slipstreaming quad to reduce "swinging" when turning slightly and to make the slipstream quad more aligned against the real movement direction.
- Make the class SlipStream not be an extension of MovingTexture. This didn't make sense, as the texture is only a small part of what the class do, and movingTexture members are needed anyway to display the 2 different meshes.
- Change the starting point of the quad from back wheels to front wheels for better behavior in curves.
- Dynamically update the size of the quad according to speed : bigger quad at bigger speed
- Adds an inner quad, smaller, where the slipstreaming bonus is collected twice as fast
- Make the default quad shape wider but slightly shorter.
- Remove the utilization of a zipper bonus when collecting the bonus (unbalanced and illogical). Instead, give a direct instant acceleration associated with slipstreaming directly with instantSpeedIncrease.
- Rework the kart_properties : add new ones for some of the changes, fixes an engine_power issue which made the slipstreaming bonus useless (without the zipper bonus), put sane values, use the target kart quad length and not some weird solution with the kart's default length.
- Use more the properties in the few places where it makes sense.
- Add better visual feedback with higher dynamic speed range and more importantly a color change when the bonus is ready to be used.
- Other things I forget and which I'll add later to this list.

To get it to work, you also need to add this line in materials.xml :  ` <material name="slipstream2.png" shader="additive" ignore="Y"/> `and the following image in assets as a texture named "slipstream2.png" (same folder as "slipstream.png").

slipstream2

Fixes #3149


Mrxx99:
Would be interesting to see a video of it

Alayan-stk-2:
On the area change : https://www.youtube.com/watch?v=FR2PaiD54vw (I've done slight tweaks after but you'll mostly understand from it)
Showing it in action with another kart : https://www.youtube.com/watch?v=XGKrc2cJE8w&lc

src/karts/kart.hpp
```diff
@@ -74,6 +74,11 @@ class Kart : public AbstractKart
      *  new lap is triggered. */
     Vec3 m_xyz_front;
 
+    /** The coordinates of the 30 previous positions */
+    Vec3 m_previous_xyz[30];
+
+    float time_previous_counter;
```
auriamg: should be m_time_previous_counter

src/karts/kart.cpp
```diff
@@ -1278,9 +1299,22 @@ void Kart::update(float dt)
     {
         m_kart_animation->update(dt);
     }
+
+    time_previous_counter += dt;
+    while (time_previous_counter > (1.0f/120.0f))
+    {
+        m_previous_xyz[0] = getXYZ();
+        for (int i=29;i>0;i--)
```
auriamg: Not so important since it's only 30 elements, but this part could probably be improved a bit by using something like std::deque

auriamg:
@hiker Except a few minor comments, I would favor merging this. Would you agree?

src/karts/kart.cpp
Alayan-stk-2: According to the documentation I just read on the topic ; deque adds insertion to the front and doesn't use contiguous memory.

Neither seem relevant to an array of fixed size without any kind of insertion/deletion (these are just values update which can be done directly, doing the loop in reverse shouldn't affect performance).

Also according to cppreference.com, "deques typically have large minimal memory cost".

src/karts/kart.hpp
Alayan-stk-2: Done

src/karts/kart.cpp
auriamg: I guess I was thinking more about ring buffers, that's one common way to implement deques, though seems like the std:: one is more complex than I'd expected. Anyway, since it's only 30 items, it probably doesn't matter much

hiker:
It feels to me that it is now a bit too easy to slipstream. I had the feeling (without checking the code admittedly) that you still (as it was in the past, I know) keep the slipstream bonus for a while when you drive out of slipstream too early, but then when you go into slipstream (potentially of a different kart), you quickly get the bonus (because you kept the bonus from the previous slipstream). Would it be better if we remove this accumulating of bonus now?
Anyway, we can keep on tweaking, but I merged this in now.

Thanks a lot for your patch!
   Joerg

src/karts/kart.cpp
```diff
@@ -370,6 +376,12 @@ void Kart::reset()
     m_has_caught_nolok_bubblegum = false;
     m_is_jumping           = false;
 
+    for (int i=0;i<30;i++)
+    {
+        m_previous_xyz[i] = getXYZ();
+    }
+    time_previous_counter = 0.0f;
+
```
hiker: The 30 appears to be dependent on the FPS (there is a 0.25 mentioned below). Use the stk_config functions!

src/karts/abstract_kart.hpp
```diff
@@ -459,6 +476,9 @@ class AbstractKart : public Moveable
      *  defined even if the kart is flying. */
     virtual const Vec3& getNormal() const = 0;
     // ------------------------------------------------------------------------
+    /** Returns the position 0,25s before */
```
hiker: 0.25, not ","

src/karts/kart.cpp
```diff
@@ -3050,6 +3084,14 @@ const Vec3& Kart::getNormal() const
     return m_terrain_info->getNormal();
 }   // getNormal
 
+// ------------------------------------------------------------------------
+/** Returns the position 0,25s before */
+const Vec3& Kart::getPreviousXYZ() const
+{
+    return m_previous_xyz[29];
```
hiker: 29 again

src/karts/kart.hpp
```diff
@@ -74,6 +74,11 @@ class Kart : public AbstractKart
      *  new lap is triggered. */
     Vec3 m_xyz_front;
 
+    /** The coordinates of the 30 previous positions */
+    Vec3 m_previous_xyz[30];
```
hiker: 30 :)

src/karts/kart.cpp
```diff
@@ -1278,9 +1299,22 @@ void Kart::update(float dt)
...
+
+    time_previous_counter += dt;
+    while (time_previous_counter > (1.0f/120.0f))
+    {
+        m_previous_xyz[0] = getXYZ();
+        for (int i=29;i>0;i--)
+        {
+            m_previous_xyz[i] = m_previous_xyz[i-1];
+        }
+        time_previous_counter -= (1.0f/120.0f);
```
hiker: See above.

src/karts/kart.cpp
hiker: Another occurrence of the constant.

src/karts/kart.hpp
```diff
@@ -470,6 +478,10 @@ class Kart : public AbstractKart
     /** Returns the normal of the terrain the kart is over atm. This is
      *  defined even if the kart is flying. */
     virtual const Vec3& getNormal() const;
+    // ------------------------------------------------------------------------
+    /** Returns the position 0,25s before */
```
hiker: 0.25 - and I assume the 0.25 explains the magic constant '30': 30/120 = 0.25

hiker:
Possible that I accumulated from several karts, I wasn't looking exactly on my watch to see what was happening ;)

Alayan-stk-2:
A side effect of removing the zipper bonus, there is currently no visual effect which shows when a bonus is triggered/active. That doesn't help to know what happens with several karts.

I don't think particles behind the kart are very convincing/appropriate ; something like what MK do seems better.

data/kart_characteristics.xml
```diff
@@ -312,21 +312,27 @@
...
+                        They are upscaled when faster and downscaled when slower.
             length: How far behind a kart slipstream works
+                    (note : this helps OTHER karts)
             width: how wide slipstream works furthest away from the kart.
+            inner-factor: The proportion of the slipstreaming area with
+                          twice faster slipstream "credits" collection.
             collect-time: How many seconds of sstream give  maximum benefit
-            use-time: How long the benefit will last.
-            add-power: Additional power due to sstreaming. 1 = +100%
+            use-time: How long the benefit will last (obsolete, unused).
```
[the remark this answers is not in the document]
Alayan-stk-2: I noticed this was unused by checking, but yeah I kept it around. Some misplaced fear of breaking things.

src/karts/abstract_kart.hpp
Alayan-stk-2: Right. In french, we use "," rather than "." as a separator there (and I can tell you that calc is very unhappy if I try to use the english way). As it is in a comment, it went fine through the compilator. Feel free to change it. I won't fix it unless I need to change other things in that file though, such a minor typo don't deserve that effort.

PR #3190, opened 2018-04-14
Nitro fixes

### LINKED ISSUE #2838: Nitro sound only plays when pressing the key
The nitro sound plays only when the key for using nitro is pressed, but if you press the button only short, nitro is used for a longer time without the sound.
The sound should be played when nitro is used (and not only when the button is pressed).

### LINKED ISSUE #3033: Using Nitro while airborne uses the nitro but has no effect.
Not even the nitro trails happen.

### LINKED ISSUE #3046: Time Bomb sound lingers after Elimination
If a kart gets eliminated in Follow the Leader while having a ticking time bomb, the ticking will persist even after they are eliminated, resulting in phantom ticking.

Alayan-stk-2:
The code changes are trivial.

But those tiny changes have a big effect : 

- Makes the nitro control more responsive (stop it faster after the key stop being pressed if the min time has already elapsed)
- Fixes #2838 
- Fixes an issue related to #2838 and much worse : it wasn't only the sound which didn't apply, it was also the speed bonus 
- Fixes #3033 (it keeps not giving a speed bonus, this is only a display change to make the player aware that the nitro is consumed if he presses the key in the air)

Another small change :
- Fixes #3046


Alayan-stk-2:
I'm adding another trivial fix to this PR. I don't think it's worth having different ones

- Fixes #3046

auriamg:
@hiker I am in favor of merging, agreed?

auriamg:
Okay turns out the changes are simpler than I thought, so I just merged them

PR #3202, opened 2018-04-17
Various slipstreaming refinements

Alayan-stk-2:
In this PR : 
- A new graphical effect indicating that the kart benefits from a speed boost due to a slipstreaming bonus. It is similar to the other effects, but use a smaller less elongated mesh. It remains always centered in front of the kart and orientated towards the direction of kart movement. It matches the bonus time : the texture movement is fast when it is the highest, and it disappears when the bonus time gets to zero.
- Add an outer quad going 10% farther than the main quad, where the slipstream bonus time is lost at slow pace (1s per 1s)
- Increases the normal pace of decrease of the slipstream_timer to 3s per 1s ; to reduce the easiness of accumulating slipstreaming credits over several karts successively overtaken
- Add logic to select the best slipstreaming target rather than the one with the lowest ID (so the graphical effect is put in the most logical and natural place)
- Add logic to auto-trigger the bonus in some situation where the kart is still in a slipstream zone but has changed from kart it slipstreams from
- Move the max collect time to the characteristics
- Use a duration multiplier rather than a small and irrelevant additive duration to the main bonus time
- Refine the characteristics
- Rename the wrongly named "setIntensity"
- Remove UpdateSlipstreamPower (not needed anymore, the total bonus duration is correctly set from the first call)
- Remove the obsolete slipstreaming characteristic (also changes the "0,25" to "0.25)

For it to work well, this line must be added in the materials.xml :` <material name="slipstream_bonus.png" shader="additive" ignore="Y"/>` and the following picture must be added in the texture folder of stk-assets as slipstream_bonus.png : 

slipstream3

(For licensing, do the same as with slipstream2)


src/karts/kart.cpp
```diff
@@ -3092,12 +3092,20 @@ const Vec3& Kart::getNormal() const
 }   // getNormal
 
 // ------------------------------------------------------------------------
-/** Returns the position 0,25s before */
+/** Returns the position 0.25s before */
 const Vec3& Kart::getPreviousXYZ() const
 {
     return m_previous_xyz[29];
```
auriamg: As hiker previously mentionned, a constant would be preferable (here and everywhere related to m_previous_xyz).

PR #3204, opened 2018-04-17
Fix #1636

### LINKED ISSUE #1636: Thunderbird is constantly displayed when race is aborted early
zrzut ekranu z 2014-10-19 19 51 11

Alayan-stk-2:
Fixes #1636

Note that thunderbird will go up infinitely in the race result screen if aborted ar, but that's inconsequential as the camera is fixed.

And yes, that's a lot of code for such a minor change, but I wanted to have it finishing going up rather than disappearing, and making it work necessitated to expand significantly the capacities of the used classes.


src/modes/world.cpp
```diff
@@ -464,7 +464,7 @@ World::~World()
     {
         // No race result gui is shown, so m_race_gui is the in-race
         // gui and this must be deleted.
-        delete m_race_gui;
+        delete m_race_gui;;
```
qwertychouskie: ?

Alayan-stk-2: Good catch. Small typo introduced when I deleted my testing printf.

auriamg:
Seems like this is way overkill. Namely, it's almost certainly not necessary to copy the referee around; passing the pointer should amply suffice, and remove the need for a copy constructor

But further than that, this seems really overcomplicated. There's likely a way to just do "m_referee->hide()" (pseudo-code) somewhere and get a much simpler solution that this unecessary complication

Alayan-stk-2:
The assignation `m_race_gui       = results;` kills m_referee. A hiding or whatever wouldn't work because it would be undefined at the moment where we would like to unhide it.

Passing the pointer around means that when the first object is deleted, the second points to a memory zone which has been freed, which leads to undefined behavior and potentially segfaults. This is really, really bad.

The less bad case is that you never free the memory for the first object, so no risk of segfault ; but you get a memory leak.

There is a lot of code in this patch relative to its goal (better now though after removing some _really_ unnecessary complications ; I hope you looked at the final version) ; but this is because there were latent weaknesses in that code I had to patch.

At first, I tried to fix this issue because, on paper, it looked simple. But I got segfault upon segfault until I went to do that copy constructor and other things of the same sort. A class with pointers member and no copy constructor is asking for issues.

PR #3207, opened 2018-04-19
SuperTux in Story Mode (and other improvements)

### LINKED ISSUE #2950: Feature Request: New Game+
It would be really cool if one could re-play the Story Mode with Supertux difficulty as an option.

Alayan-stk-2:
In this PR :
- SuperTux difficulty added to all the challenges. As they're intended to be difficult challenges, their balancing can be reliably compared to high level play, with some leeway for cakes and other race events. They're harder but all perfectly doable. At the start of the Story mode, the UI behaves as if there was only three options : the challenge selection dialog, the number of cup on top of map in overworld, the max of points when a challenge is won... When SuperTux difficulty is unlocked, this changes : a 4th cup is shown, a 4th challenge option is proposed, etc. (Fixes #2950)
- The number of karts of the challenges has been revisited. Instead of having from 3 to 5 karts across the story mode, it now ranges from 4 to 10 (except special races with less). The tracks of the first GP in easy have 4 karts, and each additional level of difficulty or next GP in story mode correspond to one more AI kart.
- Some small tweaks and slight rebalancing of challenges (some laps more in oliver's class math and some others, a few small time changes). Some challenge files whose name were too inaccurate have been renamed. They need to be changed in the overworld.blend. This breaks compatibility with the previous config, however adding SuperTux difficulty to challenges by itself broke compatibility anyway. It is important that for the next version the game uses a different config folder to avoid messing things up.
- Candela City added in the 3rd GP, and some moving around, so the Story Mode has 5 races in all GP (before, at world's end had only 4 because fort magma is locked at this point ; though if replaying the challenge after finishing story mode it would have 5)
- The possibility to lock/unlock karts, difficulty, etc. based on points or other custom conditions (for example, it would be possible to check against an achievement) has been added. This contrasts with the unlock based on having completed a specific challenge (at any difficulty) which existed before, and remains here. This uses .challenge files marking the object to lock/unlock, allowing them to hook in the existing lock/unlock system. They're internally represented as challenges, but with a flag making them ignored when not relevant (points counting, etc.)
- Some functions have been renamed for more clarity. It should be unique enough to be reliably greped.
- Adds a constant used to multiply the number of points a GP cup gives to a player, as completing a 5-races challenges demands more consistency and is more time consuming. It has been set to 3, which from my tests is rather balanced.
- Awards a GP cup of an inferior level if the player has a good enough position. For example in an expert challenge GP, finishing 2nd doesn't mean getting nothing, but getting a silver cup. This change considers that a GP is time-consuming and that doing 2nd on the superior level (or 3d two level above) demonstrates at least as much skill as winning at a given difficulty. This helps to reduce frustration : the player can retry to get the better cup if he wishes so, but if it's slightly too hard for him to win at this level, he's at least not forced to redo all the races of the GP in the level below.

Assets changes needed : 
- It needs a trophy_platinum.spm model to display after having won a challenge. Locally, I've used a renamed copy of the silver cup ; this is enough for it to work ; but for release a new model will be needed (there are a lot of possibilities open, like having diamonds, gemstones, laurel, changing the shape...). Note that the cup_platinum.png in data/gui would then also have to be replaced by an icon matching the model.
- overworld.blend needs to be updated to reflect some name changes : islandtrack -> granparadiso_island ; city -> candela_city ; tuxtollway -> green_valley ; startrack -> stk_enterprise ; jungle -> cocoa_temple. The outdated names had been kept until now for a compatibility across versions which is lost.


data/challenges/unlock_bonus_kart1.challenge
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<challenge version="3">
+    <!-- This challenge can only be completed
+         by dedicated code -->
+    <unlock_list list="true"/>
```
auriamg: This syntax looks weird to me? Ho does that work? I wonder if there's a way to make that clearer

data/challenges/xr591.challenge
```diff
@@ -1,12 +1,17 @@
 <?xml version="1.0"?>
-<challenge version="2">
+<challenge version="3">
+    <unlock_list list="false"/>
```
auriamg: IMO it shouldn't be necessary to specify that everywhere, it should defaut to normal if absent from the file

src/challenges/challenge_data.cpp
```diff
@@ -79,7 +80,46 @@ ChallengeData::ChallengeData(const std::string& filename)
         return;
     }
 
+    const XMLNode* unlock_list_node = root->getNode("unlock_list");
+    if (unlock_list_node == NULL)
```
auriamg: There could be a default to avoid having to specify that everywhere

src/challenges/challenge_data.cpp
```diff
@@ -79,7 +80,46 @@ ChallengeData::ChallengeData(const std::string& filename)
...
+    {
+        throw std::runtime_error("Challenge file " + filename +
+                                 " has no <unlock_list> node!");
+    }
+
+    std::string list;
+    unlock_list_node->get("list", &list);
+
+    if(list=="true")
+        m_is_unlock_list = true;
```
auriamg: could be written in one line : m_is_unlock_list = (list=="true");

I still don't quite understand what "unlock list" means

src/challenges/challenge_status.cpp
```diff
@@ -78,14 +86,28 @@ void ChallengeStatus::setSolved(RaceManager::Difficulty d)
...
+// ------------------------------------------------------------------------
+bool ChallengeStatus::isUnlockList()
+{
+    return m_data->isUnlockList();
+} // isUnlockList
+
+// ------------------------------------------------------------------------
+bool ChallengeStatus::isGrandPrix()
+{
+    return m_data->isGrandPrix();
```
auriamg: single-line methods can be specific in the header file to help the compiler with inlining

src/challenges/story_mode_status.hpp
```diff
@@ -118,7 +124,7 @@ class StoryModeStatus
     // ------------------------------------------------------------------------
     /** Returns a challenge given the challenge id.
      */
-    const ChallengeStatus* getChallengeStatus(const std::string& id) const
+    ChallengeStatus* getChallengeStatus(const std::string& id) const
```
auriamg: Why remove const?

src/challenges/unlock_manager.cpp
```diff
@@ -257,4 +273,65 @@ void UnlockManager::findWhatWasUnlocked(int points_before, int points_now,
...
+    int hard_challenge = CHALLENGE_POINTS[RaceManager::DIFFICULTY_HARD];
+    int best_challenge = CHALLENGE_POINTS[RaceManager::DIFFICULTY_BEST];
+    std::string challenge_id;
+
+    // Unlock SuperTux difficulty
+    // We can't test if it is already unlocked because m_locked_features is undefined
+    // when this function is called
+
+
+    int supertux_floor = hard_challenge*(10+4*GP_FACTOR)+medium_challenge*(11);
```
auriamg: lots of hardcoded magic constants, avoid please

src/challenges/unlock_manager.cpp
```diff
@@ -257,4 +273,65 @@ void UnlockManager::findWhatWasUnlocked(int points_before, int points_now,
...
+        c->setSolved(RaceManager::DIFFICULTY_BEST);
+    }
+    else if( c == NULL )
+    {
+        Log::warn("Unlock Manager", "No unlock_supertux challenge defined.");
+    }
+
+    // Unlock first bonus kart
+
+    int bonus_kart_floor = easy_challenge*(20+4*GP_FACTOR); //Always unlock before fort magma
```
auriamg: more

src/challenges/unlock_manager.cpp
```diff
@@ -257,4 +273,65 @@ void UnlockManager::findWhatWasUnlocked(int points_before, int points_now,
...
+        c->setSolved(RaceManager::DIFFICULTY_BEST);
+    }
+    else if( c == NULL )
+    {
+        Log::warn("Unlock Manager", "No unlock_bonus_kart1 challenge defined.");
+    }
+
+    // Unlock second bonus kart
+
+    bonus_kart_floor = hard_challenge*(15+4*GP_FACTOR)+best_challenge*(6);
```
auriamg: more

src/challenges/unlock_manager.cpp
```diff
@@ -257,4 +273,65 @@ void UnlockManager::findWhatWasUnlocked(int points_before, int points_now,
...
+    int best_challenge = CHALLENGE_POINTS[RaceManager::DIFFICULTY_BEST];
+    std::string challenge_id;
+
+    // Unlock SuperTux difficulty
+    // We can't test if it is already unlocked because m_locked_features is undefined
+    // when this function is called
+
+
+    int supertux_floor = hard_challenge*(10+4*GP_FACTOR)+medium_challenge*(11);
+    challenge_id = "unlock_supertux";
```
auriamg: Hardcoded challenge names, not ideal

src/challenges/unlock_manager.cpp
auriamg: Challenges based on points face a major UI hurdle, the player is never told how many points they need to unlock that

src/states_screens/race_gui_overworld.cpp
```diff
@@ -122,6 +124,7 @@ RaceGUIOverworld::RaceGUIOverworld()
 
 
     // special case : when 3 players play, use available 4th space for such things
+    // TODO : determine if there are plans for multiplayer in story mode in the future
```
auriamg: None at this time

src/challenges/challenge_data.cpp
Alayan-stk-2: Default handling added, will push when other changes are ready. The node isn't necessary anymore (though I won't remove it from the challenge file unless I have another reason to update them)

Alayan-stk-2:
The version corresponding to the last commit here should make things good regarding the main purpose of this PR ; with an icon to make the player aware that there are unlockable by points and the hardcoding removed.

However, I want to partly solve #3209 too, as the same files are involved. It's partly done, but I still need to get a story_mode_status object at the right place to be able to retrieve the list of the recently completed challenges (which is currently not used at all for the purpose of displaying what are the new unlocks).

Alayan-stk-2:
Mostly done. The only remaining issue is that the unlocked feature list is cleared between the moment I set the unlocked by points feature as newly unlocked and the moment it is used. For karts linked to normal racing challenges, it already works as it is added at a different moment.

Alayan-stk-2:
I think it is ready for merging.

The platinum cup is a recoloring of the gold cup, so I release it under the same license as the original (though for 0.9.4 release the goal is to have a new model done by an artist as discussed on the IRC).

deveee:
@Alayan-stk-2 I'm not against your work, but 30 work-in-progress emails for single pull request is not normal and I have no idea if it's final version or if it will be totally different when you will finish it. I even don't have a time to read all your emails (some of them are huge). And the 80 commits is one big mess.

Again I'm glad that you work on STK improvements and also I have no idea what other people think. It's just my opinion that you should become more familiar with git, or at least work on your local repository and then split your final version into few sensible commits. And it's because currently it's simply hard to review.

Alayan-stk-2:
@deveee I know that my way of using git is not very practical. It was agreed I'd discuss with hiker to set up things properly, but we didn't really get the opportunity to do so. I try to not submit a PR until it's mostly ready to avoid CI/mail spamming ; but when I need to alter a bunch of files after this it ends with a lot of commits (as here). If you can get it from the title, it's probably easier to bulk-delete the resulting mails.

For the review, looking at it by commit is certainly much less practical than just looking at the raw final result of changed files, or selecting a range of closely related commits with shift+click.

auriamg:
@Alayan-stk-2 As discussed on IRC, it would be preferable IMO to change the "unlock_list" syntax to something a bit less hardcoded

Alayan-stk-2:
The number of points has been moved to the config files, the name of files is not present anymore in the code (instead it uses the value of an XML node) ; what part is too hardcoded yet in your opinion ?

src/states_screens/feature_unlocked.cpp
```diff
@@ -224,26 +225,41 @@ void FeatureUnlockedCutScene::findWhatWasUnlocked(RaceManager::Difficulty diffic
     {
         addUnlockedGP(grand_prix_manager->getGrandPrix(gps[i]));
     }
+    for (unsigned int i = 0; i < karts.size(); i++)
+    {
+        addUnlockedKart(const_cast<KartProperties*>(kart_properties_manager->getKart(karts[i])));
```
auriamg: Can const_cast be avoided? That's one part of c++ that's usually best left alone

A better fix would probably be to modify addUnlockedKart to receive a const KartProperties* parameter

src/states_screens/feature_unlocked.cpp
```diff
@@ -271,12 +290,17 @@ void FeatureUnlockedCutScene::addTrophy(RaceManager::Difficulty difficulty)
...
+void FeatureUnlockedCutScene::addUnlockedKart(KartProperties* unlocked_kart)
 {
-    assert(unlocked_kart != NULL);
+    if (unlocked_kart == NULL)
+    {
+        Log::error("FeatureUnlockedCutScene::addUnlockedKart", "Unlocked kart does not exist");
+        return;
+    }
+    irr::core::stringw name = unlocked_kart->getName();
+    irr::core::stringw msg = _("You unlocked ");
```
auriamg: Prefer using string formatting such as _("You unlocked %s!", name), this will be much clearer and flexible for translators

src/states_screens/main_menu_screen.cpp
```diff
@@ -321,18 +321,14 @@ void MainMenuScreen::eventCallback(Widget* widget, const std::string& name,
...
         {
             // the passed kart will not be modified, that's why I allow myself
             // to use const_cast
-            scene->addUnlockedKart(
-                                   const_cast<KartProperties*>(
-                                        kart_properties_manager->getKart("tux")
-                                                              ),
-                                    L"You unlocked <actual text would go here...>"
-                                   );
+            scene->addUnlockedKart(const_cast<KartProperties*>(
```
auriamg: const_cast, best avoided

src/states_screens/race_gui_overworld.cpp
```diff
@@ -321,18 +331,63 @@ void RaceGUIOverworld::drawTrophyPoints()
         font->draw(hardTrophiesW.c_str(), dest, time_color, false, vcenter, NULL, true /* ignore RTL */);
     }
 
-    dest = core::rect<s32>(pos.UpperLeftCorner.X - size - 5, pos.UpperLeftCorner.Y,
-                           pos.UpperLeftCorner.X - 5, pos.UpperLeftCorner.Y + size);
+    dest += core::position2di(size*2, 0);
+    if (!m_close_to_a_challenge && !PlayerManager::getCurrentPlayer()->isLocked("difficulty_best"))
+    {
+        draw2DImage(m_trophy4, dest, source, NULL,
```
auriamg: weird indentation?

auriamg:
@hiker Apart from the last few points I mentionned, I would lean towards merging, any objections?

auriamg:
After some testing, I found a little issue in this PR : after entering and then exiting story mode (the overworld), the game's font has changed, it is now bolder? Doing something with the font?

Alayan-stk-2:
I encountered a few times the issue of bold fonts in the main menu, but it hasn't been consistent at all. For example, I tried to reproduce it now in my dev branch, and the issue did not appear either by entering the overworld/quitting it ; or by entering it, finishing a challenge, and then quitting it.

I looked at the appearances of "font" in the diff of the patch and I didn't find anything which seemed related to this issue, so I'm not even entirely sure if this regression come from here.

I'd need to find steps to reproduce consistently the issue to pinpoint it.

src/states_screens/race_gui_overworld.cpp
```diff
@@ -321,18 +331,62 @@ void RaceGUIOverworld::drawTrophyPoints()
...
+        dest = core::rect<s32>(pos.UpperLeftCorner.X - 2.5f*size, pos.UpperLeftCorner.Y,
+                           pos.UpperLeftCorner.X - 1.5f*size, pos.UpperLeftCorner.Y + size);
+
+        draw2DImage(m_locked_bonus, dest, source, NULL,
+                                                  NULL, true /* alpha */);
+
+        pos.UpperLeftCorner.X -= (2*size + number_width*0.5f);
+
+        font->draw(swg.c_str(), pos, time_color, false, vcenter, NULL, true /* ignore RTL */);
+        font->disableShadow();
```
auriamg: font->disableShadow has been moved inside a "if", whereas before it was always executed. I expect that's why

Alayan-stk-2: That would also explain why I didn't see the issue when trying to reproduce it, as this if returned "true" in my test situation.

Alayan-stk-2:
Testing with a new profile on the latest version : got the issue.

Moving the font->disableShadow out of the if, recompiling and retesting : no issue anymore.

auriamg:
It has been merged, thanks

PR #3216, opened 2018-04-23
Completely revised random powerup weights system

Alayan-stk-2:
Before going into what is in the patch, let's understand the limitation of the current system.

STK uses one weight list to determine what items a kart will get. There are 5 "position class" to which all the karts are mapped (first, last, and a division in 3 of the rest), and all karts in a position class have the same probability of receiving one kind of powerup or another. Each position class has its own weights.

Be there 5 karts or 20 karts, the same weights are used, which creates balancing issues. There is a code limiting global items to limit the worst, but that's just not enough ; as trying to catchup from 20th is much harder as from 5th.

If we consider that being at a position N in a race with M karts in the race is a unique situation, there are 210 possible combinations (not even taking into account FTL, which uses the same 5 weights). 5 combinations can't reflect this properly enough. However, it's abundantly clear that a lot of those combinations are rather close, and creating specific weights for each would be overkill (and by much).

The main goal of this patch is to replace this by a much better system, flexible enough to cover adequately the range of situations which can be encountered in STK. Of course, it also contains new weights lists which by nature differs from the old one.

What this PR does : 
- Adds 5 reference points (number and value determined in config), which each corresponds to a certain number of karts in the race. Each of those reference point gets a list with 5 weights.
- Uses linear interpolation to compute the weights used with a number of kart K. For example, if the reference points are at 9 and 14 ; when there are 12 karts in the race, the weights will be equal to 0,6* the weights at 14 + 0,4* the weights at 9. This approach allows to take into account the specific situations of a race with a low, medium or high number of karts without writing an enormous number of weights lists.
- Rather than having all the karts in a position class sharing the same probabilities, those class are repurposed as reference points. This patch uses linear interpolation to vary the weights used by each kart. For example, say we have a kart at rank N, a reference point at N-2 and another at N+1. The kart will randomly choose to use either the weights of N+1 or N-2, with probability two-thirds with N+1 and one-third with N-2. This smooths things a lot : gaining or losing a place will have gradual effects on items received ; rather than plateaus followed by sudden jumps ; while keeping the list in config at a reasonable size.
- Adds specific weights for FTL (only one reference point, but the leader gets special treatment).
- Adds new weight lists to take advantage of the new possibilities offered.

I expect that most points will be non-controversial, as they allow to improve the gameplay by offering a better balance.

However, the implicit changes to the probabilities to receive an item may beg some questions as to my choices, so I'll discuss them more in-depth.

First of all, the old distribution looked like this : 

item_distribution_old

There was also a code applied to make the parachute and the switch more rare with more karts.

In the config file, it is rather confusing as the sum of the weights differ for each position class, making direct comparison very hard.

Here are what the new distributions look like : 

item_distribution_20karts
item_distribution_14
item_distribution_9karts
item_distribution_5karts
item_distribution_1kart

To fully understand the changes, we need to separate the items into a few general type : 1)the positive items which help one given kart ; 2)the negative item which hinders one (or a few) kart not far ; 3)the global item which affects kart globally.

Giving more item of the 1st sort at rank N tends to make the kart at rank N faster. Giving more item of the 2nd sort tends to make karts around rank N slower.

Comments : 
- There is not anymore code to make global items more rare. The list by themselves ensure that the switch, the parachute and also the basket ball occurs at roughly the same overall rate (1 item per 4 boxes*number of karts in the race was my rough target, it varies slightly around it), except when there are very few karts. The averaging of weights between the reference points helps a lot here.
- The reference points at 1 kart may look weird, but the values are used in the averages when there are 2, 3 or 4 karts, and it has been designed with this in mind.
- Before, cakes were more common in worse ranks. The graphs don't show this, but the old distribution also have a significant number of triple cakes among its cakes. The thought was that giving more powerful weapons would help them. However, badly ranked karts throws those at each other. It hinders them overall. In the new design, more cakes are given in better positions, and this is done to hinder the karts in those positions compared to those in worse positions ; except for the first (and a bit less those just behind), as then the cake would mostly be used backwards against overtaking.
- The bubble gum is a very good defensive item when used as a shield. It is useful at any position. The old weights date of a time when there was no shield, and so make it too common for the first and too rare for the others. The new weights balance this out.
- Swatters are used to slow down karts close around, or as a weak defensive item (much less useful than the gum in that role, but still sometimes good). Giving more of them in better positions is, again, not really a boon.
- The differences in the quality/usefulness of the item received are more stark between the better positions and the worse positions when there are more karts in a race, because it's harder to get back from 20th than from 10th or 5th.
- The FTL weights have no cake as those are a big part of why FTL is horribly random.
- Much less plungers in bad positions. Those can be quite useful, but they are situational and often fail.

Overall, I don't pretend that the new weights are perfect. Obviously, I couldn't playtest them in the aforementioned 210 situations : they are mostly based on my knowledge of the game and its balance but could certainly benefit from tweaks following testing feedback. Still, I'm confident they are significantly better than what we have now and that the new system with averaging is more efficient and powerful.

### DIFF
--- a/data/powerup.xml
+++ b/data/powerup.xml
@@ -74,38 +74,123 @@
         force-updown="35"     force-to-target="15"
         max-distance="25"                               />
 
+    <!-- This defines the number of karts for which there is a
+       defined weight list. The value between two nodes will be
+       interpolated as a linear average of the value at each node,
+       except under the minimum node and above the maximum node
+       (in which case it stays identical to the minimum/maximum node)
+
+       The num of karts associated with each reference point should be
+       sorted in ascending order to guarantee correct behavior.
+
+       This doesn't apply to follow the leader, battle & tutorial weights
+       If the max karts number is increased, think to add values for
+       higher kart numbers. Otherwhise, it will default to value for 20. -->
+  <reference_points ref1="1" ref2="5" ref3="9" ref4="14" ref5="20" />
+
   <!-- Distribution of the items depending on position of the 
-       kart (first kart, top 33%, middle 33%, bottom 33% (but
-       not last), last kart. i
+       kart. There are defined value for first kart, last kart, and
+       at three equally spaced points between those.
+       For example, with 5 karts those points will match positon 2, 3, 4.
+       With 10 karts, they will correspond to 3,25 ; 5,5 ; 7,75.
+       It is not an issue for it to be a float, as the weights for positions
+       are linearly averaged.
        The order of items must correspond to powerup_manager.hpp.
-       The first line (w=...) corresponds to the weights of 
-       getting a single item, the second line (w-multi) to 
-       getting three identical items instead of just a single one.
-       E.g. as the first kart, you have 3/8 of a chance to get
-       a single bubble gum, 1/8 to get a bowling ball or switch,
-       and 3/8 to get a triple bubble gum.
-       'Global' items which affect all karts (switch) should
+
+       The first line (w1=...) corresponds to the weights of 
+       getting a specific item, the second line (w-multi1) to 
+       the weight at which it will yeld a triple item
+       rather than a single one
+
+       The probability to get an item is its weight divided by
+       sum of weights of all items (single AND multi). It is recommended 
+       to keep that sum equal to 200 to easily keep track of probabilities.
+
+       'Global' items which affect all karts (switch, parachute) should
        be quite rare, since otherwise the item might be used
        too often (compared with many items which will only
-       affect a karts or two).                                -->
+       affect a karts or two).
+
+       The distribution should give more similar item to different ranks
+       when there are few karts, but have more important differences when
+       there are more karts.                               -->
   <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
-  <first  w="25       5    15     5      10       10      10     0     0     0"
-    w-multi=" 0       0     5     0       0        0       0     0     0     0"    />
-  <top33  w="30      30    30    30      30       10      10    10    10     0"
-    w-multi=" 0      10    10     0      10        0       0     0     0     0"    />
-  <mid33  w="30      30    30    30      30       10      20    20    20     0"
-    w-multi=" 0      20    20    20      20        0       0     0     0     0"    />
-  <end33  w=" 0      30    30    30      30       10      20    30    30     0"
-    w-multi=" 0      30    30    30      30        0       0     0     0     0"    />
-  <last   w=" 0      30    10    90      60        0      10    60    60     0"
-    w-multi=" 0      30    30   100      60        0       0     0     0     0"    />
+  <first  w1="28       0    60    20      45       15      32     0     0     0"
+    w-multi1=" 0       0     0     0       0        0       0     0     0     0"
+          w2="26      16    52    15      46       12      33     0     0     0"
+    w-multi2=" 0       0     0     0       0        0       0     0     0     0"
+          w3="24      12    58    10      54        8      34     0     0     0"
+    w-multi3=" 0       0     0     0       0        0       0     0     0     0"
+          w4="22       8    64     5      60        6      35     0     0     0"
+    w-multi4=" 0       0     0     0       0        0       0     0     0     0"
+          w5="20       0    74     0      66        4      36     0     0     0"
+    w-multi5=" 0       0     0     0       0        0       0     0     0     0" 
+          wf="35       0    25    35      25       15      25     0     0     0"
+    w-multif="20       0     0    20       0        0       0     0     0     0"  />
+
+  <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
+  <top33  w1="30       0    59    26      40       14      27     0     0     0"
+    w-multi1=" 0       0     4     0       0        0       0     0     0     0"
+          w2="30      27    48    25      28       10      27     0     0     0"
+    w-multi2=" 0       0     5     0       0        0       0     0     0     0"
+          w3="29      30    45    24      32        7      27     0     0     0"
+    w-multi3=" 0       0     6     0       0        0       0     0     0     0"
+          w4="28      31    48    22      34        4      27     0     0     0"
+    w-multi4=" 0       0     6     0       0        0       0     0     0     0"
+          w5="27      32    48    20      37        3      27     0     0     0"
+    w-multi5=" 0       0     6     0       0        0       0     0     0     0"
+          wf="25       0    60    25      58        2      30     0     0     0"
+    w-multif=" 0       0     0     0       0        0       0     0     0     0"    />
+
+  <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
+  <mid33  w1="30       0    62    27      36       13      27     0     0     0"
+    w-multi1=" 0       0     5     0       0        0       0     0     0     0"
+          w2="30      27    45    27      27        9      27     3     0     0"
+    w-multi2=" 0       0     5     0       0        0       0     0     0     0"
+          w3="30      26    41    28      26        6      26    10     0     0"
+    w-multi3=" 0       0     7     0       0        0       0     0     0     0"
+          w4="30      25    42    29      29        3      24    10     0     0"
+    w-multi4=" 0       0     8     0       0        0       0     0     0     0"
+          w5="30      24    40    30      28        2      21    10     0     0"
+    w-multi5=" 0       0    10     0       5        0       0     0     0     0"
+          wf="35       0    55    35      25        3      25     0     0     0"
+    w-multif=" 0       0    10     0      12        0       0     0     0     0"    />
+
+  <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
+  <end33  w1="31       0    56    36      34       12      25     0     0     0"
+    w-multi1=" 0       0     6     0       0        0       0     0     0     0"
+          w2="32      24    28    38      22        7      20    16     6     0"
+    w-multi2=" 0       0     7     0       0        0       0     0     0     0"
+          w3="33      23    26    45      14        5      16    12     8     0"
+    w-multi3=" 0       0    12     0       6        0       0     0     0     0"
+          w4="27      21    23    44      12        3      14     8     6     0"
+    w-multi4=" 8       0    16     8      10        0       0     0     0     0"
+          w5="25      18    20    50      10        2      10     6     3     0"
+    w-multi5="10       0    20    10      16        0       0     0     0     0"
+          wf="25       0    40    45      15        5      15    10     5     0"
+    w-multif="10       0    15    15       0        0       0     0     0     0"    />
+
+  <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
+  <last   w1="34       0    36    55      30       10      17     0     0     0"
+    w-multi1=" 0       0    18     0       0        0       0     0     0     0"
+          w2="28      21     9    45       0        6       0    10    18     0"
+    w-multi2=" 8       0    16    35       4        0       0     0     0     0"
+          w3="20      16     7    37       0        3       0     4    15     0"
+    w-multi3="18       0    18    58       4        0       0     0     0     0"
+          w4="18      14     3    35       0        0       0     0    10     0"
+    w-multi4="24       0    25    65       6        0       0     0     0     0"
+          w5="15      12     0    25       0        0       0     0     7     0"
+    w-multi5="30       0    31    80       0        0       0     0     0     0"
+          wf="20       0    15    25       0        0       0     0    15     0"
+    w-multif="20       0    25    80       0        0       0     0     0     0"    />
+
+  <!--      bubble  cake  bowl  zipper  plunger  switch swattr rubber para  anvil -->
   <battle w="10      30    60     0       0       10      30     0     0     0"
     w-multi=" 0       0     5     0       0        0       0     0     0     0"    />
   <soccer w=" 0      30    60     0       0       10      30     0     0     0"
     w-multi=" 0       0     5     0       0        0       0     0     0     0"    />
   <tuto   w=" 0       0     0     0       0        0       0     0     0     0"
-    w-multi=" 0       0   100     0       0        0       0     0     0     0"    />
+    w-multi=" 0       0   100     0       0        0       0     0     0     0" />
 
     
 </powerup>
-
--- a/data/powerup.xml
+++ b/data/powerup.xml
@@ -88,10 +88,11 @@
        higher kart numbers. Otherwhise, it will default to value for 20. -->
   <reference_points ref1="1" ref2="5" ref3="9" ref4="14" ref5="20" />
 
-  <!-- Distribution of the items depending on position of the 
-       kart. There are defined value for first kart, last kart, and
+  <!-- This defines the probabilities to get each type of item depending on
+       the position of the kart.
+       There are defined value for first kart, last kart, and
        at three equally spaced points between those.
-       For example, with 5 karts those points will match positon 2, 3, 4.
+       For example, with 5 karts those points will match positions 2, 3, 4.
        With 10 karts, they will correspond to 3,25 ; 5,5 ; 7,75.
        It is not an issue for it to be a float, as the weights for positions
        are linearly averaged.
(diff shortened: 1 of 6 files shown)


Arthur-D:
I'm not going to comment on the code itself since I don't have anywhere near that expertise, nor will I comment on how well this works since I haven't tested it yet, but I wanted to say that the following quotes reflect well what I've seen over the years:

> Before, cakes were more common in worse ranks. The graphs don't show this, but the old distribution also have a significant number of triple cakes among its cakes. The thought was that giving more powerful weapons would help them. However, badly ranked karts throws those at each other. It hinders them overall. In the new design, more cakes are given in better positions, and this is done to hinder the karts in those positions compared to those in worse positions ; except for the first (and a bit less those just behind), as then the cake would mostly be used backwards against overtaking.

> The bubble gum is a very good defensive item when used as a shield. It is useful at any position. The old weights date of a time when there was no shield, and so make it too common for the first and too rare for the others. The new weights balance this out.

> The FTL weights have no cake as those are a big part of why FTL is horribly random.

> Much less plungers in bad positions. Those can be quite useful, but they are situational and often fail.

These are things that could be balanced better than currently even with the current single list of probabilities, and do not in themselves necessarily warrant a larger change such as this. However, given that the current list were made for having a low number of karts, such as 4-6 karts, and having more karts are no longer as taxing performance-wise revisiting how we balance items is prudent even just from a single-player perspective.

More importantly, since we will hopefully soon have the ability to play networked multiplayer with up to 12 karts or even more, and this being human players, having the item weights be as fair as reasonably possible will be something people expect. The aforementioned quotes were things I had noticed to varying degrees myself, but it being a single-player experience made them less important to get fixed. All in all, maybe the approach taken in this pull request is a bit more granular than strictly necessary (I honestly have no idea), but there's absolutely merit in shaking the item weights up and improve things.

Alayan-stk-2:
Another visualisation : 

item_distribution_20_old
item_distribution_20_new

src/items/powerup_manager.cpp
```diff
@@ -205,142 +208,343 @@ void PowerupManager::LoadPowerup(PowerupType type, const XMLNode &node)
...
+            {
+                // With default config it won't loop more than five time
+                // This is mainly to let full config editing power
+                for(int i=1;i<=50;i++)
+                {
+                    ref = "ref" + StringUtils::toString(i);
+                    std::string ref_pos;
+                    if (!node2->get(ref, &ref_pos))
+                        break;
+                    unsigned int ref_value = atoi(ref_pos.c_str());
```
auriamg: atoi should be avoided as it has very peculiar error handling. Take a look at StingUtils::parseString<int> (will require a bit more error handling but that's a good thing)

src/items/powerup_manager.cpp
```diff
@@ -205,142 +208,343 @@ void PowerupManager::LoadPowerup(PowerupType type, const XMLNode &node)
...
+            if(race_manager->isFollowMode())
+            {
+                w_inf = w_sup = "wf";
+            }
+            
+            else
+            {
+                // With default config it won't loop more than five time
+                // This is mainly to let full config editing power
+                for(int i=1;i<=50;i++)
```
auriamg: Why is 50 hardcoded? Using a constant, at bare minimum, would be better

src/items/powerup_manager.cpp
```diff
@@ -205,142 +208,343 @@ void PowerupManager::LoadPowerup(PowerupType type, const XMLNode &node)
...
+                                 unsigned int num_karts,
                                  const std::string &class_name,
                                  PositionClass position_class)
 {
-    const XMLNode *node = root.getNode(class_name);
-    std::string s(""), s_multi("");
-    if(node) node->get("w",       &s      );
-    if(node) node->get("w-multi", &s_multi);
+    const XMLNode *node = root.getNode(class_name), *node2 = root.getNode("reference_points");
+    std::string s_inf(""), s_inf_multi(""), s_sup(""), s_sup_multi(""), ref;
```
auriamg: All those variables don't have very clear names. Choosing better names, or adding documentation comments, could help clarify that a lot

src/items/powerup_manager.cpp
```diff
@@ -205,142 +208,343 @@ void PowerupManager::LoadPowerup(PowerupType type, const XMLNode &node)
...
+    m_powerups_for_reference_pos[pos_class].clear();
+
+    // This loop actually goes over the powerups twice: first the single item version,
+    // then the multi-item version. The assignment to type takes care of this
+    for(unsigned int i= POWERUP_FIRST; i<=2*POWERUP_LAST; i++)
+    {
+        PowerupType type =
+            (PowerupType) ((i<=POWERUP_LAST) ? i
+                                             : i+POWERUP_FIRST);
+        unsigned int w =m_weights[pos_class][i-POWERUP_FIRST];
```
auriamg: Those lines are pretty weird. i + POWERUP_FIRST followed by i - POWERUP_FIRST, huh? Maybe this code is correct, but it's not obvious or easy to follow

src/items/powerup_manager.hpp
```diff
@@ -126,30 +129,35 @@ class PowerupManager : public NoCopy
...
+    void          updatePowerupClass(PowerupManager::PositionClass pos_class);
     PositionClass convertPositionToClass(unsigned int num_karts,
+                                         unsigned int position, bool class_sup = false);
+    unsigned int           convertPositionToClassWeight(unsigned int num_karts,
                                          unsigned int position);
 public:
                   PowerupManager  ();
                  ~PowerupManager  ();
-    void          loadAllPowerups ();
+    void          loadAllPowerups (unsigned int num_karts=10);
```
auriamg: why 10?

data/powerup.xml
```diff
@@ -74,38 +74,123 @@
...
+          w2="26      16    52    15      46       12      33     0     0     0"
+    w-multi2=" 0       0     0     0       0        0       0     0     0     0"
+          w3="24      12    58    10      54        8      34     0     0     0"
+    w-multi3=" 0       0     0     0       0        0       0     0     0     0"
+          w4="22       8    64     5      60        6      35     0     0     0"
+    w-multi4=" 0       0     0     0       0        0       0     0     0     0"
+          w5="20       0    74     0      66        4      36     0     0     0"
+    w-multi5=" 0       0     0     0       0        0       0     0     0     0" 
+          wf="35       0    25    35      25       15      25     0     0     0"
+    w-multif="20       0     0    20       0        0       0     0     0     0"  />
```
auriamg: What is "multif" ?

src/items/powerup_manager.cpp
```diff
@@ -205,142 +208,343 @@ void PowerupManager::LoadPowerup(PowerupType type, const XMLNode &node)
...
 void PowerupManager::loadWeights(const XMLNode &root,
+                                 unsigned int num_karts,
                                  const std::string &class_name,
                                  PositionClass position_class)
 {
-    const XMLNode *node = root.getNode(class_name);
-    std::string s(""), s_multi("");
-    if(node) node->get("w",       &s      );
-    if(node) node->get("w-multi", &s_multi);
+    const XMLNode *node = root.getNode(class_name), *node2 = root.getNode("reference_points");
```
auriamg: Instead of node and node2, please give them meaningful names

src/items/powerup_manager.hpp
Alayan-stk-2: I originally had this as a default fallback value, though the call of the function in powerup_manager.cpp should always specify the number of karts, so this may be not needed.

src/items/powerup_manager.cpp
Alayan-stk-2: This is the same thing as the original loop, you can look it up on lines 286-296 of the file in master : https://github.com/supertuxkart/stk-code/blob/master/src/items/powerup_manager.cpp

The i+POWERUP_FIRST is in a ternary operator who is used to have the correct powerup type associated when the loop goes over the second part of the values to know how many triple-items there are.

The i-POWERUP_FIRST is used to convert the value from i to a correct index value, as POWERUP_FIRST, iirc, will be something like 1, while the first element of an array is at [0].

src/items/powerup_manager.cpp
Alayan-stk-2: I've put a while loop instead.

src/items/powerup_manager.cpp
Alayan-stk-2: The original code also used atoi. But it seems to work with parseString and there are now error warnings if it goes wrong, so it's better I suppose.

qwertychouskie:
> Another visualisation :

The second one looks like modern art :P

src/items/powerup_manager.hpp
auriamg: If the arbitrary default value is never need, then indeed it's preferable to remove it

data/powerup.xml
auriamg: It would be good to clarify that, at least with a comment, of by renaming it ("-ftl" maybe?)

Alayan-stk-2: Comment added. I kept it as wf ; as the clarity difference is minimal with the comment and as it would mean having to add spaces to realign over 50 lines.

auriamg:
@hiker I would lean towards merging, any objections?

Benau:
...i am not sure if after you merging it will make item synchronizing in network too difficult

Alayan-stk-2:
It shouldn't change anything for synchronization. The weights are determined at the beginning of a race, before the starting countdown.

Then : 
- The server must send the item to the client, and as it can change depending on the position of the kart, it's not really possible to predict it. However, this is already the case for the transitions between the existing position classes, so the problem is the same.
- As for kart's elimination, both before and after this PR, the weights are unchanged when a kart is eliminated (this is also true in FTL). For example, if there are 5 karts at the start, and one is eliminated, the 4th will keep receiving items as 4th in a 5-karts race and note 4th in a 4-karts race. I think this is a desirable behavior in FTL. In network races, it may not be ideal if the elimination happens right at the start, but in most cases, when it happens mid-race, that's OK I think (especially if/when it's ranked). In any case, a kart's elimination has no special impact on powerup computation/synchronization.

auriamg:
I get a crash here :
```
 	supertuxkart.exe!MaterialManager::getMaterial(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & fname, bool is_full_path, bool make_permanent, bool complain_if_not_found, bool strip_path, bool install) Line 394	C++
>	supertuxkart.exe!PowerupManager::LoadPowerup(PowerupManager::PowerupType type, const XMLNode & node) Line 169	C++
 	supertuxkart.exe!PowerupManager::loadAllPowerups(unsigned int num_karts) Line 127	C++
 	supertuxkart.exe!PowerupManager::updateWeightsForRace(unsigned int num_karts) Line 421	C++
 	supertuxkart.exe!World::init() Line 229	C++
 	supertuxkart.exe!WorldWithRank::init() Line 46	C++
 	supertuxkart.exe!LinearWorld::init() Line 66	C++
 	supertuxkart.exe!RaceManager::startNextRace() Line 550	C++
 	supertuxkart.exe!RaceManager::startNew(bool from_overworld) Line 460	C++
 	supertuxkart.exe!RaceManager::startSingleRace(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & track_ident, const int num_laps, bool from_overworld) Line 940	C++
 	supertuxkart.exe!TrackInfoScreen::onEnterPressedInternal() Line 386	C++
 	supertuxkart.exe!TrackInfoScreen::eventCallback(GUIEngine::Widget * widget, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & name, const int playerID) Line 397	C++
```

The problematic material is "zipper_collect.png", is there a file missing?

src/main.cpp
```diff
@@ -1737,7 +1737,6 @@ int main(int argc, char *argv[] )
             material_manager->addSharedMaterial(materials_file);
         }
         Referee::init();
-        powerup_manager->loadAllPowerups();
```
auriamg: I don't think you can remove that. This loads all the models and materials, and this needs to happen at this point in the init sequence. Maybe the parts that loads models and textures, and the parts that load weights, could be separated

Alayan-stk-2: Okay, that makes sense. I'll separate the weights loading from the model/texture loading.

Alayan-stk-2: Should be good now. I've retested locally and I could launch and race without issue with those latest changes.

The good thing with this error of mine is that now the splitting allows to not reload the models on each new race.

auriamg:
Still not working. Tried with 8 AI karts

```
        //Do a linear average of the inf and sup values
        //Use float calculations to reduce the rounding errors
        float ref_diff = (float) (sup_num - inf_num);
        float sup_diff = (float) (sup_num - num_karts);
        float inf_diff = (float) (num_karts - inf_num);
        float weight = (sup_diff/ref_diff)*weight_inf
                      +(inf_diff/ref_diff)*weight_sup;
        int rounded_weight = (int) weight;

        m_weights[position_class].push_back(rounded_weight);
```

```
>	supertuxkart.exe!PowerupManager::loadWeights(const XMLNode & root, unsigned int num_karts, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & class_name, PowerupManager::PositionClass position_class) Line 410	C++
 	supertuxkart.exe!PowerupManager::loadAllPowerups(unsigned int num_karts) Line 147	C++
 	supertuxkart.exe!PowerupManager::updateWeightsForRace(unsigned int num_karts) Line 432	C++
 	supertuxkart.exe!World::init() Line 229	C++
 	supertuxkart.exe!WorldWithRank::init() Line 46	C++
 	supertuxkart.exe!LinearWorld::init() Line 66	C++
 	supertuxkart.exe!RaceManager::startNextRace() Line 549	C++
 	supertuxkart.exe!RaceManager::startNew(bool from_overworld) Line 459	C++
 	supertuxkart.exe!RaceManager::startSingleRace(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & track_ident, const int num_laps, bool from_overworld) Line 930	C++

```

ref_diff is 0, resulting in a division by 0, weight is therefore NaN.sup_num = 9, inf_num= 9

hiker:
> Picking one of them with a certain probability is mathematically equivalent to averaging them with that probability as a weight. The reasoning behind this implementation choice was that we'd only have to keep the weights for the 5 classes rather than having full weights lists of 200 items for each of up to 20 karts. It also reduces significantly rounding errors in the averaging.

No, it is not. It is (very likely) equivalent from a statistical point of view, but not for one specific case: with your method you can only get a distribution according to N+1 or N-2, but not (exactly) the interpolated distribution.

I also don't understand the argument about the rounding - each number can easily be rounded to be between 0 and 100 (i.e. as %). And I also don't think you need to store the full list of 200 items for up to 20 karts - for the current race you create one list for each position, with (say) 100 entries: each entry indicating a powerup type. You then select the rank and pick a random number between 0 and 100, and have the item that is to be used. 

> I don't know how such an obvious error got until here - seems that I indeed didn't test in debug mode. Should be good now.
Would it be possible to add unit tests (a very trivial, but working 'framework' is in place ... though all we do is using asserts ;) )

Alayan-stk-2:
> No, it is not. It is (very likely) equivalent from a statistical point of view, but not for one specific case: with your method you can only get a distribution according to N+1 or N-2, but not (exactly) the interpolated distribution.

As said on the IRC, this doesn't matter since each call to getRandomPowerup will first look at the distribution to use, then pick a random item from it (using two different parts of an unique random integer). If you look at getRandomPowerup as a black box outputting an item to pick, you wouldn't be able to tell apart this method from a (non-rounded) interpolated distribution on the basis of the item you'd get.

> I also don't understand the argument about the rounding - each number can easily be rounded to be between 0 and 100 (i.e. as %). And I also don't think you need to store the full list of 200 items for up to 20 karts - for the current race you create one list for each position, with (say) 100 entries: each entry indicating a powerup type. You then select the rank and pick a random number between 0 and 100, and have the item that is to be used.

For the more common items ; 1% more or less is not a big deal ; but for the global items (parachute, switch, basket ball) it is. Rounding 1,5% to 1% or to 2% is bad ; having 2% 2% 2% 1% 1% 1% instead of 2% 1,8%, 1,6% 1,4% 1,2% 1% is poor, especially as it is not harder/more CPU intensive to have the second behavior.

> Would it be possible to add unit tests (a very trivial, but working 'framework' is in place ... though all we do is using asserts ;) )

Not sure about what you envision here. The error didn't happen when using unexpected values from another function, but inside a function not handling correctly possible expected values.

hiker:
I have the feeling that Alayan and I actually mostly agree, but we have some misunderstandings somehow. Since I am currently working on the networking part of item collection (which will likely cause conflicts with this patch), I'll finish this work first, and then merge this in and see about the few points we disagree about, but I need some time for that :) But I've assigned this pull request to me now.

Alayan-stk-2:
Github is still unhappy and see conflicts, but I basically took the version from master and merged in the lone change to race_manager.hpp this PR contained from the start ; so the version in the PR can replace the one in master.

qwertychouskie:
Not sure if it makes sense to give a bunch of plungers to the first place kart, they really have little to no use for the first place kart.

Alayan-stk-2:
That's the point.

PR #3224, opened 2018-04-30
Update custom_video_settings.stkgui

psypherium:
Change label titles to conform to title case typographical convention.

### DIFF
--- a/data/gui/custom_video_settings.stkgui
+++ b/data/gui/custom_video_settings.stkgui
@@ -18,7 +18,7 @@
         <div layout="horizontal-row" width="100%" proportion="1">
             <checkbox id="dynamiclight"/>
             <spacer width="10" height="10"/>
-            <label text="Advanced pipeline (lights, etc.)" I18N="Video settings"/>
+            <label text="Advanced Pipeline (lights, etc.)" I18N="Video settings"/>
         </div>
 
         <spacer height="4" width="10" />
@@ -49,7 +49,7 @@
             <div layout="horizontal-row" proportion="1" height="fit">
                 <checkbox id="lightshaft"/>
                 <spacer width="10" height="10"/>
-                <label text="Light shaft (God rays)" I18N="Video settings"/>
+                <label text="Light Shaft (God Rays)" I18N="Video settings"/>
             </div>
         </div>
 
@@ -69,7 +69,7 @@
             <div layout="horizontal-row" proportion="1" height="fit">
                 <checkbox id="dof"/>
                 <spacer width="10" height="10"/>
-                <label text="Depth of field" I18N="Video settings"/>
+                <label text="Depth of Field" I18N="Video settings"/>
             </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


psypherium:
Also, a minor issue is that the graphics settings box doesn't scale to higher resolutions, leaving white space on the right side. however most users will not encounter this problem

auriamg:
You mention you want to adhere to the typrographical conventions of titles, but the labels you changed aren't titles?

Though we aren't uniform, some labels are "Camel Case" and some labels are "Pascal case", I guess we should at least try to be uniform. I'd go for "Pascal case" for non-titles though. This will be very low priority though

psypherium:
Most people think of it the same way you do. https://ux.stackexchange.com/questions/28297/should-we-use-title-case-or-sentence-case-for-headlines-and-buttons?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

It probably should be written in sentence case. It's semantically interesting, it depends on if you consider a button to have a name/title or is it simply a description of a feature. Thinking about it more I agree with you.

I'll do the opposite to what I did and remove the capitalisations instead of adding more. Consistency will make it look much better.

auriamg:
Yes, I know my use of "Camel Case" and "Pascal case" were not quite the usual meaning, but you figured what I meant anyway ;)

psypherium:
This looks good now.

How about the GP's next? Two are in title-case and the other two are sentence case. Two of them are "sentences", but the sentences are the titles of the GP's so they should be title case? Like "Off the Beaten Track" instead of "Off the beaten track" if that sounds good I can try to hunt down those bits of code for you, you guys have much more important stuff to do.

data/gui/custom_video_settings.stkgui
```diff
@@ -81,7 +81,7 @@
             <div layout="horizontal-row" proportion="1" height="fit">
                 <checkbox id="glow"/>
                 <spacer width="10" height="10"/>
-                <label text="Glow (outlines)" I18N="Video settings"/>
+                <label text="Glow (Outlines)" I18N="Video settings"/>
```
auriamg: I don't believe words inside parentheses require a capital letter? Got a reference for this? I found this ( http://englishplus.com/grammar/00000123.htm ) which seems to indicate the lower case was correct

auriamg:
As for GPs, I consider that excessively minor TBH, those could be considered titles since they are the GP's title

hiker:
I would prefer title case for GP titles.

data/gui/custom_video_settings.stkgui
psypherium: It depends whether or not they are separate sentences. Is it "Glow. Outlines." or "Glow outlines"? I don't know to be honest. I was just capitalising them as if they were separate.

psypherium:
https://github.com/search?utf8=%E2%9C%93&q=repo%3Asupertuxkart%2Fstk-code+%22off+the+beaten+track%22&type=Code I've found them all, now to figure out how to batch edit them or something.

Alayan-stk-2:
No, only in the original file. The translation files will be regenerated in the preparation-to-release phase.

data/gui/custom_video_settings.stkgui
auriamg: The intent, I guess, is "Glowing, sometime also referred to as outlines" or something like that

auriamg:
The current state looks good, I've merged it

PR #3232, opened 2018-05-02
Increase minimap size and reduce its aliasing

Alayan-stk-2:
This PR : 
- Increases the minimap size. Why ? The minimap main uses are to a) get a quick overview of the track's shape ; b) see where the other karts are on the track. Both involve looking at it in a quick glance, and the small size of the current minimap make it hard, because one need to analyze smaller features and because the kart's icons overlap (especially in long a track). The bigger size doesn't hinder the view in racing (btw, in MK the minimap is significantly bigger than in STK). The overworld minimap is already bigger, it's kept as-is.
- Reduces minimap aliasing. The aliasing issue exists in master, but as the minimap is small, it's less noticed than when the minimap is made bigger. The method used doesn't deliver perfect antialiasing, but it's simple and good enough.


Mrxx99:
You should attach comparison pictures ;)

Alayan-stk-2:
Anti-aliasing : https://i.imgur.com/kATRvCg.png (x1 = old ; x2 = new ; it varies depending on minimap size, but I tested with the increased size on various resolutions and it was fine)

Old size : https://i.imgur.com/idEeQ1O.jpg
New size : https://i.imgur.com/tCuJb2g.jpg

auriamg:
I'd like other opinions too, but personally I Find the new proposed side a little distracting. I'd go for something a bit in-between

Arthur-D:
I think the proposed size is reasonable, because it's meant to help give information, and as such it's going to inherently be a bit attention grabbing. I won't protest having it a bit smaller than the proposal as a compromise, but at least the current one is on the small side and would be nice to have larger.

Alayan-stk-2:
In race, your attention is focused on your kart ; so even if on a still-image this may seem to have a distracting size, this is not the case in-race. The example picture deliberately shows it in 4:3 resolution as it's where the horizontal space is the more limited.

See those minimaps in MK : https://i.imgur.com/11p0y9i.jpg ; https://i.imgur.com/WCn8VdE.jpg. They take an even bigger portion of the screen space. You can bet that if it was too distracting, they would have not done it that way.

Benau:
I have a question:

what happen if in the future there is someone open another PR saying the minimap is too big and propose a smaller one

Alayan-stk-2:
The question would be : is that someone right ? Which leads to the key question "what is the best size ?". Unless we'd want to make it configurable (certainly doable, but probably overkill), it would be the size at which it is the most useful for most users, with a small-to-null distraction/hindering when not looked at and not too much space taken from the toplist (this pushes the best size down), and where it's the easiest to get informations on track shape and race situation in a glance (this pushes the best size up).

While the proposed size is not the pixel-perfect-ideal-size ; I think it's within a few % of the best value.

hiker:
I like the idea of a bigger map if the quality of the map is good enough. It looks a bit odd because our roads are so narrow, and there's nothing to be done about it. Can we make the size configurable (i.e. in stk_config or so), so we easily play with different sizes?

Alayan-stk-2:
It is much easier to change a config option in stk_config than in kart_characteristics (much less files to change), apparently. So it can now be changed from there for easy testing.

I've also slightly increased the icon sizes (+25% for the players and +15% for the AIs). Much less than the +80% for the minimap itself, as they were more reasonably sized and it's better if they can slightly indicate if a kart is on the left/right part of the road.

src/states_screens/race_gui.cpp
```diff
@@ -82,9 +82,27 @@ RaceGUI::RaceGUI()
...
         m_lap_width = font->getDimension(L"9/9").Width;
 
+    float map_size_splitscreen = 1.0f;
+
+    // If there are four players or more in splitscreen
+    // and the map is in a player view, scale down the map
+    if (race_manager->getNumLocalPlayers() >= 4 && !race_manager->getIfEmptyScreenSpaceExists())
+    {
+        if ( ((float) irr_driver->getFrameSize().Width / (float) irr_driver->getFrameSize().Height) >
+             (4.1f/3.0f))
```
auriamg: Where does 4.1/3 come from? That seems pretty arbitrary. That's not necessarily a showstopper, but using a constant with some explanation of where the value comes from would be preferred

data/stk_config.xml
```diff
@@ -122,6 +122,13 @@
   <replay max-time="600" delta-t="0.05"  delta-pos="0.1"
           delta-angle="0.5" />
 
+  <!--  Determines the minimap related values. 
+        size: The size of the minimap (scaled afterwards) 480 = full screen height)
```
auriamg: 480 = full screen height? Not sure I get that part

src/states_screens/race_gui.cpp
Alayan-stk-2: Screen ratios are mainly 16:9 and 4:3. This is a way to do "if you have 16:9 resolutions, do this, if you have 4:3 resolution, do that instead".

data/stk_config.xml
Alayan-stk-2: Look up the code in race_gui.cpp to understand why 480. This is not something I've touched in this PR.

I've put this information in the comment to guide (advanced) users who'd like to tweak their settings ; though ideally their should be a deeper rewrite of race_gui.cpp.

src/states_screens/race_gui.cpp
auriamg: Okay, why the .1 though? And a little comment explaining about screen ratios would be welcome there

Alayan-stk-2: It's there in case there is some pixel off (if STK try to fit the screen in windowed mode, it will go outside of the predefined resolutions ; I haven't tested extensively how this affect the screen ratio but it won't be exactly 4:3 anymore for sure) or rounding error.

auriamg:
I like that it's easily configurable. IMHO I find it still too big, comparing it with other kart games, it's usually smaller than your proposed size. Though if others ( @hiker, @Arthur-D , etc. ? ) think differently my mind could be changed I guess

hiker:
I had a quick look on a small window, and it looked alright, but will have a more thorough full screen look tonight. But considering that the map size can be configured, I am happy to merge this in and adjust the final minimap size.

PR #3243, opened 2018-05-06
Fix #3175

### LINKED ISSUE #3175: STK crash when there are 100 players profile
## Description

After having created 100 profile, going in the profile manager triggers a crash.

## Steps to reproduce

Simply create 100 profile, and then go in the profile manager.

## Configuration

STK 0.9.3

Alayan-stk-2:
The code change is trivial.

There was a badly designed portion of code which assumed that a DynamicRibbonWidget would never ever have more than 50 items in a row, and closed STK if it happened to search an item for more than 50 loop iteration. But the user screen can have an arbitrary high number of player profiles, and so an arbitrary high number of items in its DynamicRibbonWidget, all in the same row.

It becomes hard to use with a lot of profiles on only one line, but at least with this correction there is no crash.

It may be a good idea to have the player list expand on two lines if it becomes big enough to not fit on the screen, but that's a lesser concern than a crash.

(Unrelated : looking at the .stkgui file, I saw that guest accounts are disabled "for now". Any reason for this ?)


deveee:
This is ok for single row, but what if there is dynamic ribbon widget with more rows?

Anyway this loop doesn't look very efficient (for multiple columns it just scrolls by 1) and it's only for GUI, so I don't really care. And indeed m_items.size() is the highest possible value.

Issue #3156, opened 2018-03-20
Ghost race mode improvements


Alayan-stk-2:
Playing a bit with the ghost race mode, I thought of several useful Quality-of-Life improvements : 

- [x]  There is an option to display only the replays corresponding to the current difficulty. Another option to only display the best time for each race would be helpful, as it can quickly get cluttered when you have a dozen of replays for a given race. Usually, it's against the best time you want to race.
- [x] [Obsolete : with the best times options, and the ability to delete unwanted replays, this is not necessary] Often, we want to save the race only if we do a better time. Sure, it is possible to manually abort the race when we are way behind ; but when it's more close, that could be helpful to reduce clutter in the replay list. An option to "save the race but only if the ghost is beaten" would certainly be useful.
- [x] At the end of a race, when we beat the previous ghost, we often want to race against the new ghost, with other options unchanged. Currently, one need to go back to the ghost race menu, find the replay and select it. Adding a direct button in the end-of-race menu to play against the new ghost would help greatly.
- [x] Add a release version information about the replay
- [x] Allow to see only replays from the current version
- [x] Remove the column for difficulty (version) when the option to show only for current difficulty (version) is selected
- [x] Write the replay main data (track, laps, direction, difficulty, time, author) in the modal dialog used to decide what to do (rewatch, race against it, etc).
- [x] Also write information about the replay to compare against, if selected
- [x] Allow replay comparison between replays (including old replays, there just won't be a live time difference).
- [x] Display the nitro amount in the gauge in watch-only mode
- [x] Display the number of items in watch-only mode
- [x] Display attachments (parachute, anvil, bomb...)
- [x] Allow to compare two ghost replays in watch-only mode
- [x] Display live time-difference based on speed and distance on-track (extremely useful to test various strategies and find what works best)
- [x] Support for backward compatibility reading of 0.9.3 replays
- [x] Force a frame recording when some important state changes (the kart starts skidding, using nitro, using an item, receive an attachment, is going to soon cross the finish line). This gives a significantly improved replay accuracy at a small replay size cost (< 20%).
- [x] Save transforms in a more smart way, reducing file size while improving accuracy.
- [x] Don't assume that the replay is for time-trials, but rather write the mode in the replay file. This will be useful to support non-time-trial replays (egg hunts would be the easiest, but it could also work with normal races, FTL and battle)
- [x] Allow to record and replay egg hunts
- [x] Save kart color info in replay files
- [x] Display the correct item type in watch-only mode
[...]

PR #3244, opened 2018-05-07
Numerous improvements related to ghost replays

### LINKED ISSUE #3153: Nitro bar stays empty when rewatching a ghost race
In the ghost race mode, there is an option "watch only".

When it is selected to rewatch a stored race, the GUI is displayed from the ghost kart point-of-view (for example, the speed indicator gives the ghost kart's speed).

However, the nitro bar remains empty, even when there is in fact nitro stored by the ghost kart.

This happens with a race saved as ghost both in 0.9.3 and on the latest dev version.

Expected behavior : the nitro level is stored in the replay data, and shown in "watch only" mode.

Alayan-stk-2:
Too numerous to write them all here. See the full list here : https://github.com/supertuxkart/stk-code/issues/3156 (all checked checkboxes correspond to something this PR fixes)

Also fixes #3153

Demo here : https://www.youtube.com/watch?v=srKezXy8d98


src/karts/ghost_kart.cpp
```diff
@@ -130,9 +135,45 @@ void GhostKart::update(int ticks)
         m_all_physic_info[idx].m_steer, m_all_physic_info[idx].m_speed,
         /*lean*/0.0f, idx);
 
+    // Attachment
+    // Uses custom format so that replay files remain valid if
+    // internal numbering changes.
+
+    int attach_ticks = 9000;
```
auriamg: Hardcoded arbitrary constant

src/modes/linear_world.cpp
```diff
@@ -262,6 +268,45 @@ void LinearWorld::updateGraphics(float dt)
 
 }   // updateGraphics
 
+// ----------------------------------------------------------------------------
+/** This calculate the time difference between the second kart in the
+ *  race and the first kart in the race (who must be a ghost)
+ */
+void LinearWorld::updateLiveDifference()
```
auriamg: Not a big fan of adding so much ghost-specific stuff to linear world, isn't it possible to move that to some ghost class?

src/modes/linear_world.cpp
```diff
@@ -262,6 +268,45 @@ void LinearWorld::updateGraphics(float dt)
...
+{
+    // First check that the call requirements are verified
+    assert (race_manager->hasGhostKarts() && race_manager->getNumberOfKarts() >= 2);
+
+    AbstractKart* ghost_kart = getKart(0);
```
auriamg: Not a huge fan of hardcoding kart 0 / kart 1. Constants explaining which is which would be better

src/modes/linear_world.cpp
```diff
@@ -262,6 +268,45 @@ void LinearWorld::updateGraphics(float dt)
...
+
+    // Check when the ghost what at this position
+    float ghost_time;
+
+    // If there are two ghost karts, the view is set to kart 0,
+    // so switch roles in the comparison. Note that
+    // we can't simply multiply the time by -1, as they are assymetrical.
+    // When one kart don't increase its distance (rescue, etc),
+    // the difference increases linearly for one and jump for the other.
+    if (getKart(1)->isGhostKart())
```
auriamg: I'm not sure kart 1 is garanteed to exist

src/replay/replay_recorder.cpp
```diff
@@ -42,6 +45,12 @@ ReplayRecorder::ReplayRecorder()
 {
     m_complete_replay = false;
     m_incorrect_replay = false;
+
+    // Give margin to store extra events due to higher precision
+    // when a kart's tracked charateristic has suddenly changed
+    // in a non-interpolable way.
+    m_max_frames = (unsigned int)( 1.2f * stk_config->m_replay_max_time
```
auriamg: 1.2, hardcoded magic constant

src/karts/abstract_kart.hpp
```diff
@@ -323,8 +328,8 @@ class AbstractKart : public Moveable
      *  \param fade_out_time How long the maximum speed will fade out linearly.
      */
     virtual void instantSpeedIncrease(unsigned int category, float add_max_speed,
-                                     float speed_boost, float engine_force, 
-                                     int duration, int fade_out_time) = 0;
+                                    float speed_boost, float engine_force, float duration,
+                                    float fade_out_time) = 0;
```
hiker: Big NO, don't re-introduce floats here, they need to stay ticks.

src/replay/replay_recorder.cpp
```diff
@@ -116,7 +127,81 @@ void ReplayRecorder::update(int ticks)
 #ifdef DEBUG
         m_count++;
 #endif
-        if (time - m_last_saved_time[i] < stk_config->m_replay_dt)
+        // If one of the tracked kart data has significantly changed
+        // for the kart, update sooner than the usual dt
+        bool force_update = false;
+
+        int attachment;
+        if (kart->getAttachment()->getType() == Attachment::ATTACH_NOTHING)
```
auriamg: An enum already is an int, why have all that code to convert the enum to an int?

src/replay/replay_recorder.cpp
```diff
@@ -116,7 +127,81 @@ void ReplayRecorder::update(int ticks)
...
+            // If close to the end of the race, reduce the time step
+            // for extra precision for the whole race time
+            if (race_manager->isLinearRaceMode())
+            {
+                float full_distance = race_manager->getNumLaps()
+                        * Track::getCurrentTrack()->getTrackLength();
+
+                const LinearWorld *linearworld = dynamic_cast<LinearWorld*>(World::getWorld());
+                if (full_distance + 0.5f >= linearworld->getOverallDistance(i) &&
+                    full_distance <= linearworld->getOverallDistance(i) + 10.0f)
```
auriamg: 10.0, 0.5, lots of magic numbers all over

src/karts/ghost_kart.cpp
```diff
@@ -130,9 +135,45 @@ void GhostKart::update(int ticks)
...
+        m_attachment->set(Attachment::ATTACH_ANVIL,attach_ticks);
+    else if ( m_all_bonus_info[idx    ].m_attachment == 3 &&
+              m_attachment->getType() != Attachment::ATTACH_BOMB)
+        m_attachment->set(Attachment::ATTACH_BOMB,attach_ticks);
+    else if ( m_all_bonus_info[idx    ].m_attachment == 4 &&
+              m_attachment->getType() != Attachment::ATTACH_SWATTER)
+        m_attachment->set(Attachment::ATTACH_SWATTER,attach_ticks);
+    else if ( m_all_bonus_info[idx    ].m_attachment == 5 &&
+              m_attachment->getType() != Attachment::ATTACH_BUBBLEGUM_SHIELD)
+        m_attachment->set(Attachment::ATTACH_BUBBLEGUM_SHIELD,attach_ticks);
```
hiker: Can you either change m_attachment to be proper attachment type, or alternatively at least cast the integers to the type for a comparison? That should be much much shorter and easier to understand.

src/replay/replay_recorder.cpp
```diff
@@ -231,28 +332,53 @@ void ReplayRecorder::save()
                 StringUtils::xmlEncode(kart->getController()->getName()).c_str());
     }
 
+    // Calculate the unique identifier
+    // Part of it is random
+    // Part of it is based on race data
+    unsigned long long int unique_identifier = 0;
```
auriamg: What is the unique identifier needed for? Anyway, that should probably go to a new method dedicated to generating it. Also, this code is fairly weird, using a well-known hash function or UUIDs would be better if truly necessary

src/replay/replay_recorder.hpp
```diff
@@ -74,6 +82,8 @@ class ReplayRecorder : public ReplayBase
     void  save();
     void  update(int ticks);
 
+    unsigned long long int getLastUID() { return m_last_uid; }
```
auriamg: methods that don't modify the object should be const

src/states_screens/dialogs/ghost_replay_info_dialog.hpp
```diff
@@ -33,21 +31,31 @@ class GhostReplayInfoDialog : public GUIEngine::ModalDialog
...
 private:
 
-    bool m_self_destroy;
+    bool  m_self_destroy;
 
-    bool m_record_race;
-    bool m_watch_only;
+    bool  m_record_race;
+    bool  m_watch_only;
+    bool* m_compare_ghost;
```
auriamg: Why is this a pointer?

src/states_screens/dialogs/ghost_replay_info_dialog.hpp
```diff
@@ -33,21 +31,31 @@ class GhostReplayInfoDialog : public GUIEngine::ModalDialog
...
+
+    unsigned int m_replay_id; // May be updated on list refreshes
+
+    unsigned long long int* m_compare_replay_uid;
```
auriamg: Why is this a pointer?

src/states_screens/race_gui.cpp
```diff
@@ -390,6 +404,65 @@ void RaceGUI::drawGlobalTimer()
...
+
+    // Don't display the live difference timer if its time is wrong
+    // (before crossing the start line at start or after crossing it at end)
+    if (!linearworld->hasValidTimeDifference())
+        return;
+
+    float live_difference = linearworld->getLiveTimeDifference();
+
+    int timer_width = m_small_precise_timer_width;
+    if (fabsf(live_difference) >= 59.9995f)
```
auriamg: What is this magic number?

src/karts/abstract_kart.hpp
```diff
@@ -307,8 +312,8 @@ class AbstractKart : public Moveable
      *  \param fade_out_time How long the maximum speed will fade out linearly.
      */
     virtual void increaseMaxSpeed(unsigned int category, float add_speed,
-                                  float engine_force, int duration,
```
auriamg: Do NOT replace ticks with floats. Lots of work was done to switch from imprecise floats to deterministic ints, please do not undo that

src/karts/ghost_kart.cpp
hiker: We have in several instances delays in seconds hard coded (e.g. how long it takes for an item to reappear). While it might not be worth to make everything configurable, at least use:
stk_config->time2Ticks(75);   
so that we can change the tick duration without changing the game play.

src/states_screens/race_gui.cpp
Alayan-stk-2: There are 60s to a minute (I think we can assume that this won't change anytime soon).

Now, add that it is rounded to milliseconds. What's the smallest number of seconds who'd get rounded to 1 minutes ? This is it.

src/states_screens/dialogs/ghost_replay_info_dialog.hpp
Alayan-stk-2: I need to pass back the value to the main state screen, otherwise the value would be lost as soon as the modal dialog is closed.

This works as follow : 
1)GhostReplaySelection passes pointers to two values it needs the dialog to update (one bool to know if there is a comparison active, one to store the UID)
2)The dialog store the pointers value in members to be able to reuse them throughout the class without a complicated scheme to pass them as arguments
3)It uses this to update the values when appropriate
4)When the dialog closes, GhostReplaySelection stlll has the values it needs.

src/modes/linear_world.cpp
Alayan-stk-2: Isn't it guaranteed to exist if we first guarantee that there are two karts, as is done here ?

src/karts/ghost_kart.cpp
Alayan-stk-2: Yeah, that part is arbitrary ; it's just putting a big enough number (this corresponds to over a minute) to make sure the attachment won't go away until it is explicitly cleared. As the replay file knows when it should be done, there is no point in trying to recalculate it, especially as the values used to calculate it may have changed since the replay file has been saved. The attachment can't be set each frame, because this would reset its graphic size and keep it tiny.

The ticks value can't be read from the replay file as it's not saved here (and considering it would only serve a few times per replay at most, saving it in hundreds of lines may be overkill).

The downside is that with this way of doing it, the clocktime of the bomb will be wrong and that the bubbleshield won't flicker before disappearing (don't matter currently as none of it can be gotten in time-trial, but to keep in mind if expanding replays to more type of races).

src/replay/replay_recorder.cpp
Alayan-stk-2: Some unique identifier is needed for comparisons, as the temporary IDs are unstable (list sorts can shuffle files around, for example).

Calling classic hash functions/UUID is overkill for the use of this identifier (while part of the goal is that shared ghost files have different IDs, a random collision would not matter). Also, reversibility of the UID is a feature. Baking in some race data in it is useful for some consistency checks on a replay file. I'd even add more general data if I could have a 128-bits number rather than a 64-bits one.

Moving this to a separate function would certainly be better.

src/replay/replay_recorder.cpp
Alayan-stk-2: The memory to store the replay data is allocated at the start of the recording, for m_max_frames.

While it would be possible to update m_max_frames each time there is a forced update, this would also mean having to resize several arrays each and every time it happens. Though the forced updates don't happen that often, copying huge arrays from one place in memory to another dozens of times would be very wasteful - especially as it is unlikely for m_max_frames to be reached anyway.

If updating dynamically is wasteful, then giving some room at the start for forced updates to fit within the replay max time is more efficient. It doesn't need to be perfect, just to work as long as the player is not trying to break things, by staying entire minutes just before the finish line in the last lap, or spending minutes to catch and burn some nitro, or that sort of thing.

Sure, it would be possible to add an extra config option for this, but is it really worth it ?

src/replay/replay_recorder.cpp
Alayan-stk-2: Sure, this code is not strictly necessary at the moment, and it's not elegant either, but this is a precaution which will one day prove useful.

Why ? Because the underlying values behind "ATTACH_NOTHING", etc. can change. Say, for example, you add a new attachment (this will happen, maybe in five years, but it will). You'll put it before ATTACH_MAX and ATTACH_NOTHING ; and this will increment the value of ATTACH_NOTHING.

You won't think for a second that you've broken something by adding a new attachment type. Everything will seem to work while you'll test this attachment in normal play. It'll even work in new replays you'd record after the change. But if there is not this conversion to guarantee the independence of the stored replay values from the internal ordering (storing by meaning rather than by raw encoding), you'll have effectively broken all existing version 4+ replays, which may well be all the players replays from the last 2-3 years. If this is caught before release, probably in RC testing, you'd only have to do a code similar to the one I did here for things to keep working. Worse case scenario, this gets into release. You'd be stuck between dropping support for old replays and adding code to manage two possible encodings - the latter solution being definitely more heavy and less elegant that this code.

Note that I could have added similar code for skidding_state, but I estimated that in that case, it was extremely unlikely for the enum to change, even more twice.

src/karts/ghost_kart.cpp
Alayan-stk-2: m_attachment is deliberately not an attachment type, see my answer to Auria below. Using the enum names as a basis rather than their raw values makes things work semantically rather than assuming a specific encoding, and avoid having old replays broken in new version if that encoding change.

That's actually the whole point of using enums with pretty names (I remember getting some flak for using raw int in an early PR).

The key difference is that the encoding for the whole code is defined by what's in the enum at compilation time, while for the replays it depends on the executable binaries at save time. So while you don't need extra-care inside the code base, when interacting with files which may have been produced with other binaries, it is a requirement.

Think of those times where you opened a text file only to discover it was broken because your editor assumed the wrong encoding (and that's best case, worse case someone did save broken values after having opened the document and received encoding errors ; breaking it completely).

Also, sure, I could default to having the number used the same as the values from the current enum, but it wouldn't be safe unless I'd keep that section of (admittedly not pretty) elseifs. And if we're stuck with them, setting "NOTHING" to "0" is better than setting "NOTHING" to "9" (current value in the enum). And using 0 to 5 is better than 0 to 3, 6 and 9. I can use a different encoding, and it does not break because the interface between the two is semantical.

src/states_screens/dialogs/ghost_replay_info_dialog.hpp
Benau: GhostReplaySelection::getInstance()->setComapreReplayUid(uint64_t uid);

data/gui/ghost_replay_info_dialog.stkgui
```diff
@@ -1,16 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
Benau: http://www.downloadclipart.net/browse/1612/cartoon-ghost-clipart
Image
Can not be used with any content found on this website for commercial purposes.

src/karts/ghost_kart.cpp
Benau: First, it's always time trial so you can only see anvil and parachute.

And if you add it an attachment for ghost kart you will have translucent kart with solid attachment mesh?

data/gui/ghost_replay_info_dialog.stkgui
Alayan-stk-2: Right, with everything else I forgot to talk about the picture.

First, let's mention that its unnecessarily large (512px when other icons are 128px). I simply couldn't get GIMP to downscale it with proper antialising, and I gave up. Second, though I edited the picture to (aside from adding the blue button) increase the border's thickness, in some resolutions it doesn't seem entirely right.

Now about the license. This image is found on several websites, as they copy each other. The question then is who is the original source.

I got it from here, where it is listed as CC0 : https://pixabay.com/en/ghost-halloween-ghostly-horror-35852/

Benau: There is vector_art/open game art icons/ghostrace.svg in media repo 100% original by myself so you can do whatever you like (and i can declare it's public domain right now)

and unknown original source is not acceptable

or couldn't you use a simple blue cross?

Alayan-stk-2: I first tried to look up for the ghost used in-game, but didn't think about the media repo.

It's possible to use a simple blue cross, sure. I just think it's nicer with a ghost.

I'll sort it out tomorrow - either I can find the original for what I used, or I'll simply do a new version based on yours as then the license question is easy.

Alayan-stk-2: OK, I think I found the original source : https://openclipart.org/detail/17847/cartoon-ghost

Added in 2008 by a user named lemmling, who also added a few dozens other vector images in around a month, before mostly stopping except two images in 2009 and 2010. Not the sort of profile of someone who's stealing pictures from others.

A japanese illustration site has also tracked down the picture there, see the source button : http://free-illustrations.gatag.net/2013/11/17/190000.html

The ghost's license is indeed CC0.

> We use the Creative Commons Zero 1.0 Public Domain License every time an artist uploads a piece of clipart to Openclipart to make it clear the artist is releasing the creative work for anyone to use for any reason, even commercially. This act of "sharing" is the foundation Openclipart is based upon. Sometimes people want a disclaimer or legal assurance that our clipart is safe to be used. This extra security request often comes from people working for companies of all sizes and, or, publishers. If you need this type of document, please visit, save, download and, or print the following for your needs: Creative Commons Zero 1.0 License.

From : https://openclipart.org/share

Benau: Then remember to edit License.txt in data/gui

src/states_screens/dialogs/ghost_replay_info_dialog.hpp
Alayan-stk-2: Ok, thanks for the tip, I will try and change it accordingly

src/states_screens/race_gui.cpp
auriamg: implicit in my question was : adding a comment like "// smallest number of seconds that could get rounded to 1 minute" would help clarify a lot. Or even better, if this is about rounding, then checking "rounding(...) >= 60; // 60 seconds" would be even clearer

src/modes/linear_world.cpp
auriamg: Ok I'd missed the >= 2 part above

src/karts/ghost_kart.cpp
Alayan-stk-2: It's currently always time trial, but there is nothing inherently preventing to expand replays in the future to normal races. Not a priority, certainly, but a real possibility.

Yes, the translucent kart with solid attachment is currently an issue. If you know how to make the attachment translucent too, tell me, it'll save me some time.

Benau: in attachment constructor if ghost kart
    m_node = irr_driver->addAnimatedMesh(
                         attachment_manager->getMesh(Attachment::ATTACH_BOMB), "bomb",
                         NULL, std::make_shared<RenderInfo>(0.0f, true));

src/states_screens/race_gui.cpp
Alayan-stk-2: Using a rounding function is not very practical - I'd need to multiply by 1000 to round to milliseconds, then divide by 1000 again to compare to 60. I've added a comment.

Benau:
auria do you really want / need easter mode ghost replay?

Alayan-stk-2:
Summary of the latest changes : 
- Adds a race type info to the replay file
- Adds replay support for egg hunts
- Update the replay selection UI for egg hunts replays
- Smooths the displayed speed in watch-only mode with interpolation
- Reduce the number of unneeded transforms when interpolation allows a larger dt. On average, the replay files are now smaller than before, despite storing additional parameters and being in most situations more accurate (by triggering new transforms earlier if an important kart characteristic changes).
- Reset comparison when lauching a comparison race
- Make attachment transparents for ghost karts
- Support of kart color
- Use dedicated functions to encode/decode attachment and item types
- Support item types in replay recording and decoding - it would even be possible to enable recording for normal races with few changes

auriamg:
@Alayan-stk-2 We can discuss adding replay to other game modes and decide if it's a feature we're interested in, but PLEASE stop adding new features to pull requests that you've declared final and that we've already started to review, at such stage ideally the PR should only receive bugfixes and requested changes, that will make things much easier. Thank you

src/karts/ghost_kart.cpp
```diff
@@ -130,9 +138,60 @@ void GhostKart::update(int ticks)
...
         /*lean*/0.0f, idx);
 
+    // Attachment management
+    // FIXME : don't handle if an attachment is renewed
+    //         e.g., a bubble shield is used again with another active
+    //         potentially reaching arbitrary high active time
+
+    Attachment::AttachmentType attach_type = ReplayRecorder::codeToEnumAttach(m_all_bonus_info[idx].m_attachment); 
+    int attach_ticks = 0;
+    // FIXME The exact value depends on kart properties. Should it be written in the replay file ?
```
auriamg: I think this is only a visual effect so I don't think accuracy is all that important, but maybe add a little comment explaining those are approximate durations used for visualisation only?

src/modes/easter_egg_hunt.cpp
```diff
@@ -154,6 +162,16 @@ void EasterEggHunt::collectedEasterEgg(const AbstractKart *kart)
...
 }   // collectedEasterEgg
 
+//-----------------------------------------------------------------------------
+/** Called when a ghost kart has collected an egg.
+ *  \param world_id The world id of the ghost kart that collected an egg.
+ */
+void EasterEggHunt::collectedEasterEggGhost(int world_id)
+{
+    m_eggs_collected[world_id]++;
+    printf("Egg collected, new total is %d\n", m_eggs_collected[world_id]);
```
auriamg: printf should not be left in the code. If it's valuable logs, use our Log:: framework. If not valuable, remove

src/modes/easter_egg_hunt.cpp
```diff
@@ -169,8 +187,17 @@ void EasterEggHunt::update(int ticks)
  */
 bool EasterEggHunt::isRaceOver()
 {
-    if(m_eggs_found == m_number_of_eggs)
+    if(!m_only_ghosts && m_eggs_found == m_number_of_eggs)
         return true;
+    else if (m_only_ghosts)
+    {
+        printf("The game think there are only ghosts\n");
```
auriamg: printf

src/replay/replay_play.cpp
```diff
@@ -283,45 +381,136 @@ void ReplayPlay::readKartData(FILE *fd, char *next_line)
...
+{
+    for(unsigned int i = 0; i < m_replay_file_list.size(); i++)
+    {
+        if (m_replay_file_list[i].m_replay_uid == uid)
+        {
+            return i;
+        }
+    }
+
+    return 0; //keep the compiler happy
```
auriamg: Actually, I wouldn't say it's "keep the compiler happy", not finding a searched element seems like it should be handled error handling. Maybe return -1, or assert(false), or another way to signal it was not found?

src/replay/replay_play.hpp
```diff
@@ -124,9 +138,21 @@ class ReplayPlay : public ReplayBase
     // ------------------------------------------------------------------------
     void               setReplayFile(unsigned int n)
                                                 { m_current_replay_file = n; }
+
+    // ------------------------------------------------------------------------
+    void               setSecondReplayFile(unsigned int n, bool second_replay_enabled)
+                           {                      m_second_replay_file = n;
```
auriamg: weird indentation

src/replay/replay_recorder.cpp
```diff
@@ -116,13 +129,125 @@ void ReplayRecorder::update(int ticks)
...
+        int attachment = enumToCode(kart->getAttachment()->getType());
+        int powerup_type = enumToCode(kart->getPowerup()->getType());
+        int special_value = 0;
+
+        // In egg hunt mode, use store the number of eggs found so far
+        // This assumes that egg hunt mode is only available in single-player
+        // TODO : when adding battle mode support, store the number of lives here
+        if (race_manager->isEggHuntMode())
+        {
+            //TODO : figure out why the compiler don't like calling numberOfEggsFound
```
auriamg: Well that should be figured out and the TODO removed before this can be merged :)

src/replay/replay_recorder.cpp
```diff
@@ -116,13 +129,125 @@ void ReplayRecorder::update(int ticks)
...
+            // If nitro stops being used
+            // (also generates an extra transform on collection,
+            //  should be negligible and better than heavier checks)
+            if (kart->getEnergy() == b_prev->m_nitro_amount &&
+                b_prev->m_nitro_amount != b_prev2->m_nitro_amount)
+                force_update = true;
+
+            // If close to the end of the race, reduce the time step
+            // for extra precision
+            // TODO : fast updates when close to the last egg in egg hunt
```
auriamg: A left-over TODO?

src/replay/replay_recorder.cpp
```diff
@@ -267,13 +462,96 @@ void ReplayRecorder::save()
...
+
+    return code;
+} // enumToCode
+
+/* Returns an encoding value for a given item type
+ * \param type : the type of item to encode */
+
+int ReplayRecorder::enumToCode (PowerupManager::PowerupType type)
+{
+    int code =
```
auriamg: All of that could be replaced with (int)type, no? An enum already has an integer value

src/replay/replay_recorder.cpp
```diff
@@ -267,13 +462,96 @@ void ReplayRecorder::save()
...
+/* Returns an encoding value for a given attachment type.
+ * The internal values of the enum for attachments may change if attachments
+ * are introduced, removed or even reordered. To avoid compatibility issues
+ * with previous replay files, we add a layer to encode an enum semantical
+ * value the same way, independently of its internal value.
+ * \param type : the type of attachment to encode */
+
+int ReplayRecorder::enumToCode (Attachment::AttachmentType type)
+{
+    int code =
```
auriamg: All of that could be replaced with (int)type, no? An enum already has an integer value

src/states_screens/dialogs/ghost_replay_info_dialog.cpp
```diff
@@ -59,17 +96,98 @@ GhostReplayInfoDialog::GhostReplayInfoDialog(unsigned int replay_id)
...
 GhostReplayInfoDialog::~GhostReplayInfoDialog()
 {
 }   // ~GhostReplayInfoDialog
 
+// -----------------------------------------------------------------------------
+void GhostReplayInfoDialog::updateReplayDisplayedInfo()
+{
+    m_replay_info_widget->clear();
+
+    int n=2;
```
auriamg: Can you maybe find a more descriptive name that "n" so it's easier to understand what it is?

src/states_screens/dialogs/ghost_replay_info_dialog.cpp
```diff
@@ -59,17 +96,98 @@ GhostReplayInfoDialog::GhostReplayInfoDialog(unsigned int replay_id)
...
+    if (m_compare_ghost)
+        n++;
+
+    bool is_linear = GhostReplaySelection::getInstance()->isActiveModeLinear();
+
+    for (int i=0; i<n; i++)
+    {
+        int id;
+
+        if (i<=1)
```
auriamg: It's unclear to me why replays 0-1 are special?

src/states_screens/ghost_replay_selection.cpp
```diff
@@ -97,36 +139,193 @@ void GhostReplaySelection::loadList()
...
+            (translations->fribidize(track->getName()) , -1, 9));
         row.push_back(GUIEngine::ListWidget::ListCell
-            (StringUtils::toWString(rd.m_kart_list.size()), -1, 1, true));
+            (StringUtils::toWString(rd.m_kart_list.size()), -1, 3, true));
+        if (m_active_mode_is_linear)
+            row.push_back(GUIEngine::ListWidget::ListCell
+                (rd.m_reverse ? _("Yes") : _("No"), -1, 3, true));
+        if (!m_same_difficulty)
+            row.push_back(GUIEngine::ListWidget::ListCell
+                (rd.m_difficulty == 3 ? _("SuperTux") : rd.m_difficulty == 2 ?
```
auriamg: RaceManager::getDifficultyName could be reused instead of hardcoding this here (I know this isn't exactly your code, but if the code is being changed we might as well improve it)

src/states_screens/race_gui.hpp
```diff
@@ -120,6 +130,9 @@ class RaceGUI : public RaceGUIBase
     /** Display items that are shown once only (for all karts). */
     void drawGlobalMiniMap     ();
     void drawGlobalTimer       ();
+    //TODO : support live difference computation outside of ghost replays
```
auriamg: Is this TODO really useful? We needen't specify all possible future eventual improvements as todos all over the code

src/modes/easter_egg_hunt.cpp
Alayan-stk-2: Sorry, forgot to remove those.

src/karts/ghost_kart.cpp
hiker: Also that line looks way too long - ~80 characters maximum!

hiker:
One general comment: can you please push less frequently? E.g. you had 11 pushes within less than 40 minutes! That means all of us devs get 11 emails (not to mention potentially several CI related emails to those). Certainly do individual/smaller commits, but push only when you are sure you are done (or at least only when you are done for the day and want a backup).

src/replay/replay_recorder.cpp
Alayan-stk-2: This one is actually not a left-over : it would need code to compute the kart's distance with still available egg items. This is doable, but a bit complex for a small benefit, so I didn't want to stall the PR for it.

src/modes/linear_world.cpp
hiker: +1

src/replay/replay_recorder.cpp
Alayan-stk-2: See my previous comments on the topic in this PR, to you and hiker. This conversion layer is here so that the replay format is not dependent on the enum's internal structure. Encapsulation, in a way : don't rely on the internals. If those enums are changed (new item or attachment added, for example, or assigning new IDs to have easier access on a subset like it has been done for minor race modes), those functions allow replays written before the changes to keep working automatically, instead of having them broken, having to debug the issue when it'll surface out, and then either dropping support for old replays or adding a conversion layer.

src/states_screens/dialogs/ghost_replay_info_dialog.cpp
Alayan-stk-2: Actually, there are 3 things this function push to the list : the header, the data for the current selected replay (the one the player whose selection triggered the modal dialog), and the replay used for comparison.

As the code for the first replay and for the second replay is mostly identical, I do a if/else to get the id, get the ReplayData from it, and then use it. The header don't need that ReplayData, but in this setup I need to feed some id value to it.

It would probably clearer if I moved the header code out of the loop.

src/karts/ghost_kart.cpp
```diff
@@ -130,9 +138,60 @@ void GhostKart::update(int ticks)
...
+    if (attach_type == Attachment::ATTACH_BUBBLEGUM_SHIELD)
+        stk_config->time2Ticks(10);
+    else if (attach_type == Attachment::ATTACH_BOMB)
+        stk_config->time2Ticks(30);
+    // The replay history will take care of clearing,
+    // just make sure it won't expire by itself
+    else
+        stk_config->time2Ticks(300);
```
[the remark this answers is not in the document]
Alayan-stk-2: Good catch, I forgot the assignation.

src/karts/ghost_kart.hpp
```diff
@@ -42,11 +42,15 @@ class GhostKart : public Kart
...
+    std::vector<ReplayBase::BonusInfo>       m_all_bonus_info;
+
     std::vector<ReplayBase::KartReplayEvent> m_all_replay_events;
 
+    unsigned int                             m_last_egg_idx = 0;
+
 public:
-                  GhostKart(const std::string& ident,
-                            unsigned int world_kart_id, int position);
+                  GhostKart(const std::string& ident, unsigned int world_kart_id,
```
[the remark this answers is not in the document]
Alayan-stk-2: 81 characters according to my editor for the whole line. I'll keep it like that.

src/modes/linear_world.cpp
Alayan-stk-2: This doesn't require to be tied to ghosts (i.e., it could potentially be used elsewhere in the future), so I'm not a fan either of moving it to a ghost-specific class.

Maybe it could use its own class, somewhere in the kart or race directory ? (not sure about the most appropriate directory) It would be a relatively small one. If that way is chosen, I'd rewrite it to be able to work with more than 2 karts.

auriamg:
This happened when I tried to ghost replay the Northern Resort replay that comes with STK (look at the minimap, there's a random wilber icon flying around)
https://www.amazon.ca/clouddrive/share/LzJhwGL1HyRZpMWNkKSDdNv5vLpkw6oJJT5oEf86tmZ

This happened when I tried to compare ghost replays:
https://www.amazon.ca/clouddrive/share/KZhLB4JEITTvwJJhx8Ffb7UZP2moguDWBZNAPNueKvm

Alayan-stk-2:
For the random wilber, it seems that there is a ghost which is compared when it shouldn't - and a ghost from another track at that, which is why he flies around in a way unrelated to Northern Resort.

Same thing later, the wilber replay is clearly from another track.

When you click compare ghost, it should only displays ghosts with the same track/direction/laps. Also, when you choose to not be in watch-only mode (first case), it's supposed to disable the replay comparison. From your video, it's clear that there are ways to go around this. What would be helpful to fix that is thus to know what you did in the GUI.

auriamg:
Links including the UI - all videos taken right after starting STK
https://www.amazon.ca/clouddrive/share/VUNJG5MpvwIprORUlt9iLAGF3vQMRk4HXtNVtyaC6nq
https://www.amazon.ca/clouddrive/share/GEbb9ArhcrEVyUUscjRmppAIIgz5XJ74hbknRoZ9hZd

Issue #3160, opened 2018-03-22
Add an in-game link to STK account password reset


Alayan-stk-2:
Currently, there is no in-game indication that the STK's account password can be reseted, or how.

Fix : in the options menu, where the player is given the option to connect to the server, add a "Forgot password ?" link directing to : http://addons.supertuxkart.net/password-reset.php

Issue #1946, opened 2015-02-01
Add a reset password feature in game


math07:
Hey,
Congratulation for releasing beta 2 !

1- As an improvement, I think an option in game to reset our password is important or at least a link to the stkaddons website.
2 - Are accounts still tied to the accounts website ? I reset my password on 
http://stkaddons.net/password-reset.php but the game still says the new password is invalid. This might be a bug.

konstin:
to 2): Retry at http://addons.supertuxkart.net

math07:
konstin :Thanks 2) isn't a bug then.

For 1), I see two options :
In those two screens, there could be a label Forgot password? with the website link or a button Forgot password? that opens a pop up in game to enter name/email/captcha and send the email.
![a1](https://cloud.githubusercontent.com/assets/9955284/5992683/597e1766-aa01-11e4-9404-14c5007a9c9e.png)
![a2](https://cloud.githubusercontent.com/assets/9955284/5992684/607b4a48-aa01-11e4-831e-c4d0a1d944bd.png)



In the login screen where there is the

yehudaclinton:
but what if i forgot my username, so now i cant make new account with that email
is there any way to reset with just email address

qwertychouskie:
Fixed in https://github.com/supertuxkart/stk-code/pull/3252.

PR #3252, opened 2018-05-15
Donate and reset password buttons

### LINKED ISSUE #3141: Donation button inside the game [Feature Request]
Put a link to https://supertuxkart.net/Donate somewhere inside the game, ideas:

-  Create a button named _Support_ or _Donate_ with an icon like the heart of the website(or any other icon), write a short description and put the donate link
- Put the link inside the _( i ) About_ screen

qwertychouskie:
I tested it on a few *buntu VMs and on Wine, and it worked, though it should be tested on macOS and real Windows before merging.  Closes #3141, #3160, #1946.


deveee:
I'm not convinced to this feature on linux, because there is a lot window managers and browsers that may behave differently. For example if you have a window manager that doesn't support netwm, then still you will have displayed STK over your opened web browser. Or some broken window managers may change your resolution when you leave STK and not change it back after closing a browser etc...

I also found such topic:
https://www.reddit.com/r/androiddev/comments/2vsyq4/beware_google_bans_app_from_playstore_because_of/
so I'm not sure if adding donation link is good idea.

Alayan-stk-2:
It would need to be disabled for Android then, but Google's selfish rules ("we want our share of your donations !") should certainly not decide what go into the main release.

deveee:
Google is rather strict for applications on the play store :P But indeed ```#ifndef ANDROID``` is ok.

But anyway I think that leave fullscreen on linux is unsafe. I had a lot of issues in the past (for example game window was black when I returned to the game after alt+tab, on some older gnome versions it wasn't even possible to use alt+tab or move to secod desktop, because fullscreen window was displayed on top of everything etc.). Even now on ubuntu 18.04 nvidia blob enables panning for some reason when I set lower resolution in the game and it causes a mess. Plus the problem with netwm (some exotic WMs don't support it).

Anyway it's not that big problem, because nobody is forced to click on this link. I just can't guarantee that it will work for everyone.

data/gui/online/register.stkgui
```diff
@@ -34,6 +34,8 @@
                     <textbox id="password" proportion="2" height="fit" I18N="In the registration dialog"/>
                 </div>
 
+                <!--<button id="password_reset" width="fit" text="Reset password"/>-->
```
auriamg: What is this commented line doing there?

src/online/link_helper.cpp
```diff
@@ -0,0 +1,40 @@
...
+#ifdef _WIN32
+#include <windows.h>
+#include <shellapi.h>
+#endif
+
+using namespace Online;
+
+namespace Online
+{
+    void LinkHelper::OpenURL (std::string url) {
```
auriamg: Braces go on a newline

src/online/link_helper.cpp
```diff
@@ -0,0 +1,40 @@
...
+#ifdef _WIN32
+        ShellExecuteA(NULL, "open", url.c_str(), NULL, NULL, SW_SHOWNORMAL);
+#else 
+#ifdef APPLE
+        std::string op = std::string("open ").append(url);
+#else
+        std::string op = std::string("xdg-open ").append(url);
+#endif
+        system(op.c_str());
```
auriamg: You should handle the return code and display an error message if an error code is returned. This is shaky as it is, error handling is not optional

src/states_screens/credits.cpp
```diff
@@ -393,6 +394,10 @@ void CreditsScreen::eventCallback(GUIEngine::Widget* widget,
     {
         StateManager::get()->escapePressed();
     }
+    if (name == "donate")
+    {
+        Online::LinkHelper::OpenURL("https://SuperTuxKart.net/Donate"); // Open donation page
```
auriamg: This should preferably come from the stk_config.xml config file and not be hardcoded, I think? @hiker would you agree?

src/states_screens/register_screen.cpp
```diff
@@ -428,6 +432,10 @@ void RegisterScreen::eventCallback(Widget* widget, const std::string& name,
             onEscapePressed();
         }
     }
+    else if (name == "password_reset")
+    {
+        Online::LinkHelper::OpenURL("http://addons.SuperTuxKart.net/password-reset.php"); // Open password reset page
```
auriamg: Same thing, should ideally come from a config file

src/states_screens/user_screen.cpp
```diff
@@ -342,6 +346,10 @@ void BaseUserScreen::eventCallback(Widget* widget,
         }
         makeEntryFieldsVisible();
     }
+    else if (name == "password_reset")
+    {
+        Online::LinkHelper::OpenURL("http://addons.SuperTuxKart.net/password-reset.php"); // Open password reset page
```
auriamg: config file

auriamg:
Help would be welcome to test this on various linux distributions. It's true that we probably can't get it to work everywhere, though if it works for most that would be acceptable

nado:
Tested with xmonad, it opens correctly in the browser, the text on the buttons disappears when resized under a certain size.
It works the same in fullscreen or tiling : it is the size that matters.

Built with `cmake .. -DUSE_FRIBIDI=0 -DUSE_WIIUSE=0 -DCMAKE_BUILD_TYPE=Release`

auriamg:
On hold until the requested changes have been applied

deveee:
I can confirm that it works for me under Gnome. But please add the ```#if !defined(ANDROID)``` to avoid problems on google play. The ```xdg-open``` won't work on android anyway.

src/states_screens/register_screen.cpp
auriamg: This still applies

src/states_screens/credits.cpp
auriamg: This still applies

src/online/link_helper.cpp
```diff
@@ -0,0 +1,49 @@
...
+            Log::error("OpenURL", "Command returned non-zero exit status");
+        }
+#elif defined(__linux__) && !defined(__ANDROID__)
+        std::string command = std::string("xdg-open ").append(url);
+        if (system(command.c_str()))
+        {
+            Log::error("OpenURL", "Command returned non-zero exit status");
+        }
+#else
+        Log::error("OpenURL", "Not implemented for this platform!");
```
auriamg: In cases where it's not supported, the donate button should be hidden.No use having a button that does nothing when you click on it

src/states_screens/user_screen.cpp
auriamg: Still applies

src/online/link_helper.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+using namespace Online;
+
+namespace Online
+{
+    void LinkHelper::OpenURL (std::string url)
+    {
+#if defined(_WIN32)
+        ShellExecuteA(NULL, "open", url.c_str(), NULL, NULL, SW_SHOWNORMAL);
+#elif defined(APPLE)
```
auriamg: should be __APPLE__

jpenguin:
Apple's Clang distribution can be identified by the presence of the __apple_build_version__ macro.
GCC & Intel C++ on OSX can be identified by the presence of the __APPLE__ && __MACH__ macro.

https://sourceforge.net/p/predef/wiki/browse_pages/

src/states_screens/credits.cpp
```diff
@@ -205,7 +206,11 @@ void CreditsScreen::loadedFromFile()
         // translations should be just before the last screen
         m_sections.swap( m_sections.size() - 1, m_sections.size() - 2 );
     }
-
+#ifdef __ANDROID__
```
auriamg: The special case for Android is ok, but there should also be a similar check to disable the button for unsupported platforms. LinkHelper::isSupported() or somethinG like that

This is the donate button, but same thing for the reset password button

qwertychouskie:
Thanks!

PR #3253, opened 2018-05-15
Fix #3236

### LINKED ISSUE #3236: Placement of the item hud.
I find the placement of item hud inconvinient, especially in Mulitplayer mode. Because the item hud is placed straight over the karts, it is difficult to see where to drive or shoot. It is an issue because you have to turn to see where other karts are, and then aim without seeing the target. My suggestion is to move the item hud to one of the corners. 

temple-2018 05 03_17 17 05

Alayan-stk-2:
Pictures showing how it looks like in splitscreen (no change in normal mode unless using an ultra-wide resolution) : 

item_splitscreen_1
item_splitscreen_2

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -57,10 +57,3 @@ before_build:
 build:
     parallel: true
     project: build\ALL_BUILD.vcxproj
-
-
-on_success:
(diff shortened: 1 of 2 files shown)


nado:
Protip : you can add [skip ci] in a commit message to avoid triggering a build instead of adding two commits in the history to do that ;)

Issue #3222, opened 2018-04-28
Rubber banding issues


Alayan-stk-2:
What I call rubber banding here is the slowdown of AI karts when they are too far ahead ; in easy, intermediate and expert. There is no such slowdown at all in SuperTux.

Here are the issues I noticed : 
- [ ] All the AI karts, from the first AI kart to the last, use the same distances to slow down. This results in all the AI karts being concentrated in a small space.
- [ ] The slowdown seems to be relative to the first player kart (I've not looked up the code to see how it works exactly). This means that in multiplayer, if one player do well, he makes the AI karts go faster and so the race harder for the other human player. So, to increase the chance of the other player to not be last, the better player must voluntarily slow down to have the AI do so as well. This is not enjoyable.

The first issue means that, as long as the player was able to drive as fast as the slowed down AI karts and keep close enough, doing well for the final 10-15% of the race can be enough to snatch victory even if the beginning was terrible (while a good beginning won't save the player at all). It also means that the margin between being first and being last can be too small : a small difference in skill can cause a disproportionate difference in ranking. Finishing mostly last or mostly first is not the most enjoyable experience.

For the first issue, my suggestion is to tweak the slowdown so that the last AI kart will slowdown earlier (possibly even at some distance behind the player's kart), while the first AI kart will wait to be slightly more ahead to slow ; with the other AI karts in-between.

For the second issue, I'm not sure about the best solution. Having it relative to the last player kart could make things trivial for the first player, and trying to have it relative to multiple player karts would risk being overcomplicated.

PR #3256, opened 2018-05-16
Make rubber-banding distances different between AI karts

Alayan-stk-2:
See #3222 for the reasoning behind this PR, which aims to fix both points.

I've also made the AI slow down before it even reaches the player in the easiest difficulty and increased the distance range over which the changes happens (for reference, math class has 500 of length, so the 30 to 60 range from full-speed to min-speed which existed before was definitely tiny)

The only slight issue I see with the change is that using the current AI position as a base may tend to make light karts (who have a start bonus) better ranked and heavy AIs worse ranked. Other possibilities exist, like using the kart's world ID, though it may things slightly easier.


src/karts/controller/skidding_ai.cpp
```diff
@@ -1965,23 +1972,69 @@ void SkiddingAI::computeNearestKarts()
...
-        unsigned int kart_id =
-            m_world->getPlayerKart(i)->getWorldKartId();
-        if(m_world->getOverallDistance(kart_id)>max_overall_distance)
-            max_overall_distance = m_world->getOverallDistance(kart_id);
+        unsigned int kart_id = m_world->getPlayerKart(i)->getWorldKartId();
+        overall_distance.push_back(m_world->getOverallDistance(kart_id));
+    }
+
+    // Sort the list (the list is very short, so a basic bubble sort suffice)
+    for(unsigned int i=0; i<n-1; i++)
```
auriamg: Even so, please don't reinvent the wheel. Just use one of the standard sorting methods

src/karts/controller/test_ai.cpp
```diff
@@ -315,7 +315,7 @@ void SkiddingAI::update(int ticks)
     computeNearestKarts();
 
     m_kart->setSlowdown(MaxSpeed::MS_DECREASE_AI,
-                        m_ai_properties->getSpeedCap(m_distance_to_player),
+                        m_ai_properties->getSpeedCap(m_distance_to_player, 2, 3),
```
auriamg: Why are 2,3 hardcoded?

Alayan-stk-2: It returns the average of the first speed cap and the last speed cap. I could have put 1,1 as well, though, as the test_ai is only used in profile_mode where there will be no slowdown anyway.

Plus, the test_ai is supposed to be updated with the new skidding_ai once changes have been validated (it has not been done recently, but it should be), at which point this will be overwritten by the changes seen in skidding_ai.

Basically, it doesn't matter and the main motivation behind changing anything at all in this file is avoiding a compilation error.

auriamg: A little comment explaining that would be welcome. something like 1,1 /* N/A, no rubberband in profile mode */);

PR #3267, opened 2018-05-24
Let user know when STK is compiled without libopenglrecorder

qwertychouskie:


deveee:
I'm fine with the idea, but:
- ideally single line should be <= 80 chars
- maybe just "Recording not available." (or ifdef whole recording menu?) is enough, because people that compile STK will get cmake warning anyway
- it's a kind of unsafe, because your menu item will execute handleContextMenuAction with cmd_id = -1. So maybe add ```if (cmd_id == -1) return;``` at the beginning

qwertychouskie:
@deveee I pushed a change that fixes the cmd_id issue by just adding the RECORDER_STOP back.  I also clarified the message (I think Ubuntu doesn't currently have `libopenglrecorder`), and also added a log message (useful if the person usually uses the keyboard shortcut).

Issue #3203, opened 2018-04-17
GP time issues when skipping a race


Alayan-stk-2:
When using the "skip race" option, the following issues appears in GP timing : 

1. If there are karts before the lap line, their time get estimated at the maximum possible (99:59.99)
2. The karts time estimation can be very weird during the first half-minute or so It would be better to use a prediction using potential speed with a random element rather than purely using the current average speed on track if a certain percentage of the total race distance has not been done.
3. The player who use the "skip race" option has no time added to his total GP time. i.e., it is effectively zero.

dumaosen:
It seems that the problem still exists.
![2019-04-19 19-55-04屏幕截图](https://user-images.githubusercontent.com/35170687/56453040-c6d5db80-636d-11e9-82bc-04929d206915.png)

PR #3276, opened 2018-05-29
Correct finish time for GPs when skipping a race

Alayan-stk-2:
Partial #3203 fix : the active human players don't have their race time counted as 0 if they skip a race ; instead the formula specified in endRaceEarly (but which wasn't applied !) is used.

Also, don't mark karts as eliminated in networked races, as endRaceEarly will be called not when the player gives up but when he doesn't finish within the allocated overtime after the 1st kart.


src/modes/standard_race.cpp
```diff
@@ -98,36 +98,55 @@ void StandardRace::endRaceEarly()
...
+
             // Have to do this to keep endSetKartPosition happy
             setKartPosition(kartid, kart->getPosition());
-            continue;
         }
 
-        if (kart->getController()->isPlayerController())
+        else if (kart->getController()->isPlayerController())
         {
+            if (estimateFinishTimeForKart(kart) > worse_finish_time)
```
auriamg: Don't call estimateFinishTimeForKart twice, call it once and store the result in a variable, no use doing the calculation twice

src/modes/standard_race.cpp
```diff
@@ -98,36 +98,55 @@ void StandardRace::endRaceEarly()
...
+                worse_finish_time = estimateFinishTimeForKart(kart);
+
             // Keep active players apart for now
             active_players.push_back(kartid);
         }
         else
         {
+            if (estimateFinishTimeForKart(kart) > worse_finish_time)
```
auriamg: same

Issue #2876, opened 2017-07-02
Use FPV cam in cutscenes (sometimes) causes segfault


qwertychouskie:
Seems to happen more with more complicated scenes (e.g. my new GPwin scene).

deveee:
Could you provide a backtrace from debugger?

qwertychouskie:
```
Thread 1 "supertuxkart" received signal SIGSEGV, Segmentation fault.
0x00007ffff569be48 in main_arena () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff569be48 in main_arena () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x0000555555ae0607 in CutsceneWorld::update (this=0x555558078820, 
    dt=0.0430000015)
    at /home/qwerty/stk-gpwin/stk-code/src/modes/cutscene_world.cpp:302
#2  0x0000555555af5e73 in World::updateWorld (this=0x555558078820, 
    dt=<optimized out>)
    at /home/qwerty/stk-gpwin/stk-code/src/modes/world.cpp:861
#3  0x0000555555addacc in MainLoop::run (this=0x5555572a3ee0)
    at /home/qwerty/stk-gpwin/stk-code/src/main_loop.cpp:248
#4  0x0000555555860832 in main (argc=<optimized out>, argv=<optimized out>)
    at /home/qwerty/stk-gpwin/stk-code/src/main.cpp:1780
```

auriamg:
That seems to be the cutscene camera interfering with the FPS camera.

For now, just don't use the FPS cam in cutscenes... though fixes welcome.

PR #3279, opened 2018-06-01
Fix segfault in cutscene when pressing CTL+Shift and when using FPV cam

qwertychouskie:
Fixes https://github.com/supertuxkart/stk-code/issues/2876.


src/modes/cutscene_world.cpp
```diff
@@ -297,28 +297,32 @@ void CutsceneWorld::update(int ticks)
...
-            matrix.rotateVect(up);
-            m_camera->setUpVector(up);
-
-            SFXManager::get()->positionListener(m_camera->getAbsolutePosition(),
-                                          m_camera->getTarget() -
-                                            m_camera->getAbsolutePosition(),
-                                            Vec3(0,1,0));
-
+            Camera *camera = Camera::getActiveCamera();
+            if (camera && camera->getType() != Camera::CM_TYPE_FPS)
```
auriamg: maybe checking getType() == CM_TYPE_NORMAL would be more future-proof? It's weird to single out FPS cameras like that

qwertychouskie: Fixed and force pushed.

PR #3282, opened 2018-06-06
Fedora deepedency installation

jpenguin:


deveee:
There is IMO a bit more work to do...

@hiker @auriamg @Benau Are you ok to move build instructions to INSTALL.md or something like this? Because it's already quite big...

I wouldn't add git/svn to dependencies, because you can download src package from sourceforge. And it's not listed in ubuntu dependencies, so we can keep consistency.

The pkg-config is needed, is it a part of development-tools?

IIRC on Fedora you can use system angelscript, so it can be added too. And maybe we should add enet and glew to both ubuntu and fedora.

Now we require openssl, so we should add it instead of gnutls.

jpenguin:
Good point on git/svn
angelscript? enet? I don't have either installed, I can dot network multiplayer & download addons.

auriamg:
@deveee Agreed with splitting into sub files, this is getting too big

nado:
@jpenguin the build script defaults to the bundled ones in case they are unspecified on the build command.

PR #3288, opened 2018-06-09
Ranking formula refinements

Alayan-stk-2:
Some refinements to time scaling for ranking points distribution. This acknowledges that 2 races of a total of N seconds are more random than a single N seconds race, and make points proportional to time^1.5 rather than time. In theory, this issue also affect the spread tendencies (the expected results), but this can't be accounted for without experimental data.


Benau:
STK website need updating too?

PR #3289, opened 2018-06-10
 Completely revised random powerup weights system (bis)

### LINKED ISSUE #1667: Finer-grained items weight list
The priority is moderate ; that's in background and is taken as a given by most players, but it has direct influence on global balance

Add finer-grained items weights list so weights will be adapted no matter the number of karts. Include adding a new position_class for finer-grained weights with high number of karts.
Should be flexible enough so the number of divisions and their range (for kart numbers, e.g. 16 to 20 karts)  would be 100% in powerup.xml, making thus future maintaining easier.

Actual values for powerup.xml will then have to be determined with help of several tests, making sure it works well on the various tracks.
Focus will be given to something working well on SuperTux difficulty, with the assumption it will work correctly for lower ones (some tests to confirm this will be needed). Adding per difficulty weights would be obviously overkill.

Alayan-stk-2:
This PR replaces #3216 - it was simpler resolving the conflicts on a new branch. It is otherwise virtually identical to the latest version (I only changed a few logs to be consistent with others which had been changed in the networking branch).

See the previous PR for the details of the content and of the goal.

Fixes #1667


Alayan-stk-2:
For reference, the current random item determination method in the PR needs 21 bits of entropy for the random number.

hiker:
Hi,

I've created a separate branch based on your pull request:
https://github.com/supertuxkart/stk-code/tree/STK-helper-ItemWeights
@Alayan-stk-2 , can you do a review and test? I only did some rather limited tests, including a 99 lap AI only race, and the collected items seems to correspond to the weights defined. Note that in powerup_manager.cpp, line 536 there is an ifdef that prints some basic statistics to the log file, which I then used to verify the distribution. Note that I noticed significant differences in the number of items - most karts had around 200 items collected, so I could (roughly the collected number of items with the weight), but some karts had under 150.

Feedback welcome!

hiker:
Oh, one additional comment: there is no hard-coded number of 'sections' for a certain rank anymore (but it must be the same number of sections for all number-of-karts in one game mode). So it should be possible to add one more entry for FTL: so we have entry 1 for the leader, and then entries 2-5 for the non-leader karts. Not that this is necessary of course, just pointing out that this is an option now. The code on the other hand now assumes that the weights are given in the right order (while previously we had w1, w2 etc iirc) - but I don't think that this is an issue, when tweaking the weights you would look at them in the right order ;)

Alayan-stk-2:
I've not yet done tests in race, this will take time.

The code changes were a bit long to review, but that's for the best. The new system with WeightsData is clear, and doing away with things like the positions enum is good.

That the number of sections is not fixed goes right with the purpose of this PR of making this more flexible. There is already entry 1 for the leader, and entries 2-5 for the non-leader karts, but I think you meant adding a 6th entry so that non-leader karts have 5 entries like in normal race mode.

Assuming that weights are given in the right order seems reasonable enough, and the comment in the xml file explains it, so that's fine I think.

The way the random item is determined now is much nicer than the previous one which took an unnecessary memory space.

I'd like to say this means that there is no cost to having a more exact interpolation, as it wouldn't affect the way the comparisons are done...

But I'm really worried about the way the random_number is determined.

I mean, I'm not a great fan of randomness, but with a list containing 200 items, at 12 "random number ticks" per second, it takes over 16s to loop over the list. This is way way over what would be exploitable, let alone noticeable. This would mean, for example "no basketball in the first 15s of the race is possible". If there is 25% of chance of having a bowling ball, this means a period of 4s where you will have a bowling ball. I did some tests by editing my powerup.xml on a recent version which already contains this random_item determination method, and I could tell what item I would get before picking an item simply by looking at the time I picked it up ! Sure, when it also depends on your current rank in the race it's a bit harder to know exactly what you will get, but not much really, knowing on which part of the distribution you will fall is already highly informative.

hiker:
As discussed in irc, this work was somewhat refactored in a separate branch (STK-helper-ItemWeights), and has been merged into master for now. So I am closing this pull request for now. Any additional work please create a new pull request off master.
Thanks a lot for this work!!!!!

PR #3291, opened 2018-06-11
Better checkboxes

### LINKED ISSUE #3235: The UI checkboxes are too similar between checked and unchecked states.
Probably a minor issue, but the checkboxes in the UI look too much alike when checked and unchecked : the background color remain the same, and the cross itself has a small contrast with the background.

This will mostly affect users with a poor eyesight ; but even with a good eyesight, you need to look close to the checkbox itself to get its state ; the difference is too small for the peripheral vision to get it when looking half-a-screen away or so.

Changing the background color when checked or having a more contrasting cross (contrasting color and luminosity) would fix the issue.

Alayan-stk-2:
Fixes #3235 

The "x" in the new checkboxes is both slightly thicker and much darker, creating a much better contrast.

checkboxes


Mrxx99:
Why is the checkbox for anti-aliasing bigger than the rest?

deveee:
It's much better than the previous checkboxes. But don't you think that now it's a bit too dark, or maybe too thick? At least "checked_focus" looks better for me than "checked". Or if it's already good for you, then I would just merge it.

Alayan-stk-2:
There is a trade-off between having crosses with a visible coloring and having them dark enough to have a good contrast.

The normal crosses are at around 25 luminosity (out of 100, varies slightly between colors) while for the focused boxes it is around 37 (because the surrounding color is also lighter).

For you and me, having it slightly lighter would probably still work well enough, but I'd lean towards having it that way for users with worse eyesight. The slightly better color perception from a lighter "x" seems less important. Same thing for the thickness, I'd put clarity first.

deveee:
ok

PR #3293, opened 2018-06-11
New dark skin (coal)

Alayan-stk-2:
This PR adds a new skin to STK.

The colors used are mostly dark grey, with white text.

coalskin1
coalskin2
coalskin3


auriamg:
@hiker @Arthur-D Any opinions?

Alayan-stk-2:
Video demo showing the theme : https://youtu.be/DzigpKvzUHU

auriamg:
I found a little issue, in splitscreen multiplayer, when selecting kart, the focus color should be the same for the player name spinner and the kart selection focus, but in this screen they are not

Alayan-stk-2:
The ruby skin also suffer from the same issue, although that's less blatant, as it's a pink focus and a deep red spinner.

I'll recolor the spinner so that it matches.

Also, I've been working on a fix for the news text not being readable on the too light bottom bar ; I'll do a separate PR for it soon.

PR #3307, opened 2018-06-15
Multi directional GUI navigation

### LINKED ISSUE #3295: Improve menu navigation with keyboard and gamepads
Currently, STK menus are exclusively navigated from element to element with the up-down buttons (the tab button also do the same thing as the down button), and inside an element with the left-right button.

This is non-intuitive and can make accessing to a desired element with a keyboard/gamepad rather hard.

Ideally, up/down/left/right would find the closest matching element or sub-element situated in the required direction and switch to it.

Roadblocks to clear : 
- determine the relative position of the widgets
- make sure it's not possible to be trapped within an element having some kind of scrolling

Alayan-stk-2:
Fix #3295 

This works by : 
1)Letting the widget handle it with left/up/right/downPressed. If the widget returns EVENT_BLOCK (default when not handled, but can be used when the widget detect the input should be used to navigate outside of it), then we go to step 2. EVENT_BLOCK_BUT_HANDLED has been added, as EVENT_LET would trigger an unwanted selection in list widgets.
2)Making sure there is a focused widget. If not, focus the first one of the screen and stop here.
3)Going through the list of widgets (with a dumb but efficient enough loop over ids ; simplicity avoids hard-to-find bugs) and finding the closest one depending on the requested direction. A closest widget to wrap to is also calculated and used if there is no widget in the requested direction (so "down" at the bottom sends back to the top, etc)

The heuristics used for finding the closest widget in the requested direction are rather simple, but work well in all the menus I tested (which is most of them, but there is a few I probably missed), and could be easily tweaked if found lacking in some future situation.

Known issue :
- In the addons screen, there are two spinners aligned horizontally. As they capture left/right input, it's currently impossible to go from one to the other, or to access the rightmost one. This can (and will) be fixed with a slight redesign of that UI screen, but I won't mix it with this PR.


src/guiengine/event_handler.cpp
```diff
@@ -427,172 +394,243 @@ const bool NAVIGATION_DEBUG = false;
...
+    // Used when there is no suitable widget in the requested direction
+    int closest_wrapping_widget_id = -1;
+    int wrapping_distance = 0;
+    int smallest_wrapping_distance = BIG_DISTANCE;
+
+    // In theory, it's better to look recursively in m_widgets
+    // In practice, this is much much simpler and work equally well
+    // Usual widget IDs begin at 100 and there is rarely more
+    // than a few dozen of them - but it's very cheap to have some margin,
+    for (int i=0;i<1000;i++)
```
auriamg: This is a pretty dirty hack

auriamg:
in the option menu, when going up, it skips the "vertical sync" checkbox by going directly from resolution to custom settings

in the options menu, "User interface" tab, going down, it goes straight from the user interface tab to the "always show login screen" checkbox, skipping the skin dropdown

in the players option screen, going down, it goes straight from "Remember password" to "Password", skipping "username"

in the players option screen, going up, it goes straight from "Password" to my user name, skinning the username textbox and both checkboxes

src/guiengine/event_handler.cpp
Alayan-stk-2: This is not great from the plan of theory and idealized conceptual computer science, where the time needed to write the perfectly elegant program is not factored in.

But from the engineering point-of-view, it does the job well, it is easier to maintain that some recursive-access widget selection, and it saves me the hours needed to find, write and debug the clumsy chain of OOP objects needed to get the data (the mixing of data and functions in OOP is very impractical).

Alayan-stk-2:
The players option screen behavior was due to the name field there being uneditable in some situation, without the UI making clear that modification is disabled.

The other observed issues have been fixed, with improvements in the behavior used to skip over invisible/disabled widgets.

auriamg:
One thing I noticed, that likely existed before, is that a focused list is visually identical to a non-focused list, so focusing an empty list especially is awkward

The GP edit screen and addons screens will need fixes due to placing things horizontally next to spinner (TBH the GP editor as a whole is really awkward to use with a gamepad/keyboard though)

The online network lobby is problematic, it's impossible to access the chat with keyboard only (see screenshot). This one is problematic, not too sure how to fix it...
menu-2018 06 17_18 14 55

Alayan-stk-2:
For empty lists being focused, I'm not sure how we should visually represent it. Something with the header, maybe, like applying the "down" visual property to all column headers at once ?

I think it would improve things for the online network lobby involves being able to recognize boxes, and use them as steps. In this example, it would identify the top-left box when going down, then recognize there is nothing selectable inside the box, then jumping to the box below with the chatbox. This ability to use boxes as guidance could be useful in other places (for example, in my addons screen tentative redesign, it would fix a difficulty I encountered).

However, the impossibility to access the chatbox does not stem from the algorithm but from the screen design.

The green/red buttons are inside a common ribbon which take over the left/right movement, as when it was done, this was the only way to have a correct left/right movement. But this now makes it impossible to get out of that ribbon by using left/right.

Removing the parent ribbon for these buttons and instead putting them in a div should make the chatbox easily accessible with left/right.

auriamg:
@Alayan-stk-2 I think it's ready to merge, of course further improvement to  the addons screen, for instance, are still welcome to improve their usability in this case

PR #3315, opened 2018-06-19
Optimize image size in the GUI folder

Alayan-stk-2:
I've optimized images in the GUI folder with pngquant, reducing its size by over 50% from ~3.1MB to ~1.5MB.

As to why not just using the optimize script, I'll quote what I wrote on the IRC :

> The optimize script is lossless. The optimization with pngquant converts images to an optimal 256-color palette (and it supports alpha). As icons only use a small subset of the available 32-bits colors, there is very little information loss and the visual perception is identical (I've compared a good number of images using varied colors, and it held true each time)


deveee:
Merged, thanks!

PR #3325, opened 2018-06-20
Navigation improvements

Alayan-stk-2:
This PR contains a few keyboard/gamepad navigation improvements : 

- Don't focus empty lists/empty ribbons (nothing indicates that the empty element is focused ant that's confusing)
- Ignore deactivated children of a ribbon when navigating in a ribbon.

### DIFF
--- a/src/guiengine/widgets/ribbon_widget.cpp
+++ b/src/guiengine/widgets/ribbon_widget.cpp
@@ -649,7 +649,7 @@ EventPropagation RibbonWidget::propagationType(const bool horizontally)
 
 /**
  * Move to the next child widget in the requested direction.
- * If it is inactive, move again, until it finds an active child or go over 100 loops
+ * If it is inactive, move again, until it finds an activated child or test all childs
  */
 void RibbonWidget::selectNextActiveWidget(const bool horizontally, const bool reverse,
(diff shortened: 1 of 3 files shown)


src/guiengine/widgets/ribbon_widget.cpp
```diff
@@ -666,6 +647,51 @@ EventPropagation RibbonWidget::propagationType(const bool horizontally)
...
+
+            bool select_zero = (m_event_handler && left) || (!m_event_handler && !left);
+
+            m_selection[playerID] = select_zero ? 0 : m_active_children.size()-1;
+        }
+
+        loop_counter++;
+        if (loop_counter > 100)
+        {
+            Log::warn("RibbonWidget", "Empty ribbon widget focused !");
```
auriamg: Can't you just detect that the ribbon is empty by checking for the number of children, instead of counting loop iterations?

Alayan-stk-2: The list of children still keeps deactived children afaict. That's used when a ribbon disable some children under some condition, but enable them in other (see the GP editor). Loop counting is somewhat necessary.

Though rather than an arbitrary 100, I think it would work with m_active_children.size() or something like that.

PR #3327, opened 2018-06-21
Better random items

Alayan-stk-2:
This PR fixes most of the remaining issues in the random item code :
- Add a simple PRNG whose outputs are statistically random for a few thousand items (more than enough for a race) ; and easy to sync as it depends only on race state and a seed (usual PRNGs update their state when generating a PRN, this one do not). Currently, its input seed is 0. It works decently with this, but ideally a unique seed per race (an usual 32-bit int) would be provided, as for the first item boxes for the first kart on a given track, the sequences are quite repeatable.
- Fix a nasty bug which inverted weights in WeightsData interpolation (e.g. : for 8 karts, with reference points at 5 and 9, the obtained value was 0,75x that of 5 and 0,25x that of 9, that is, the value expected for 6 karts)
- Guarantee that the sum of the weights in an interpolated WeightsData are the same as that of the two reference WeightsData (assuming they have the same sum)
- Guarantee that the sum of the precomputed weights per rank is identical
- Scale the sum of the precomputed weights to match the range of the random number outputed by the PRNG ; so that the beginning of the distribution has not a higher likelihood to get a matching random number than the end. This also makes rounding error in the precomputed weights per rank insignificant.
- Make the sum of weights in powerup.xml equal to 1000 per weight node instead of 200 (I simply multiplied by 5, I didn't try to tweak). It's as easy to read as %, if not more, and it reduces WeightsData interpolation rounding error, especially for rare items.

The documentation has only been partially updated.

### DIFF
--- a/data/powerup.xml
+++ b/data/powerup.xml
@@ -121,7 +121,7 @@
        a triple item rather than a single one. The probability to get
        an item is its weight divided by the sum of weights of all items
        (single AND multi). It is recommended to keep that sum equal
-       to 200 to easily keep track of probabilities.
+       to 1000 to easily keep track of probabilities.
 
        'Global' items which affect all karts (switch, parachute) should
        be quite rare, since otherwise the item might be used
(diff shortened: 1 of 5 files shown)


hiker:
*sigh* I just wrote a long message, which was based on a flawed testing method - there goes another day trying to figure out why I had too many rubber balls :P  Back to drawing board, I should have an update tomorrow.

hiker:
Hi,
as discussed on irc, I am not convinced by the recursion - the need for it, that it improves things. I tested your implementation with a simple PRNG (i.e. only one iteration). Both produced a good distribution of items (as in: very close to the distribution specified in the config file). Both had some 'long runs', i.e. sequences where the same item was picked over and over again. In my tests, I saw a run of picking the same item with a simple PRNG 4 times in a row once, everything else was less. With your implementation, I saw a run of length 10 twice, and in some cases (depending on item number) runs of length 8, 7, 6, ... That might not necessarily be a flaw, there is a certain probability to get the same item X times in a row, but considering that the recursive and simple PRNG give the same user experience, and the simple one seems to avoid longer runs better, I just implemented this bug fix with a simple PRNG.

I applied the bug fix you have found, and will also cherry pick the change to the item weights, but I also don't care if the interpolated weights are not exactly 200 (well, 1000 then), since we take the changed weight into account.

Thanks a lot for your work here, good catch of the issue with picking items, and nice solution!
So I don't see any b

Issue #3321, opened 2018-06-20
Allow running unit tests without assets


qwertychouskie:
The unit tests currently require the assets, though it probably should not.  See https://github.com/supertuxkart/stk-code/pull/3130.

hiker:
After some discussion here our preferred solution for unit testing:
1) Add a new file that controls all unit tests (i.e. move the corresponding function out of main.cpp)
2) Do a test as the second statement in main (i.e. after initialising CommandLine):
```
   if(CommandLine::has("--unit-testing")  ...
```
    i.e. without loading any assets.
3) modify all existing unit tests to work without assets. The tests that need a model/texture/... could initially be disabled, or we could add a second command line option (--unit-test-with-assets). Long term: we can include some very limited assets (i.e. a model and a texture or two) to a data/test directory for re-enabling more tests.

Considering the very very limited amount of testing, I set this as a very low priority, though it would still be great to get this (and more tests) in!!

hiker:
#3328 included unit testing by downloading the required assets. While we want to avoid this, I leave the reference here in case some code there would be useful.

Alayan-stk-2:
There are very few unit tests in the first place, and running them automatically on GitHub action is even lower priority.

PR #3328, opened 2018-06-21
Add unit testing to Travis - V3

qwertychouskie:
Only downloads needed assets.


auriamg:
in theory proper unit tests should be able to run without so many dependencies... this PR works I guess, but doesn't feel clean

hiker:
I've updated #3321 with some suggestion how this works. Downloading assets at build time is not something we want to do (esp. with SF svn being somewhat unreliable), so I am closing this pull request.

Issue #2767, opened 2017-02-04
Friction, power engine and parachutes


Alayan-stk-2:
Right now, the parachute speed dropping is mainly caused by the engine_power drop applied in kart.cpp.

Engine_power determines in part the speed, but it plays a bigger role even in the acceleration.

From what I have gathered, this explain a lot why parachutes can't drop that much the speed of a kart at a reasonably good speed ; while making any acceleration at a low speed nearly impossible.

The friction factor (is it meant to represent air friction ?) has close to no effect compared to the engine_power drop. This shouldn't be the case.

Alayan-stk-2:
I've worked on a fix. But the engine gears are getting on my way, as dropping below a target speed can change the engine power by a factor of 2 or 3 (!).

The kart properties on a whole are hilariously unbalanced.

PR #3331, opened 2018-06-23
Kart characteristics rebalance

### LINKED ISSUE #3188: Weight class of karts are (very) unbalanced
I suspected the issue from some time ; but after seeing closely how weirdly done are the kart_properties values ; I decided to run a test.

Putting 4 noloks, a gavroche (I used an older build to have the same test_ai, so it used my last player kart, but it's the same weight class as nolok) and 4 pidgins (heavy weight) against each other in time-trial races ; with the same AI for both.

I got +210 by doing so. The maximum possible in a race with those conditions is +10, and there was 42 races. Thats +5 per race. Those were only 4 laps race, which drags things towards 0 in comparison to longer tests.

The test confirms that medium karts are much much superior to heavies.

About specific tracks : all tracks where there is a risk of crashing/braking (and so a need to reaccelerate) had a much superior results for medium karts against heavies, but even on other tracks medium clearly dominated.

In normal races, where items force to reaccelerate often, the unbalance is probably even worse.

EDIT : new tests with noloks against wilbers (lightweight). Result : +252. Even worse. But in regular races with items, the better acceleration of lightweight karts at least partly compensate.

Alayan-stk-2:
This PR contains several related elements : 
- A fix for the way parachute work (see #2767). Previously, parachutes reduced acceleration, which was slow at reducing speed and made braking less effective (!) than when there is no parachute. Now, higher speed = higher slowdown. For this, a non-linear air friction was introduced. That element is currently in kart.cpp ; it probably should be moved somewhere in the physics, but as it uses some kart characteristics and I don't know that other part of code, I've left it as is for the moment.
- A small engine multiplier for nitro, so that using nitro to accelerate faster can be effective
- Many tweaks to kart characteristics in order to fix #3188 : 
   - Make acceleration differences between kart classes less stark, mainly by making medium/heavies accelerate faster.
   - Reduce the speed gap between light and mediums, while giving a higher top speed to heavies
   - Fix relative masses at 0.6/0.75/1 and make values logically related to mass have appropriate ratio
   - Reduce the max speed increase from slipstreaming, as it tended to be too strong
   - Make acceleration reduced less than max speed in lower difficulties
    - Introduce some turn radius differences between kart classes ~~(but considering the turn radius has some significant issue, this may need a new look once the related bugs are fixed and AI is made more smart about how to take curves)~~
    - Make more values relative to the base value, for easier maintenance
    - Other small tweaks

For reference, the test of 5 nolok against 4 pigdin, which gave a +210 previously (out of +420 possible) has now given a +71. The test of 5 nolok against 4 wilber, which gave a +252 previously has now given a -49. Those values are partially dependent on some AI limitations (lighter karts crush others in STK Enterprise for example, but for a good player medium or heavies are probably better), but playtesting directly, it feels rather balanced.

I also did a 5 pidgin against 4 wilber test : the wilbers won with +84, but there are also tracks (abysses, candela city, shifting sands) on which the pidgins won with a good margin (and the new mansion track would probably put the result closer to +70)

At that point, it's not easy to estimate what part of the difference is related to what the AI is better able to handle, and to what is really better ; before even factoring randomness in the tests.

Overall, I think it's much more balanced. However, I do expect some smaller scale tweaks in the future as the AI is improved, some bugs are fixed, and feedback is received.


qwertychouskie:
Is there any advantage to using a heavy kart now?

Alayan-stk-2:
While the parachute changes are not necessary for the others to work, the acceleration rebalance requires the air friction change to work best, and at this point switching the parachute behavior to it too was obvious.

Before responding to your concerns, I'd like to point out what's wrong with the way the parachute work without this PR : 
- It reduces engine power. This means that if you are already at a high speed, it has only a moderate effect in decreasing your speed. It can greatly be countered by getting skid bonuses, too.
- At a low speed, however, it kills your acceleration. It makes it more punishing at a low speed than at a high speed, as a general rule. Actually, that's why in 2017 I introduced a parameter to make the parachute lasts longer when getting it at a higher speed, but it was only a bandaid.
- A side effect of working by reducing engine power is that braking is made much LESS effective when a parachute is active. This doesn't make sense, and if you try to detach the parachute by slowing, you'll end up overshooting because braking will suddenly become much stronger once the parachute is dropped.
- The friction parameter in the config was actually useless

Now on to your concerns : 
- There is already a (linear) "air friction" effect, somewhere in the physics (I can't tell you where). I discovered it by doing tests while logging data. I put "air friction" in quotes, because it is more a sort of anti-inertia, slowing down proportionally to mass.
- When testing, this got in my way to get a good acceleration behavior. When smoothing out the huge (2-3x) jumps from one gear to another (5x jumps from first to last), I noticed that it was not possible to both get the kart accelerating fast enough at the start and not making it too quick to reach the max speed.
- The "air friction" I introduced (it's proportional to speed^1,5, instead of squared speed, thus the quotes) allows to get both a fast acceleration at the beginning and to not attain top speed too fast.
acceleration
- In turn, not having huge jumps in acceleration makes it smoother, and avoid to have a huge plateau effect with parachutes when dropping at a speed which triggers a stronger gear.
- The new parachute behavior allows to fix the existing issues I mentioned above. It can't be done just by tweaking the values of the config file. It can be done by using friction only for it, rather than having a general air friction (this was my first version, at which point I bumped into the gears change issues), however that last solution is overall inferior.
- The air friction doesn't remove the role of the max speed capping to determine the top speed according to difficulty and active bonuses ; it is meant to influence how quick that speed is reached. 
- Part of the reason to tie it all together in one PR is that many parameters interact with each other. Tweaks to the engine power boost given by various bonuses are balanced to work with the friction effect.
- On the networking side, as far as I can tell, all the necessary bits are already transmitted by the state (kart speed, what bonuses are active, etc). My guess would be that no changes are needed, though additional testing in networking would be required to validate that it works fine with rewind, obviously.

Alayan-stk-2:
The "long essay" is a 3-minutes read, which takes me much less work than splitting the PR in two. Acceleration, friction and parachutes are interlinked. Also, acceleration is linked with other kart characteristics to determine class balances. I submit that together because the changes are related and **splitting the code in 2+ PR, while technically possible, is useless** (and would break playtesting).

Also, to reiterate, there is no reason for this code to introduce networking issues, because all the bits relevant to compute the effective acceleration are exactly the same as before.

hiker:
I started to have a look at this, and while I initially had the strange feeling that something has changed with the steering, I think it was just ... the shaking of the train or so :)
I have the feeling that heavy karts have an advantage now, I had absolutely no change to catch a heavy kart in front of me with a medium kart in sandtrack - though that might of course be track dependent. It would be good if someone else could test with their favourite kart.

@Benau, can you check as well that there is nothing interfering with networking? As much as I saw it's only a modification once per frame depending on other values.

hiker:
Merged into dev branch, which means it will be merged into master next time we create android packages.

Benau:
Alayan seems the when if u release back button the air fricition cannot be applied so that the kart will keep going back!!

Alayan-stk-2:
The issue was that the linear slowdown compensation got the wrong sign for negative speeds. Fixed in the dev branch.

Closing this PR, as its content has been merged into the dev branch.

Benau:
you need this for battle mode

Alayan-stk-2:
For some reason, setAllBrakes seems to have no effect. I don't know why, looking at the code didn't give me any obvious answer. I don't see where the braking of the wheel would be applied, but I don't see it either in the current master, where it definitely has an effect.

My logging indicate that the code do go inside the if where the setAllBrakes command is called.

Benau:
...hiker is looking at it right now!

PR #3332, opened 2018-06-26
New icon

qwertychouskie:
Might be changed later if LCP's new icons are added, but for now, it's better.

Although there are a lot of icons, they only take ~512KB altogether, so space shouldn't be an issue.


hiker:
I don't see a reason to change the STK icon - the one we have is quite ... iconic (pun intended :) ). I am happy to add higher res version of our current icon, and more resolutions, but don't think we should replace our established icon.

qwertychouskie:
When it was discussed on the forum (https://forum.freegamedev.net/viewtopic.php?f=17&t=7968) most people were in favor is replacing the icon.  Also, from the SuperTux Discord server:

image

Apparently the new icon is pretty iconic too, and IMO it's much more visible at smaller sizes.  The old icon screams "0.8! 0.8! 0.8!", at high resolutions you can even see the text on the kart's spoiler, making it feel old.  The new icon is consistent with STK's visual identity seen in the GUI, while the old icon was not consistent in style with much of anything (even the render of the icon feels lifeless and flat, with no reflection/gloss/etc.).

Can we get some opinions from @auriamg , @Arthur-D , @Alayan-stk-2 , @Benau , @deveee , and @samuncle ?

Maybe what we could do is use the icon in the beta, then collect feedback and use/not use it based on the feedback from the public.  I should note that I have been using the icon for a long time (since ~June) and it looks quite natural.

Alayan-stk-2:
To restate my take on this : I don't like the current STK desktop icon which indeed "screams 0.8" and is not at the quality level of modern STK. But I don't really prefer the icon suggested in this PR. I prefer having a kart with Tux in the icon, and something with colors rather than just a grey thing.

PR #3336, opened 2018-07-01
Add macOS to travis builds

qwertychouskie:
For now, this just tests building.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@
 #
 sudo: false
 language: cpp
+os:
+  - linux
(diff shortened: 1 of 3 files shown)


auriamg:
The build is failing.

qwertychouskie:
The mistakes were fixed, though it seems to have found a real bug.  The server-only build is trying to link with OpenGL and failing.

As an added bonus, the GCC builds are faster now.

qwertychouskie:
Closing until the build issues are worked out.

PR #3357, opened 2018-08-02
Install tuxkart to external media by default

leggewie:
Tuxkart being a big game should be installed to external media by default in my opinion in line with recommendations from https://developer.android.com/guide/topics/data/install-location  Note, this change does not prevent the game to be installed to internal memory, it merely sets a sane default.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -3,7 +3,7 @@
           package="org.supertuxkart.stk_dbg"
           android:versionCode="1"
           android:versionName="git"
-          android:installLocation="auto">
+          android:installLocation="preferExternal">
 
     <!-- This .apk has no Java code itself, so set hasCode to false. -->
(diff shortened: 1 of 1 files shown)


deveee:
They suggest that large games should allow installation on external storage and we already allow it. Currently system chooses which location is better (i.e. depending on available free space). And anyway user can always move it to/from sdcard. I don't see any reason why we should prefer external storage.

leggewie:
I got this app via f-droid.  It was installed to internal memory despite internal memory being tight and despite Google Apps being installed to external memory by default.  I was then unable to move the app to the SD card due to insufficient memory (on the internal flash).  Not sure why it would need additional memory to move, but that's what it was.  In the end, I uninstalled the app.

Such a large app should install to the SD card by default, in my opinion.  The user can override the choice and move to internal memory if they so desire.  The sane default IMVHO is external memory.

PR #3364, opened 2018-08-04
Use a stable sort for replay list sorting

Alayan-stk-2:
This allows better sorting by preserving relative order when elements are equal according to the main sorting key.

For example, you can sort by time and then by track name, and you'll get replays grouped by track, and for each track they will be sorted by time.

This is also useful when using the compare ghost feature : if you sort by time, select a ghost, use the compare ghost option, then all the possibilities will be sorted by time.

The code change is trivial, as it's simply replacing std::sort by std::stable_sort

### DIFF
--- a/src/states_screens/ghost_replay_selection.cpp
+++ b/src/states_screens/ghost_replay_selection.cpp
@@ -51,6 +51,7 @@ void GhostReplaySelection::refresh(bool forced_update, bool update_columns)
 {
     if (ReplayPlay::get()->getNumReplayFile() == 0 || forced_update)
         ReplayPlay::get()->loadAllReplayFile();
+    defaultSort();
     loadList();
 
     // Allow to disable a comparison, but not to start one
@@ -136,7 +137,6 @@ void GhostReplaySelection::init()
  */
 void GhostReplaySelection::loadList()
 {
-    ReplayPlay::get()->sortReplay(m_sort_desc);
     m_replay_list_widget->clear();
 
     if (ReplayPlay::get()->getNumReplayFile() == 0)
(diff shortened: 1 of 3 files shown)


deveee:
The code change is indeed trivial, but for me it's not clear what is better. With std::sort I will always get expected result, because it will be the same every time I use it. So I can first sort by name, find what I needed, then sort by time, look at something, then back to sort by name and my replay is still at the same pos where it was before. Anyway I don't care much, but I will see what others think.

Alayan-stk-2:
If you sort according to a criteria with std::sort, you'll always get the same list, but besides the sorting key, the order is akin to a random wobble. When you have 20 or 30 replays matching one criteria or another (let's not talk about sorting by lap number, etc, as it can be much more), this mess is confusing, as you need to scan items one by one to find the one you're looking for, instead of a much faster lookup by successive criteria.

Aside from improving the default sorting order, an option would be to systematically do a multi-key sorting when the user request a sort by a given criteria.

e.g., say, the user ask for a sort by lap number.

The game would in the background sort by reverse mode, time, track name and only then by lap number. Of course, this would need to use stable_sort.

This should produce always the same result (time is more or less unique) without the extremely confusing pseudo-random distribution of list items with an equal value for the main sorting key that we have currently.

The downside of this solution is that the user would be stuck to a predefined order, instead of having the ability to tailor the selection for what he wants. Defaults need to be the most commonly useful combination, but sometimes somebody could need/prefer to have it differently.

auriamg:
As discussed on IRC, I think sorting on multiple fields like you are describing in your comment is the best solution. A stable sort is slightly better than a regular sort, but still risks being confusing to players who will have no idea how the sort works underneath

src/states_screens/ghost_replay_selection.cpp
```diff
@@ -457,13 +461,29 @@ void GhostReplaySelection::onColumnClicked(int column_id)
...
+
     loadList();
 }   // onColumnClicked
 
+// ----------------------------------------------------------------------------
+/** Apply the default sorting to the replay list
+ */
+void GhostReplaySelection::defaultSort()
+{
+    ReplayPlay::setSortOrder(ReplayPlay::SO_TIME);
```
auriamg: sorting multiple times in a row is not the optimal way to do it, though I guess the difference would be tiny

PR #3368, opened 2018-08-05
Fix list sorting

### LINKED ISSUE #3366: Sorting issues in the replay list
Besides that the current sorting is not stable (see my latest PR), there are two things which ought to be improved/fixed : 
- [x] When first opening the list, it is only sorted by track name (sorting by track name doesn't work well, see #3234). It would be better if on the first loading (and maybe when using the reload button), it was sorted according to several criteria, so the resulting list makes more sense and is more directly usable without having to sort manually. The best default order I think would be time -> laps -> reverse mode -> track name (main key). That would result in replays grouped by track, then by reverse or not, then by lap number, then by time.
- [x] Each time a column is clicked, the sorting order (ascending/descending) is reversed.

Example : you select time and it sorts in ascending. Then you select lap, it sorts in descending. You select lap again, it sorts in ascending. You select time again, now it sorts in descending.

The offending code is : 
`    m_sort_desc = !m_sort_desc;`
In ghost_replay_selection.cpp

To make the matter worse, **this directly contradict what the button in the list header is saying** : down triangle on first click on a given column (= expected ascending sort) ; up triangle on second click on the same column (= expected descending sort) ; column deselected on third click (= expected : no sort anymore, but currently a sort is still triggered).

Maybe the column deselection step could be removed, but the down/up button and sorting order need to work in sync with each other.

Alayan-stk-2:
Some sortable lists (ghost replay list, server selection) inverted sorting order on each click on a column, even if it was another one. This behavior didn't match the UI sorting indication (up/down triangle) and was weird and confusing.

Some others had custom code managing this.

With this PR, the main logic to decide to use default/ascending/descending sorting is done in the list widget, reusing the values computed for the UI indicator ; and all lists have been adapted accordingly.

The friends comparison is also made less redundant and a default sort to show the closest server by distance for the server selection screen has been added.

Fix #3366

### DIFF
--- a/src/guiengine/widgets/list_widget.cpp
+++ b/src/guiengine/widgets/list_widget.cpp
@@ -479,7 +479,7 @@ EventPropagation ListWidget::transmitEvent(Widget* w,
         m_header_elements[col].getIrrlichtElement<IGUIButton>()->setPressed(true);
         */
 
-        if (m_listener) m_listener->onColumnClicked(m_sort_col);
+        if (m_listener) m_listener->onColumnClicked(m_sort_col, m_sort_desc, m_sort_default);
 
         return EVENT_BLOCK;
     }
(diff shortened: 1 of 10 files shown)


auriamg:
Great improvement, thanks!

PR #3372, opened 2018-08-08
android/make.sh: Use bash

qwertychouskie:
A user on IRC reported this error: "make.sh: 8: export: Builds/STK_8-7-2018/stk-code/android: bad variable name"
It was fixed by running `bash make.sh`

### DIFF
--- a/android/make.sh
+++ b/android/make.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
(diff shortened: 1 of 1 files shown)


deveee:
Tbh. I prefer more portability instead of forcing bash shell. Maybe try to ask him what is the shell that he is using and maybe it's easy to fix. If it's just wrong export usage, then most of these exports before variables are not needed and can be removed.

PR #3373, opened 2018-08-08
Clearer spinner arrows

Alayan-stk-2:
Slight improvements on the ocean, ruby and coal theme to make the focused arrow in spinners more contrasting.

Especially, the coal arrow could give the opposite impression of what it was supposed to convey.


qwertychouskie:
@Arthur-D This only changes a few .png files, the build error must be unrelated.

deveee:
Cmake should disable wayland support if xkbcommon is not found. Maybe try with clean build?

PR #3374, opened 2018-08-08
Improved spinner

Alayan-stk-2:
Avoids the spinner blocking left/right navigation, which solves issues in some menus like the addons screen where you previously couldn't reach some part of the UI with the keyboard.


Arthur-D:
Seems to work pretty okay, though when hopping between arrows for the sound effects spinner I get a sound even if the volume didn't change at all. Not sure if it's desired behavior or not.

auriamg:
@Alayan-stk-2 I still think there's more for improvements to increase clarity, I will nonetheless merge this as it is a step in the right direction

PR #3385, opened 2018-08-20
Assign kart type to karts not having a valid one

### LINKED ISSUE #3375: Kart without a category should be assigned one by default
Some addons kart don't have a kart class in their files.
`[warn   ] KartProperties: Can't find kart type '' for kart 'Elephpant'`

Rather than flat-out disabling the kart, the game do a warn and use default values... But default values as a whole don't match those of any of the kart categories. For example, nitro use matches that of light kart, while mass is that of heavies, etc.

While fixing the addons would be nice too, the code should assign a category by default to the kart (probably medium).

Alayan-stk-2:
This uses the list of kart types in the XML file, without hardcoding valid types in code ; so it still works if someone change the names and/or numbers of kart types.

Consequently, the default type is "light" because it is the first in the list.

Fixes #3375

### DIFF
--- a/src/karts/kart_properties.cpp
+++ b/src/karts/kart_properties.cpp
@@ -355,13 +355,10 @@ void KartProperties::combineCharacteristics(PerPlayerDifficulty difficulty)
 
     // Try to get the kart type
     const AbstractCharacteristic *characteristic = kart_properties_manager->
-        getKartTypeCharacteristic(m_kart_type);
-    if (!characteristic)
-        Log::warn("KartProperties", "Can't find kart type '%s' for kart '%s'",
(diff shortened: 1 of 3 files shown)


qwertychouskie:
Is there any chance to set the default to Medium?  I think this would make more sense.

Alayan-stk-2:
The code doesn't assume how many kart classes there are, and it picks the first because it is the only one guaranteed to exist. It would be possible to pick the 2nd, but it would make an unsafe assumption (potential crash if the file has been modified).

The main goal is to not have impossible properties mixing different kart classes (and actually being overall best than any kart class) when there is none supported.

Afterwards, it's up to the addon's author to fix his kart by specifying a category.

PR #3391, opened 2018-08-22
Display time-trial and egg hunt times up to the ms

Alayan-stk-2:
The ghost replay screen already, and since a long time, display times with the precision to the ms.

It only makes sense to use the same precision for the modes available in ghost replay ; especially for time-trials, and considering that the bigger issues with the finish time accuracy have been fixed (though one known issue remains).

### DIFF
--- a/src/race/race_manager.hpp
+++ b/src/race/race_manager.hpp
@@ -724,6 +724,14 @@ public:
         return m_minor_mode == MINOR_MODE_TIME_TRIAL;
     }   //  isTimeTrialMode
     // ------------------------------------------------------------------------
+    int currentModeTimePrecision()
+    {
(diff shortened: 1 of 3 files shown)


src/race/race_manager.hpp
```diff
@@ -724,6 +724,14 @@ class RaceManager
         return m_minor_mode == MINOR_MODE_TIME_TRIAL;
     }   //  isTimeTrialMode
     // ------------------------------------------------------------------------
+    int currentModeTimePrecision()
```
auriamg: A little comment explaining it's a number of decimals would be welcome, otherwise those 2 - 3 numbers are a bit unclear

PR #3402, opened 2018-08-27
Fix egg hunt finish time

Alayan-stk-2:
getTime() is called immediately when the end condition (all egs collected) is detected, and the result is saved (and not changed afterwards : isRaceOver is called every frame even after it has return true) ; instead of calling it only later when the time is going to be displayed.

The time displayed at the end of the race now matches the ghost's finish time.

### DIFF
--- a/src/modes/easter_egg_hunt.cpp
+++ b/src/modes/easter_egg_hunt.cpp
@@ -33,6 +33,7 @@ EasterEggHunt::EasterEggHunt() : LinearWorld()
     m_use_highscores = true;
     m_eggs_found     = 0;
     m_only_ghosts    = false;
+    m_finish_time    = 0;
 }   // EasterEggHunt
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


deveee:
Tbh. It looks like an ugly hack and it doesn't do what the function names suggest. The isRaceOver() should just check if race is over and return true/false. And that's why you need to hack it with ```if (m_finish_time == 0)```

I suppose that there is no better fix? For example by moving estimateFinishTimeForKart() so that it's called immediately after race finish? If not, then we can merge it.

Or you can use ```if (m_eggs_found == m_number_of_eggs) m_finish_time = getTime();``` in collectedItem() which feels a bit less confusing for me and you will get a time when last egg was collected.

Alayan-stk-2:
Whatever way to do it needs, on the first time isRaceOver returns true, to save the current time, and to not save it again afterwards.

There are a few possible ways to achieve this : 
- insert the code in isRaceOver as I did there ; the check is already there, it's a matter of changing what is done when the if is true.
- add additional checks in collectedItem as you suggest : check for the end condition too, but also check about ghost kart status. There is some duplication compared to the first solution, though it's rather negligible.
- Add a call to isRaceOver inside easter_egg_hunt.cpp and depending on the result save finish time
- Next to the existing call ; if isRaceOver sends true, do a cast to egg hunt world, and if the cast is successful, call a function which saves the finish time.

I don't think these last possibilities are really superior.

Moving when estimateFinishTimeForKart() is called is maybe possible, I've not investigated in detail.

auriamg:
I understand deveee's concern, though IMO the fix as proposed is good enough

PR #3426, opened 2018-09-09
Add dirt to karts

samuncle:
https://forum.freegamedev.net/viewtopic.php?f=17&t=8086&p=77880#p77880

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -377,6 +377,8 @@ void Kart::reset()
     m_view_blocked_by_plunger = 0;
     m_has_caught_nolok_bubblegum = false;
     m_is_jumping           = false;
+    m_dirt_factor          = 0.0f;
+    m_dirt_reset_counter   = 0.0f;
 
     for (int i=0;i<m_xyz_history_size;i++)
     {
@@ -1165,6 +1167,14 @@ bool Kart::isOnGround() const
 }   // isOnGround
 
 //-----------------------------------------------------------------------------
+/** The kart as a dirt factor (if the player skids and go in dirty places)
+ */
+ float Kart::getDirtFactor() const
+ {
+     return m_dirt_factor;
+ }   // getDirtFactor
+
+//-----------------------------------------------------------------------------
 /** The kart is near the ground, but not necessarily on it (small jumps). This
  *  is used to determine when to stop flying.
 */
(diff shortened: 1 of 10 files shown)


Mrxx99:
Looks great, but your link doesn't work, I guess you meant this: https://forum.freegamedev.net/viewtopic.php?f=17&t=8086

deveee:
Looking at the screenshot, I'm fine with the idea. Maybe add it as an option and enable only on high settings, because it doesn't affect gameplay. And as Alayan said it only has a sense on some kind of terrains. It would look strange during skidding on an asphalt road.

And please don't multiply foat * integer, eg. ```vec3 dirt = texture(g_dirt_albedo, uv * 4).rgb;``` because GLES doesn't allow it and I will have to fix it anyway.

qwertychouskie:
Agreed, should be an option in the settings.

src/karts/kart.cpp
```diff
@@ -1890,6 +1900,29 @@ void Kart::handleMaterialGFX(float dt)
 {
     const Material *material = getMaterial();
 
+    if (m_skidding->isSkidding())
+    {
+        m_dirt_factor += 0.001f;
```
Alayan-stk-2: This should depend on the dt this function is called with, so that the time to dirty is frame-rate independent.

src/karts/kart.cpp
```diff
@@ -1890,6 +1900,29 @@ void Kart::handleMaterialGFX(float dt)
...
+        if (m_dirt_factor > 1.0f)
+        {
+            m_dirt_factor = 1.0f;
+        }
+        m_dirt_reset_counter = 2.0f;
+    }
+    else
+    {
+        m_dirt_reset_counter -= 0.01f;
```
Alayan-stk-2: This should also depend on the dt.

src/karts/kart.cpp
```diff
@@ -1890,6 +1900,29 @@ void Kart::handleMaterialGFX(float dt)
...
+        if (m_dirt_reset_counter < 0.0f)
+        {
+            m_dirt_reset_counter = 0.0f;
+            m_dirt_factor -= 0.002f;
```
Alayan-stk-2: This too should depend on the dt.

Besides, apart from waiting ~6s without dirtying the kart, maybe the cleaning of the kart could depend on some other things : 
- Speed
- Rescue

src/karts/kart.cpp
```diff
@@ -1890,6 +1900,29 @@ void Kart::handleMaterialGFX(float dt)
 {
     const Material *material = getMaterial();
 
+    if (m_skidding->isSkidding())
```
Alayan-stk-2: As discussed previously, this is not a good enough test for merging in master, as it is weird on many tracks.

The test should be something like this : 
`if (isOnGround() && material->isDirty())`

And you would then use skidding only to compute the rate of dirtying (it would dirty the kart faster than normal driving), speed could also be used as a factor (for reminder, 25 = normal supertux speed).

Alayan-stk-2:
No update in two years, closing.

If someone wants to take from samuncle's WIP and make it into something production-ready, it would be welcome as a new PR.

PR #3431, opened 2018-09-15
Allow to disable a keyboard config without deleting it

### LINKED ISSUE #3241: Add the possibility to disable a control configuration without deleting it
I'm not sure if this issue is relevant outside keyboards, but it at least exist for them.

It's currently not possible to have a configuration created but inactive. However, it can be handy to have stored configurations you only use in some situations.

The most obvious situation is having one single player configuration, and two/three multiplayer configurations (multiplayer with keyboards is actually quite playable with 2 or 3 keyboards to avoid the "keyboard are evil" issues). As those other configurations also get keys for moving in menus (selecting kart, etc) ; keeping them around means disabling those keys for normal inputs (i.e. : entering some name needing one of the affected keys). But deleting them means having to recreate them the next time they are needed, which is frustrating (especially as there is often some key which is mapped wrong when doing so, needing to check again, change again...)

The option to disable a configuration, but keeping it in memory for easy reuse by simple enabling, gives the best of both worlds.

Alayan-stk-2:
Adds the possibility to disable a keyboard config, and an icon to display the on/off status of a keyboard config. Make sure there is always at least one enabled keyboard config.

Note that even if the default config is disabled, the default arrow keys still keep working for navigation, this is not an issue in this PR, as this behaviour exists in the current master too.

Benau tested with a gamepad and noticed no issue with disabling/enabling it, so there should be no regression.

Fix #3241


deveee:
@Alayan-stk-2 I will look tomorrow.

auriamg:
I'll let deveee review too since he was requested, just mentionning that I tool a look and this looks good to me

deveee:
Ok, merged. Actually we could allow to disable android gamepad too, but there is a bit more work to do (at least add a way to restore default keys), so maybe keep it as it is for now.

PR #3465, opened 2018-09-23
Race UI improvements

### LINKED ISSUE #1192: Too much bots for toplist
**Author: usul1**

Currently STK allows 18 opponents. Sadly this number is to big and during the race the opponent list at the OSD does only dislay the very first players, but not all that take part at the current race. But after the race, the toplist scales well.

I use STK at 1376x786px. 

Migrated-From: https://sourceforge.net/apps/trac/supertuxkart/ticket/1136

Alayan-stk-2:
This PR includes : 
- A new lap flag icon next to the number of laps
- An option to control the minimap display to make everybody happy. Tastes vary, and one preset can't satisfy everybody. The options are bottom-left corner (same as current UI, and default option), right side (between the speed indicator and the lap indicator) and hidden. For Android, the bottom-left corner option is disabled.
- As the smallest supported resolutions are 1024x768/1280x720 (instead of 800x600 before), some changes to the icon scaling in the left of the screen. If the minimap is hidden or is displayed on the right side of the screen, it can display up to 20 icons, so this fixes #1192 
- Misc. fixes (reducing the lap indicator font size was useless ; the icons on the left could go over the minimap as the minimap's margin with the bottom of the screen wasn't accounted for)

This PR do not include moving/changing the rank indicator, despite it being the main focus of the IRC discussions yesterday, as I prefer to not mix too much things at once.

I tested with multiple resolutions, so hopefully it should work fine across them.

A few screenshots in 1360x768

With the minimap hidden and 8 karts : 
race_uic
With the minimap on the right side and 20 karts : 
race_uib
The options menu : 
race_uia


deveee:
If you want this feature, then I don't really care.

Maybe don't use #ifdef ANDROID in this case, because:
- it's possible to play STK on Android TV with keyboard/gamepad and without touch gui
- maybe in future I will add support for touch device on wayland (easy to do, but I don't have hardware to test)

I would just add something like 
```
if (m_multitouch_gui != NULL && UserConfigParams::m_minimap_display == 0)
{
    UserConfigParams::m_minimap_display = 1;
}
```
in race_gui and allow all values in settings.

And btw. is there a reason to use custom height for checkboxes? Because now they are smaller than on other screens. It's ```height="6%"``` vs ```height="fit"``` in checkbox div.

Alayan-stk-2:
I think it would be confusing to have the setting be "bottom-left" and then see it in the right of the screen in the race, it would seem buggy unless the text was expanded to explain this limitation (which wouldn't really fit).

But point taken about the possibility of using gamepads/keyboards with android, I'll do the test with m_multitouch_gui instead of the ifdef.

deveee:
Now it will fail if you disable touch device -> set minimap on left -> enable touch device :P So it still needs a check in race_gui.

deveee:
And btw. another off-topic in addition to small checkboxes, because I noticed some discussions in IRC logs. I don't expect that someone will create ranked server on android device... It just mostly for players that want to connect their two phones/tablets and play a multiplayer lan game.

Alayan-stk-2:
> Now it will fail if you disable touch device -> set minimap on left -> enable touch device :P So it still needs a check in race_gui.

Actually, if m_multitouch_gui is enabled, it will automatically use the right side position, because I didn't change your specific code checking for this.

There is indeed a _tiny_ confusion potential remaining there, but I wouldn't worry too much about it, unless there is an elegant solution for it.

I'll have a look at those checkboxes.

Mrxx99:
If you need to get anything tested on Android TV just mention me.

Mrxx99:
And btw. maybe you could try changing the lap icon a bit making the black part white and the white part transparent. I think that should fit much better the current UI

Alayan-stk-2:
@Mrxx99 Racing checkered flags are black & white ; having it white and transparent wouldn't be pretty and would lose the recognizability value of this icon.

src/states_screens/options/options_screen_ui.cpp
```diff
@@ -108,6 +108,20 @@ void OptionsScreenUI::loadedFromFile()
     skinSelector->m_properties[GUIEngine::PROP_MAX_VALUE] = StringUtils::toString(skin_count-1);
 
 
+    // Setup the minimap options spinner
+    GUIEngine::SpinnerWidget* minimap_options = getWidget<GUIEngine::SpinnerWidget>("minimap");
+    assert( minimap_options != NULL );
+
+    minimap_options->m_properties[PROP_WRAP_AROUND] = "true";
+    minimap_options->clearLabels();
+    minimap_options->addLabel( core::stringw(_("In the bottom-left")));
```
auriamg: translator cues would be very useful here, otherwise the context is not clear

src/states_screens/race_gui.cpp
```diff
@@ -1158,13 +1185,30 @@ void RaceGUI::drawLap(const AbstractKart* kart,
     // don't display 'lap 0/..' at the start of a race
     if (lap < 0 ) return;
 
+    // Display lap flag
+    irr::video::ITexture *lap_flag= irr_driver->getTexture(FileManager::GUI_ICON,
```
auriamg: Calling getTexture repeatedly on each render is a bad practice, this will do a string lookup each time. BEst keep a reference to the texture

PR #3467, opened 2018-09-24
Fix the turn radius bug

Alayan-stk-2:
This PR makes the turn radius of all karts independent of their length. It now only depends on the value defined in kart_characteristics.xml.

I have deliberately not included a rework of the values in kart_characteristics.xml ; but for information, they are close to the tighter turn radius observed before.

The turn radius at 25 of speed is now ~15 ; while the previous values for a few karts were : 
- Wilber : ~14,88
- Tux : ~16,15
- Emule : ~18,65
- Adiumy : ~22,4


hiker:
No idea what I measured that I said it does not work with skidding, I started again, and now it seems to work just fine.

Issue #3246, opened 2018-05-11
The basket ball sound is heard too late


Alayan-stk-2:
This is related to the way positional sounds are treated and to the ball's default volume.

When a basketball is coming, you're supposed to hear it soon enough to react and use a bubble gum, if you have one (or to try to destroy it with a bowling ball).

However, even when having the music sound lowered (which isn't very pleasant but help to better hear other sounds), you often hear it too late, even when driving at around 2/3rd of the ball speed (which increases the time you should hear it). Needless to say, it is extremely frustrating to have the means to defend yourself but to not be able too because of this sound issue.

In addition to increasing the sound's volume (or making its decrease due to distance slower), other things which could resolve that issue (and help in similar situations with cakes) would be to : a) have an icon on the screen border alerting to the incoming item ; b)have the option to look backwards without using the "backward" button, with a kind of small "rear-view mirror" in a screen corner (would probably be too small in splitscreen, however).

auriamg:
I see what you're getting at, but

* We usually can't make sounds louder than their current maximum without distortion
* Increasing the range might help, but it would become annoying, you would constantly hear bouncing when racing
* A rear view in the corner would require rendering the scene once more, that's too heavy for many people with lower-end systems. I don't think it would look good anyway.

Maybe an icon on the minimap is the best/simplest wayI can think of...

Alayan-stk-2:
If a rear-view (I took the idea from a game which had one where the items are in STK) would require a second rendering, that's indeed not an option.

I'm not sure if increasing the range would do that the ball is constantly heard when racing, if it's a moderate increase this may not be the case. Though considering that tracks are loops, the direct distance between two points is often much less than the distance when driving on-track, so I may underestimate this.

For the icon, while it may work on the increased minimap, it would be more obvious a la MK, in a square indicating the rough position the dangerous item is coming from.

PR #3472, opened 2018-09-26
Move the rank indicator to the top-left

Alayan-stk-2:
It is now in the "position/total number of karts" format, and sit close to the icon list to clarify its rank function.

It works fine in single-player or 2-players splitscreen (if horizontal splitscreen) ; but due to the overabundance of UI elements in 3-4 players splitscreen, this PR disables the icon lists.

tl_rank_3
tl_rank_2
tl_rank_1

### DIFF
--- a/src/states_screens/race_gui.cpp
+++ b/src/states_screens/race_gui.cpp
@@ -285,11 +285,14 @@ void RaceGUI::renderGlobal(float dt)
         }
     }
 
-    if (!m_is_tutorial)
+    if (!m_is_tutorial &&
+        (race_manager->getNumLocalPlayers() == 1 ||
+        (race_manager->getNumLocalPlayers() == 2 &&
+         UserConfigParams::split_screen_horizontally == true)))
     {
         if(UserConfigParams::m_minimap_display == 0 || /*map in the bottom-left*/
            (UserConfigParams::m_minimap_display == 1 &&
-            race_manager->getNumLocalPlayers() >= 2))
+            race_manager->getNumLocalPlayers() == 2))
             drawGlobalPlayerIcons(m_map_height + m_map_bottom);
         else // map hidden or on the right side
             drawGlobalPlayerIcons(0);
@@ -334,7 +337,8 @@ void RaceGUI::renderPlayerView(const Camera *camera, float dt)
     if (m_multitouch_gui == NULL)
     {
         drawPowerupIcons(kart, viewport, scaling);
-        drawSpeedEnergyRank(kart, viewport, scaling, dt);
+        drawSpeedEnergy(kart, viewport, scaling, dt);
+        drawRank(kart, viewport, scaling, dt);
     }
 
     if (!m_is_tutorial)
@@ -653,7 +657,7 @@ void RaceGUI::drawGlobalMiniMap()
 
 //-----------------------------------------------------------------------------
 /** Energy meter that gets filled with nitro. This function is called from
- *  drawSpeedEnergyRank, which defines the correct position of the energy
+ *  drawSpeedEnergy, which defines the correct position of the energy
  *  meter.
  *  \param x X position of the meter.
  *  \param y Y position of the meter.
@@ -771,7 +775,7 @@ void RaceGUI::drawEnergyMeter(int x, int y, const AbstractKart *kart,
 }   // drawEnergyMeter
 
 //-----------------------------------------------------------------------------
-/** Draws the rank of a player.
+/** Draws misc. info inside the meters circle.
  *  \param kart The kart of the player.
  *  \param offset Offset of top left corner for this display (for splitscreen).
  *  \param min_ratio Scaling of the screen (for splitscreen).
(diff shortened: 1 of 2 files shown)


qwertychouskie:
I'm not convinced.  What would it look like on top of the rank list rather than to the side of it?  Also the speedometer now seems rather empty.  (Maybe re-add a speed number?  Hmm...)

Alayan-stk-2:
With the rank on top-left : 
rank top-left

The "X first" overflows a bit, so icons need to be slightly lower. That takes out the space for around 2 icons.

For the empty disk inside the meters, the idea of some alert icon has been floated, but it's still only an idea.

Alayan-stk-2:
Here's a video showing a test for an alert icon : https://www.youtube.com/watch?v=nKNEjUTsxDg

This is not definitive, so feedback is welcome.

deveee:
For me it's too messy and there is too much duplicated information. I have a lot of kart icons on left side, so why do I need that top-left indicator? Maybe just highlight human player somehow instead... And you can see your/other karts positions on the minimap too. And for the bottom-right indicator you can sometimes use "look backwards" button instead. But anyway if you want it, then I don't care that much.

qwertychouskie:
I like the basketball indicator, but I'm not a big fan of it also showing karts (seems too busy).

Alayan-stk-2:
@deveee The human player is already higlighted in the icon lists : its icon is a bit bigger, and it is surrounded by a semi-transparent bubble. While this already gives the position information (if all icons fit, which they may not with bottom-left minimap + a lot of karts), this requires too much attention when concentrating on driving ; so the rank indicator remains needed.

For the bottom-right indicator, it's a bit of the same : while it is useful at times to look back to see what's going on, some tracks segments require precise driving making a lookback quite risky.

However, it's clear that the indicator showing the closest kart icon too is disliked, so I'll remove it, and I'll see if I can make the indicator better (more discrete) when there is no danger.

hiker:
I agree with Deve, and don't see a good reason for this additional indicator.

Alayan-stk-2:
To recap : 
- The current position indicator is very confusing for new players because of its location and because it is only a number (e.g. ; "7" instead of "7th" or "7/12").
- This is the only thing I'm adamant in changing, because it's poor UI. It generated a lot of confusion at the time of the change, and there is no doubt that even though new users are much less vocal than a regular discovering a change, it continues to do so.
- "7th" is impossible because of translation issues and because the digit font doesn't support letters
- Place is limited inside the meter circle, and "7/12" doesn't fit in one line (even more for "11/12"). My initial proposal, with "7" on top and "/12" below was rejected by Auria as too crowded.
- Several tries were made to find a better position for the rank indicator. Currently, it's the top-left (to associate it with the icon lists), though if a suitable icon is found to differentiate it from the lap indicator, the top-right below the lap indicator is probably better.
- Moving the rank indicator out of the meter resulted in Qwerty and Mrxx99 finding the meters circle empty, which I think was a rather fair criticism.
- To patch the "circles are empty" issue, the most obvious path was adding something inside (a full meter redesign to have them inside each other instead would be a lot of work). Because of #3246 and following a suggestion by Arthur, I tried to add this "alert indicator" to fix both issues at once.

Obviously, neither the rank indicator nor the alert indicator are satisfying enough at the moment considering the feedback received.

**But while telling me one's dislike of a proposal is fine, the question is : "What would be the best place to fit a complete rank indicator ("A/B" and not just "A"), and how to solve the secondary issues this may create ?".**

Because if to patch issue A perceived by one person, I have to create issue B perceived by another one, it will never end. Note that I don't mind adding one or two config options if that's required to make everybody happy ; however I think this should be more of a last resort once the possibilities themselves are as good as possible and it's down to personal taste ; not as in "the feature is half-baked, let's give an opt-out".

Alayan-stk-2:
Here is a new iteration which tries to address most critiques so far : https://www.youtube.com/watch?v=2yKQ3kAfNUc

Changes compared to the previous one : 
- Rank indicator moved under the lap indicator ; with a greyscale podium icon. The text's red shadow has not been removed, so it stands out a bit as different, but that's a small detail which can be changed. (As a note : cddepppp256 suggests making the max size of the current position equal to the size of the total number of karts.)
- Make the background of the alert icons semi-transparent to stand out less when there is nothing inside
- Display icons of other karts only if they have a bomb or swatter and are close. This makes it significantly rarer to have an icon pop up there, so when it does for more serious alerts it stands out more, while it reduces "UI overload" in other situations.

auriamg:
I like the podium icon!. good idea. probably it could have a little more spacing between this one and the other icon above it

As for the warning icon, it's interesting, but when it displays nearby karts IMO it's a little too busy/distracting and not so helpful (what can you do with the info that there is a nearby kart?).

Alayan-stk-2:
Different languages use different rules, unless you ask to translate _every single possible position_, your method doesn't work. This was already discussed (see IRC logs).

As for the warning icon : 
- Considering this is a feature I've coded without even being sure it will be included, it is obviously still a bit rough. Threats in front are relevant, alerting about them isn't really problematic even if you could see it without the icon ; except I think for the basket ball as it isn't a threat at all when ahead.
- If put outside the meters circle, a single fixed location wouldn't really do, but positional computations to determine the x position to display it aren't something I'm willing to code, at least in the foreseeable future.

qwertychouskie:
It would be 20 strings to translate, which sounds like a lot, but when each "string" is just 1 word, it's not much at all.  For reference that sentence was 25 words :)

hiker:
Considering that most of the solutions (including the one we are currently using) is a work around the translation problem, I'd suggest we actually look at solving it properly:
1) In some languages simply adding a "." at the end will be enough (certainly for German, and according to Benau for Chinese as well).
2) Some languages need a subscript (English, French apparently) - maybe also superscript? So we would need to support this.

We should have more flexibility with the fonts nowadays - what is stopping us from making this work? We would need to translate 20 terms, and could use _ and ^ to indicate sub/super-script, e.g.:
1^st. I assume we can get the extend/position of the number string, that should allow us to compute the super/subscript (or probably better): right justify the number, then the super/subscript is actually fixed.

IIrc in the past @Arthur-D added some missing characters to a font, so very worst case we could just add some characters to the font (e.g. for English st, nd, ...) - or maybe just find a font.

Re the warning display, let's discuss this in a separate issue.

Alayan-stk-2:
Okay.

I'll try to separate the changes in my branch between those who can be merged right away (like making the drawing of one kart icon with the check for attachments, etc into a separate function), and those concerning the current proposals for the warning display and the rank.

auriamg:
Another attempt for brainstorming, no idea if it will be any popular ;)

stk3

Issue #3462, opened 2018-09-22
The lapline in the Old Mine in reverse can be incorrectly triggered


Alayan-stk-2:
See screenshot : 
![oldmine_lap](https://user-images.githubusercontent.com/25536748/45912484-7ec7f800-be22-11e8-8281-773fb4c728cb.png)

Alayan-stk-2:
The core issue is not in the blend ; it's in the lapline validation accepting interpolated quads for the kart's position.

PR #3475, opened 2018-09-27
Don't have the distance frozen when going off-road/in cannons

Alayan-stk-2:
This fixes :
- Kart being incorrectly considered first when crossing the lapline off-road
- Distance freezing when going outside of the drivequads, which can happen with normal driving in many places. This had a slightly bad effect on kart positions, and a very bad effects on the live difference feature.
- Distance freezing in cannons (but it only works well if the kart is considered ~on-road, the interpolation tends to fail badly ; for example it works fine in Cornfield Crossing if I go in the cannon slowly, but if I go fast entering it, it will do a very poor interpolation midway in the cannon). I suspect that a full fix may require alterations of the tracks with cannons where it doesn't work well.

The rescue node is kept at the last drivegraph where the kart was on-road, so e.g. rescuing in the middle of a shortcut (invalid or not) can never allow to win time.

An important reason I'm not merging this right away is that it requires some change in the network code.

m_estimated_valid_track_coords do not need to be fully sent ; only the Z coord is ever used.

However, several optimizations are clearly possible in the existing state synching for TrackSector : 
- The graph nodes are sent as INT32 ; however there is not even one track remotely approaching the limits of an INT16, as far as I can tell.
- The last triggered checkline is sent as an INT32 ; an INT8 would certainly suffice as no track has more than a few checklines
- The on-road status could even be merged with the last triggered checkline (one bit used for it, the 7 others for the last triggered checkline)
- It does not seem useful to send the full track_coords variables. For the current coord ; the Y variable seems unused (the Z is used for distances without checklines check, the X for distance to center of track). For last_valid one ; only the Z coord seems ever used.

So while the new values add 48 bits to synch, there is up to 160 bits to save on the existing state ; i.e. there should be no issue for networking to integrate the new values in the game.

### DIFF
--- a/src/modes/linear_world.cpp
+++ b/src/modes/linear_world.cpp
@@ -24,6 +24,7 @@
 #include "audio/sfx_manager.hpp"
 #include "config/user_config.hpp"
 #include "karts/abstract_kart.hpp"
+#include "karts/cannon_animation.hpp"
 #include "karts/controller/controller.hpp"
 #include "karts/ghost_kart.hpp"
 #include "karts/kart_properties.hpp"
@@ -240,7 +241,9 @@ void LinearWorld::updateTrackSectors()
 
         // Nothing to do for karts that are currently being
         // rescued or eliminated
-        if(kart->getKartAnimation()) continue;
+        if(kart->getKartAnimation() &&
+           !dynamic_cast<CannonAnimation*>(kart->getKartAnimation()))
+            continue;
         // If the kart is off road, and 'flying' over a reset plane
         // don't adjust the distance of the kart, to avoid a jump
         // in the position of the kart (e.g. while falling the kart
--- a/src/tracks/track_sector.cpp
+++ b/src/tracks/track_sector.cpp
@@ -39,10 +39,11 @@ TrackSector::TrackSector()
 // ----------------------------------------------------------------------------
 void TrackSector::reset()
 {
-    m_current_graph_node       = Graph::UNKNOWN_SECTOR;
-    m_last_valid_graph_node    = Graph::UNKNOWN_SECTOR;
-    m_on_road                  = false;
-    m_last_triggered_checkline = -1;
+    m_current_graph_node         = Graph::UNKNOWN_SECTOR;
+    m_last_valid_graph_node      = Graph::UNKNOWN_SECTOR;
+    m_estimated_valid_graph_node = Graph::UNKNOWN_SECTOR;
+    m_on_road                    = false;
+    m_last_triggered_checkline   = -1;
 }   // reset
 
 // ----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


auriamg:
"Distance freezing when going outside of the drivequads", this was done on purpose because trying to guess where an off-road kart is could lead to all sorts of rescue and ranking issue. We'd need to be super careful not to reintroduce those

Alayan-stk-2:
Rescue is a non-issue here because it still uses the last time the kart was on road.

As for ranking, this is a trade-off.

If someone uses a shortcut which is allowed, then it is more accurate. If someone just goes way off-track in a spot where it serves no purpose (including shortcuts which miss a checkline), basing the ranking on the rescue spot is more cautious, but there is no great solution if the kart goes back on track without missing a checkline a bit further.

If someone is off-track, but not by much (by far the most usual situation ; like going on the grass near the road on many tracks), the estimation is quite good, and it is much better to base the distance on it. This is especially important for the live difference with a replay, as the freezing makes it very unreliable.

Alayan-stk-2:
To answer your points : 
"0")The live difference currently exists only when racing against a ghost replay, or when comparing ghost replays. The feature could be extended to live local races with many karts (there is no technical reason it couldn't ; for networked races it probably wouldn't work well however, at least when the difference is <0,5s), but its not its current focus. An important point is that **off-track does not mean far from the track, just out of the drivequads** (not to mention that sometimes, because the drivequads don't cover the full road (happens in many tracks), being on road can still be considered by the game as off-track). See the multiple incorrect jump in the difference in this video demo : https://youtu.be/yFYzsvwJpgQ ; false rank changes can also result of similar situation

When going in a shortcut, a similar effect is observed, in that whichever kart has a slightly higher distance when entering a shortcut will be displayed as a few seconds ahead, with the difference changing progressively towards the real difference at the end of the shortcut.

The tension effect can still be argued for the position indicator, though.

1)If you look at the code change in my commit, you'll see why it didn't work until now : it is an animation, and update of the current/valid track sectors are disabled during animations in the current master.

With my commit, it works as long as the kart remains relatively close to the driveline during the animation, but this is not always the case. This indeed requires a fix in the track drivequads rather than in the code.

2)Remember, once again, _off-track does not mean far from the road_. I consider instances like #3462 as track design bugs, however crossing the lapline in the grass, 20cm to the right of the drivequads limit is also considered off-track and will trigger the "1st place bug". But it is not a track design issue, and there is no confusion for the players as to why this is considered as crossing the line.

3)Track sector is also used by the basket ball iirc, so it is probably used by more than just karts. Likely flyables in general, I have not checked in the code though.

If you think it could be helpful, I could have a look across the codebase to what is saved in the network state to see when smaller types could be used (I'd open a separate issue for it)

4)To compute the positions, the local game will use distances ; so if the local track coords are wrong ; the distance will be wrong. This isn't really a concern if the kart is on the drivequads (it will be updated), but when it goes off-track, the method used in the current master requires to know what was the last valid position. However, thinking about it, this concern doesn't really apply to the method used in this PR, so this may be a non-issue indeed ; and even with the method used in the current master, the difference is likely to be minimal.

PR #3482, opened 2018-09-30
Fixes the cushioning issues

Alayan-stk-2:
This PR fixes the two main issues which plagued cushioning since its last revision this summer : 
- The kart abruptly slowing its fall mid-air (with a big but not so big jump) and flying off-ground in some downward slopes (snow peak, shifting sands, green valley, black forest).
- The kart slowing down too much and flying a bit again after landing on a downward slope ; or not enough and having the chassis crash when landing on an upward slope.

These two issues were the result of determining if the cushioning should be triggered on the base of the vertical speed without taking into account the real height at which the kart evolves.

By using the current height and the height in the previous frame, it is possible to entirely remove the issue of cushioning incorrectly mid-air or lying off the ground ; and to cushion the kart just right for the slope it lands on.

The downside of that approach is that if other factors than the kart falling trigger a brutal height change, this method could be tricked in a wrong cushioning catapulting the kart very high in the air.

I have iteratively identified several situations where this could happen and fixed of all them with additional sanity checks.

The code in this PR fixes all the gameplay issues I'm aware of with cushioning ; but it may be possible to simplify/clean-up some parts which are unneeded leftovers from the previous fixes attempts and the several iterations of this one.

By the way, I suspect that the 1s "no cushioning" duration could be reduced significantly, if it is even needed in the first place. 1s is long enough that sometimes, it would be useful to cushion twice.

### DIFF
--- a/src/karts/rescue_animation.cpp
+++ b/src/karts/rescue_animation.cpp
@@ -25,6 +25,7 @@
 #include "karts/kart_properties.hpp"
 #include "modes/three_strikes_battle.hpp"
 #include "network/network_config.hpp"
+#include "physics/btKart.hpp"
 #include "physics/physics.hpp"
 #include "physics/triangle_mesh.hpp"
 #include "tracks/drive_graph.hpp"
@@ -61,6 +62,7 @@ RescueAnimation::RescueAnimation(AbstractKart *kart, bool is_auto_rescue,
     m_up_vector   = m_kart->getTrans().getBasis().getColumn(1);
     m_xyz         = m_kart->getXYZ();
     m_kart->getAttachment()->clear();
+    m_kart->getVehicle()->resetGroundHeight();
 
     if (NetworkConfig::get()->isNetworking() &&
         NetworkConfig::get()->isServer())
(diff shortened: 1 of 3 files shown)


hiker:
Hi,

I had a look at this issue, and started to fix some related issues first. It turns out that the code that is supposed to keep karts parallel to the ground was de-facto not working (anymore?). I've fixed this, since having the karts parallel to the ground they are landing on will help with the landing (since this way all 4 suspensions can act, not only the rear or front ones). 
Then considering the amount of grief the cushioning has given us, I just disabled it, and instead tweaked the physics - increasing the maximum force the suspension can deliver, and the damping compression (which delivers more force when the suspension is compressed). The result is as far as could tell from my tests that the jumps now seems to work fine, without any additional cushioning code.

Could you try the branch parallel_no_cushion_high_max_force? It is a hack, and also gives you unlimted zipper and nitro to make it easier to test.

Thanks a lot!
   Joerg

hiker:
@Alayan-stk-2 , as discussed on irc I have created a branch fix_cushioning ... and just merged that into the new 'dev' branch (which contains all changes that has an impact on physics, so we can keep servers and clients a bit better in synch, without slowing down our development).

PR #3492, opened 2018-10-03
Network item debugging

hiker:
Hi Benau,

can you test and review my branch?

Thanks!
   Joerg


Benau:
i think
```
void NetworkItemManager::forwardTime(int ticks)
 {
     int new_ticks = World::getWorld()->getTicksSinceStart() + ticks;
     World::getWorld()->setTicks(new_ticks);
```
should be setTicksForRewind, and i still see (rare) stutter when u collected bubblegum(you can try reproduce with spamming bubble gum with m_powerup->set tricks

and there is still assert m_confirmed_state[index] != NULL,
from screenshot seems that it happens when 2 karts collect bubblegum closely,

Benau:
https://streamable.com/n7rh8
~1:22 which this can happen

PR #3495, opened 2018-10-05
Revision of the achievements system

Alayan-stk-2:
This is a rewrite of the greatest part of the achievements system.

Instead of having hardcoded achievements having their progress incremented in various spots in the code ; general counters are used to collect the required data, while the update of the achievements is entirely based on the definition of the achievement in achievements.xml, whose format has been revised.

Several achievements related bugs are fixed by this PR, but its main point is to make much easier to add new achievements in the future.

The only notable downside at the moment is that the achievements progress is not well shown in the achievements screen - an "all tracks" goal will be X/1 instead of X/num_of_tracks ; and it doesn't yet compute what X should be (because I have not yet determined what it should be in complex situations) so it will be either 0 (not done) or N/N.

In general, I don't think that a single number does a good job at capturing the progress ("3/4 goals done, great, but what is the one I'm missing ???"). Ideally, we'd have the ability when selecting the details of an achievement to see its progress tree, though that would require a significant rework.

### DIFF
--- a/src/achievements/achievement.cpp
+++ b/src/achievements/achievement.cpp
@@ -19,8 +19,10 @@
 
 
 #include "achievements/achievement.hpp"
+
 #include "achievements/achievements_manager.hpp"
 #include "achievements/achievement_info.hpp"
+#include "config/player_manager.hpp"
 #include "guiengine/message_queue.hpp"
 #include "io/utf_writer.hpp"
 #include "config/player_manager.hpp"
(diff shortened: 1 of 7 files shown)


data/achievements.xml
```diff
@@ -1,69 +1,178 @@
 <?xml version="1.0"?>
+<!-- List of counters the achievements can query.
+     The format to use is <goal type="name_of_the_counter" value="X"/>
+     where X is the desired value of the counter ;e.g. <won_races value="10"/>
+
+     -- Warning! -- If a goal node is malformed, it is ignored.
```
auriamg: A warning should be logged to the console, too

src/achievements/achievement.cpp
```diff
@@ -47,179 +47,203 @@ Achievement::~Achievement()
...
+ *  This do the same, but with achieved value or fullfilled goals.
  */
-int Achievement::getValue(const std::string & key)
+irr::core::stringw Achievement::getProgressAsString()
 {
-    if (m_progress_map.find(key) != m_progress_map.end())
-        return m_progress_map[key];
-    return 0;
-}
+    //TODO : add a progress computation function.
```
auriamg: Remaining TODO?

src/achievements/achievement.cpp
```diff
@@ -47,179 +47,203 @@ Achievement::~Achievement()
...
-void Achievement::reset()
+void Achievement::setGoalValue(std::string &goal_string, int value)
 {
-    std::map<std::string, int>::iterator iter;
-    for (iter = m_progress_map.begin(); iter != m_progress_map.end(); ++iter)
+    if(m_achieved) // This should not happen, but it costs little to double-check
+        return;
+
+    bool AO = true;
+    bool SAAO = true;
```
auriamg: What does AO and SAAO mean? I assume And-Or, Sum-And-At-Once, but clearer names would make the code easier to read

src/achievements/achievement_info.cpp
```diff
@@ -47,121 +47,126 @@ AchievementInfo::AchievementInfo(const XMLNode * input)
...
+        child.type = type;
+        child.value = value;
+        if (operation == "none")
+            child.operation = OP_NONE;
+        else if (operation == "+")
+            child.operation = OP_ADD;
+        else if (operation == "-")
+            child.operation = OP_SUBSTRACT;
+        else
+            continue; // incorrect operation type, ignore node
```
auriamg: Log a warning instead

src/achievements/achievement_info.cpp
```diff
@@ -47,121 +47,126 @@ AchievementInfo::AchievementInfo(const XMLNode * input)
...
+
+        if (type=="AND" || type=="AND-AT-ONCE" || type=="OR" || type=="SUM")
+        {
+            if (type == "SUM")
+            {
+                if (value <= 0)
+                    continue; // SUM nodes need a strictly positive value
```
auriamg: Log a warning

src/achievements/achievement_info.cpp
```diff
@@ -47,121 +47,126 @@ AchievementInfo::AchievementInfo(const XMLNode * input)
...
+            }
+            else
+            {
+                // Logical operators don't have a value or operation defined
+                if (value != -1)
+                    continue;
```
auriamg: Log a warning here and below

src/achievements/achievement_info.cpp
```diff
@@ -47,121 +47,126 @@ AchievementInfo::AchievementInfo(const XMLNode * input)
...
-    return false;
-}
+} // copyGoalTree
+
+// ----------------------------------------------------------------------------
+/** Returns a string with a numerical value to display the progress of
+ *  this achievement.
+ *  If it has multiple goal, it returns the number of goals. If it has
+ *  only one (it can be a sum), it returns the required value for that goal.
+ * FIXME : don't work well for "all tracks" goals.
```
auriamg: This should be fixed ideally, it's weird to merge in regressions otherwise

src/achievements/achievements_status.cpp
```diff
@@ -157,22 +374,198 @@ void AchievementsStatus::sync(const std::vector<uint32_t> & achieved_ids)
...
+void AchievementsStatus::increaseDataVar(unsigned int achieve_data_id, int increase)
 {
-    //reset all values that need to be reset
-    std::map<uint32_t, Achievement *>::iterator iter;
-    for ( iter = m_achievements.begin(); iter != m_achievements.end(); ++iter ) {
-        iter->second->onRaceEnd();
+    if (achieve_data_id<ACHIEVE_DATA_NUM)
+    {
+        m_variables[achieve_data_id].counter += increase;
+        if (m_variables[achieve_data_id].counter > 10000000)
```
auriamg: what is this hardcoded value? Finding a way without harcoding would be better. If no possible, using instead a "const" with a little explanation would be better

src/modes/world.cpp
```diff
@@ -1610,3 +1569,129 @@ void World::setAITeam()
...
+//-----------------------------------------------------------------------------
+/* This function takes care to update all relevant achievements
+ * and statistics counters related to a finished race. */
+void World::updateAchievementDataEndRace()
+{
+    const unsigned int kart_amount = getNumKarts();
+
+    for(unsigned int i = 0; i < kart_amount; i++)
+    {
+        // TODO : does this work in multiplayer ?
```
auriamg: Should be checked, and the TODO removed

Alayan-stk-2: I'm almost certain it doesn't work, but I'm nearly as sure that it doesn't work either in current master (the core logic in calling the PlayerManager hasn't really changed).

I'll do some tests and post the results back.

Alayan-stk-2:
Considering the huge number of changes in this PR, if there is the need to look back at some subset of its change in the future, I think it'll be easier if it's rebased and merged, rather than squashed in a single gigantic commit and merged.

Issue #3304, opened 2018-06-14
If the user keeps pressing a navigation key in a menu, it should navigate faster


Alayan-stk-2:
In a lot of programs allowing navigation with a key/button, keeping the key/button pressed long enough will trigger a faster navigation between items.

This is especially useful when navigating lists (language, addons, ghost replays) without the mouse, where the current navigation rate feels really sluggish.

EDIT : Also, several close separated inputs should not be throttled. This makes the menu feels very unresponsive.

deveee:
I reverted the pull request, because it causes issues for navigation with gamepad:
https://github.com/supertuxkart/stk-code/issues/3527
And it behaves the same for my gamepad on linux too.

Alayan-stk-2:
@deveee Do you have an idea of what exactly in the PR caused the gamepad bug ? It may be worth a try to see if it can be fixed while keeping the good behavior on keyboards.

Benau:
you can latest master if good for you...

Alayan-stk-2:
Looks good.

PR #3515, opened 2018-10-16
Fix close inputs being throttled, and speed up scrolling when a key is held

rye761:
Fixes issue #3304 

Previously, the timer only worked properly on joysticks I believe, which has now been fixed, and scrolling speeds up if you hold the key!


deveee:
I will test at home, but after quick look it's ok.

Alayan-stk-2:
Tested on my end, and the keys input now work as expected.

Great fix, this bug was very annoying when using the keyboard rather than a mouse.

Issue #3474, opened 2018-09-26
Add word-wrap to the GP editor


Alayan-stk-2:
See picture of the issue

![gp_editor_bug](https://user-images.githubusercontent.com/25536748/46101547-8a257580-c1cc-11e8-9003-8f654acacf8e.png)

Alayan-stk-2:
@rye761 After testing, the fix still has some issues of its own : the track pictures have become tiny, and some track name takes 3 lines and don't fit.

There are a few approach possibles. Finding another way to display the GP name could offer some much needed vertical space, maybe displaying one or two track less at once, and making sure that if it takes 3 lines, it will fit (or having the font smaller so that it fits in 2 lines, whatever look best or is easier to do).

![gp_word_wrap](https://user-images.githubusercontent.com/25536748/47157033-afe3fd00-d2e8-11e8-90a2-d3b13e90a303.png)

The screenshot above is at 1280x720, 1360x768 present similar issues.

At 1024x768 ; there is only 6 tracks displayed at once, but the track icons are bigger and the text is smaller, fitting in 2 lines.

rye761:
Ugh, that's frustrating. Thanks for finding the issue though. I'll continue working on this. What this most likely boils down to though is increasing the child_width for that ribbon, and failing that, making some sort of a change to the way DynamicRibbonWidget computes its cols to better account for long labels.

deveee:
Pause dialog is now far from perfect too:
![zrzut ekranu z 2018-10-18 20-35-35](https://user-images.githubusercontent.com/5108558/47176377-c05e9c80-d315-11e8-93d0-f65afe62b206.png)

I know that 854x480 is rather small resolution and we don't support it, but it's noticeable on 1280x720 too and may happen in other places.

rye761:
Do you have a screenshot of how the pause dialog looked previously?

Alayan-stk-2:
The icons in the second line previously had the same size as the back icon on the first line. The expected result is the icons still having that size. There is more than enough space in the dialog for it.

I noticed the pause dialog issue too by testing in 21:9, but I thought it an artifact of that resolution (the number of elements in the GP editor is the same in 21:9 than in 16:9 and it doesn't fit worse than in 16:9)

qwertychouskie:
Although the text sizing bug is now fixed, the it would still be good to add proper word wrap IMO.  This should probably be re-opened until word wrap is implemented.

Alayan-stk-2:
[...]

PR #3519, opened 2018-10-18
GP Editor Word Wrap

rye761:
Fixes issue #3474 

The issue would only occur as shown in the issue screenshot when you clicked another prix first, and then clicked one where the names were too long. This is because font scaling is done in buildInternalStructure() which was not being called if the number of items was the same as the previous list. This issue was fixed with the first commit.

Second commit adds word wrap, which was suggested in the issue and looks better than tiny text.
image


auriamg:
Merged, thanks!

PR #3520, opened 2018-10-18
Help screen tweaks

qwertychouskie:
Just some small wording and spacing tweaks.

### DIFF
--- a/data/gui/screens/help1.stkgui
+++ b/data/gui/screens/help1.stkgui
@@ -21,7 +21,7 @@
                 <icon-button id="page6" width="128" height="128" icon="gui/icons/mass.png"
                         I18N="Tab in help menu" text="Kart classes"/>
                 <icon-button id="page7" width="128" height="128" icon="gui/icons/options_input.png"
-                        I18N="Tab in help menu" text="Multi-player"/>
+                        I18N="Tab in help menu" text="Multiplayer"/>
             </vertical-tabs>
 
             <spacer width="2%" height="100%"/>
@@ -35,12 +35,12 @@
                 </div>
                 <spacer width="5" height="1%"/>
 
-                <div width="100%" proportion="5" layout="horizontal-row">
+                <div width="100%" proportion="4" layout="horizontal-row">
                     <icon align="center" width="64" height="64" icon="gui/icons/gift.png"/>
                     <spacer width="25" height="25"/>
                     <bubble proportion="7" height="100%" word_wrap="true"
                             I18N="In the help menu"
-                            text="Collect blue gift boxes, they will give you powerups."/>
+                            text="Collect gift boxes, they will give you powerups."/>
                     <spacer width="25" height="25"/>
                     <icon align="center" width="64" height="64" icon="gui/icons/banana.png"/>
                     <spacer width="25" height="25"/>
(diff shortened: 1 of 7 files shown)


data/gui/screens/help1.stkgui
```diff
@@ -35,12 +35,12 @@
...
                 <spacer width="5" height="1%"/>
 
-                <div width="100%" proportion="5" layout="horizontal-row">
+                <div width="100%" proportion="4" layout="horizontal-row">
                     <icon align="center" width="64" height="64" icon="gui/icons/gift.png"/>
                     <spacer width="25" height="25"/>
                     <bubble proportion="7" height="100%" word_wrap="true"
                             I18N="In the help menu"
-                            text="Collect blue gift boxes, they will give you powerups."/>
+                            text="Collect gift boxes, they will give you powerups."/>
```
Alayan-stk-2: Your point maybe is that gift boxes are only blue. But the fact that they are blue is an information. If it's in relation with your christmas suggestion, it would make more sense to do that kind of edit if and when it is merged, no ?

data/gui/screens/help1.stkgui
```diff
@@ -54,28 +54,28 @@
                     <spacer width="25" height="25"/>
                     <bubble proportion="3" height="100%" word_wrap="true"
                             I18N="In the help menu"
-                            text="Collecting nitro allows you to get speed boosts whenever you wish by pressing the appropriate key or button. You can see your current level of nitro in the gauge at the bottom-right of the race screen."/>
-                    </div>
-
-                <div width="100%" proportion="4" layout="horizontal-row">
-                    <icon align="center" width="64" height="64" icon="gui/icons/gui_lock.png"/>
+                            text="Collecting nitro allows you to use small speed boosts by pressing the assigned key or button*. You can see your current level of nitro in the gauge at the bottom-right of the race screen."/>
```
Alayan-stk-2: You can use nitro to get a speed boost, you can't use a speed boost.

I don't see the point of writing "small". In SuperTux while skidding, nitro still gives +17% of speed. In Novice, it gives +33% of speed. The duration depends of how much nitro you have, but isn't that bad. In my book, that doesn't deserve to be called small.

data/gui/screens/help1.stkgui
```diff
@@ -54,28 +54,28 @@
...
 
-                <div width="100%" proportion="6" layout="horizontal-row">
-                    <icon align="center" width="64" height="64" icon="gui/icons/android/drift.png"/>
+                <div width="100%" proportion="4" layout="horizontal-row">
+                    <icon align="center" width="64" height="64" icon="gui/icons/gui_lock.png"/>
                     <spacer width="25" height="25"/>
                     <bubble proportion="2" height="100%" word_wrap="true"
                             I18N="In the help menu"
-                            text="You can skid by pressing a special key or button. Successive short skids help to take sharp turns; while medium skids will boost your speed, long skids more so. You can't stop turning while skidding, so orient your kart carefully before!"/>
+                            text="If you see a button with a lock like this one, you need to complete a challenge in Story Mode to unlock it."/>
```
Alayan-stk-2: It's true that challenges only happen in story mode. I'm not so sure it is worth to break translations to add it, though, especially as the story mode help also explains it. Have there been people who had difficulty finding challenges ?

data/gui/screens/help1.stkgui
```diff
@@ -54,28 +54,28 @@
...
+                            text="Collecting nitro allows you to use small speed boosts by pressing the assigned key or button*. You can see your current level of nitro in the gauge at the bottom-right of the race screen."/>
+                </div>
+                
+                <div width="100%" proportion="7" layout="horizontal-row">
+                    <icon align="center" width="64" height="64" icon="gui/icons/android/drift.png"/>
                     <spacer width="25" height="25"/>
                     <bubble proportion="2" height="100%" word_wrap="true"
                             I18N="In the help menu"
-                            text="If you see a button with a lock like this one, you need to complete a challenge to unlock it."/>
+                            text="You can skid by pressing the assigned key or button*. Successive short skids help to take sharp turns; while medium to long skids will give you a small or medium boost, depending on the length of the skid. You can't stop turning while skidding, so orient your kart carefully before!"/>
```
Alayan-stk-2: There is a reason my formulation for the skid help text was not the one you suggested initially. Yours was too long. This new modification suffers from the same flaw.

What's wrong with the current text ?

hiker:
I've applied all changes that fixes typos, but as discussed by Alayan left the other text unchanged.

Thanks a lot for the patch!
   Joerg

PR #3522, opened 2018-10-19
Gyroscope support for Android

pelya:
Gyroscope is superior to accelerometer in every way - it's more responsive, less shaky, has cool camera tilt effect, and works even when the phone lies flat on the table.

### DIFF
--- a/data/gui/dialogs/android/init_android.stkgui
+++ b/data/gui/dialogs/android/init_android.stkgui
@@ -10,6 +10,8 @@
             <ribbon id="control_type" proportion="1" width="100%" align="center">
                 <icon-button id="accelerometer" width="fit" height="fit" icon="gui/icons/difficulty_medium.png"
                         I18N="Control type" text="Accelerometer"/>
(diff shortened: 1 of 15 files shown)


qwertychouskie:
Is it possible to get a video of this in action?

pelya:
Here: https://www.youtube.com/watch?v=qNDcElUvZmo

pelya:
Nice, thanks!

Accelerometer should stay at 30 FPS, because it is shaky - it detects any small movement of your hands, so if we enable camera tilt effect for accelerometer only, the camera will shake wildly. At 60 FPS accelerometer will simply shake more often, without providing any increase in accuracy.

Gyroscope on the other hand ignores any lateral movement, and only detects rotation, so you can even walk while playing and the camera will not shake much. Since we have immediate visual representation of how much you are rotating your phone to steer, it needs to measure at the same rate as your FPS, to keep the game camera horizon parallel to the real-world horizon.

PR #3523, opened 2018-10-19
Always update font size in DynamicRibbonWidget

### LINKED ISSUE #3474: Add word-wrap to the GP editor

rye761:
Partially fixes #3474 

This pull request ensures that the font size is always updated when DynamicRibbonWidget items are updated by moving the font sizing code to updateItemDisplay() instead up buildInternalStructure(), which is only called when the number of items is changed. The text overlap occurred because the build-in prixs have the same number of items and switching between them didn't call buildInternalStructure(), so the font size wasn't updated.

This pull request doesn't implement word wrap, it only ensures DynamicRibbonWidget behaves consistently and as expected.

### DIFF
--- a/src/guiengine/widgets/dynamic_ribbon_widget.cpp
+++ b/src/guiengine/widgets/dynamic_ribbon_widget.cpp
@@ -399,13 +399,6 @@ void DynamicRibbonWidget::buildInternalStructure()
         ribbon->m_properties[PROP_ID] = name.str();
         ribbon->m_event_handler = this;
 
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
I had a quick look at the code, and it seems to make sense. I expect it to work without regression, but I have not tested it to confirm yet.

auriamg:
I've done more thorough testing this time, and found no issue, thanks!

PR #3529, opened 2018-10-23
Fix for gyroscope camera using world coordinates instead of camera orientation

pelya:
Noticed that bug too late, sorry


deveee:
Merged, thanks!

PR #3533, opened 2018-10-30
Achievement sorting

### LINKED ISSUE #3513: Make the achievements list sortable
It is currently not, be it in 0.9.3 or in current git.

The header display the sort icons when clicked, but it does nothing.

rye761:
Fixes #3513 

Makes the achievements list sortable by implementing IListWidgetHeaderListener.

I used online_profile_friends as a model, the resulting class structure is very similar.

Works with both the local achievements screen and viewing another online player's achievements.

### DIFF
--- a/src/states_screens/online/online_profile_achievements.cpp
+++ b/src/states_screens/online/online_profile_achievements.cpp
@@ -48,6 +48,9 @@ BaseOnlineProfileAchievements::BaseOnlineProfileAchievements(const std::string &
                              : OnlineProfileBase(name)
 {
     m_selected_achievement_index = -1;
+    m_sort_column = 0;
+    m_sort_desc = false;
+    m_sort_default = true;
 }   // BaseOnlineProfileAchievements
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
Couldn't get it to crash or to break things :)

But the sort ought to be improved for numeric values, the alphanumerical comparison doesn't work well.

The two issues I notice are : 
- 0/1 placed after 1/1 in the goal column when sorting in ascending order (I don't get why it happens though, 0 should be less than 1)
- 5/5 placed after 24/24, because 2 < 5, even though 24 > 5.

And there is another consideration : isn't 3/3 more complete than 4/5 ? So shouldn't 3/3 be displayed after 4/5 ? If we decide that it should sort by "completion progress", then we need to break up the string and to compute the fractional progress. (3/3 = 100% = 1.0 while 4/5 = 80% = 0.8).

Alayan-stk-2:
It doesn't seem like this will get fixed, so I'll merge this and fix it myself.

rye761:
Sorry, I have been pretty busy lately. I did do some work on it but the issue I kept running into was substring giving me a vector subscript out of range error even though the string should have been large enough. Hopefully you find a better method. Thanks!

Alayan-stk-2:
I will manage it, don't worry.

Try to see if you can get your keyboard navigation speed fix to not create issues with gamepads, it would be great to have it back !

PR #3538, opened 2018-11-02
added lap icon to the the kart icons in RaceUI that have finished the race already

Mrxx99:
This PR adds the lap flag, that is used in the current Race UI as a symbol for the lap, to the kart icons if they have finished the race. I think this is really informative because you may adjust your strategy if you know the karts in front of you are already finished.

I tested this in single player, splitscreen and online play and made sure there are no problems with the other game modes (The lap icons will only be shown for normal races and time trials)

lapiconforfinishedkarts

Here is a video: https://youtu.be/Em5OfQT3cBk

### DIFF
--- a/src/states_screens/race_gui_base.cpp
+++ b/src/states_screens/race_gui_base.cpp
@@ -1039,6 +1039,21 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w)
                                                       NULL, true);
         }
     }
+
+	//lap flag for finished karts
+	if (kart->hasFinishedRace())
+	{
+		video::ITexture *icon_lapflag = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
+		if (icon_lapflag != NULL)
+		{
+			const core::rect<s32> rect(core::position2d<s32>(0, 0),
+				icon_lapflag->getSize());
+			const core::rect<s32> pos1(x - 20, y - 10, x + w - 20, y + w - 10);
+			draw2DImage(icon_lapflag,
+				pos1, rect, NULL,
+				NULL, true);
+		}
+	}
 #endif
 }   // drawPlayerIcon
 
--- a/src/states_screens/race_gui_base.cpp
+++ b/src/states_screens/race_gui_base.cpp
@@ -99,6 +99,7 @@ RaceGUIBase::RaceGUIBase()
     m_gauge_full_bright     = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_full_bright.png"));
     m_gauge_empty           = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_empty.png"));
     m_gauge_goal            = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_goal.png" ));
+    m_lap_flag              = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
     m_dist_show_overlap     = 2;
     m_icons_inertia         = 2;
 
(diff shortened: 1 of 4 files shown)


src/states_screens/race_gui_base.cpp
```diff
@@ -1039,6 +1039,21 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w)
                                                       NULL, true);
         }
     }
+
+	//lap flag for finished karts
+	if (kart->hasFinishedRace())
+	{
+        video::ITexture *icon_lapflag = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
+        if (icon_lapflag != NULL)
+		{
```
auriamg: our conventions are to indent using spaces, not tabs

src/states_screens/race_gui_base.cpp
```diff
@@ -1039,6 +1039,21 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w)
                                                       NULL, true);
         }
     }
+
+	//lap flag for finished karts
+	if (kart->hasFinishedRace())
+	{
+        video::ITexture *icon_lapflag = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
```
auriamg: preferably, this texture should be retrieved one and stored in a  variable, not retrieved by name on every frame

src/states_screens/race_gui_base.cpp
```diff
@@ -1039,6 +1039,21 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w)
...
+        if (icon_lapflag != NULL)
+		{
+            const core::rect<s32> rect(core::position2d<s32>(0, 0),
+                                       icon_lapflag->getSize());
+            const core::rect<s32> pos1(x - 20, y - 10, x + w - 20, y + w - 10);
+            draw2DImage(icon_lapflag,
+                                                      pos1, rect, NULL,
```
auriamg: weird indentation

src/states_screens/race_gui_base.cpp
```diff
@@ -1039,6 +1039,21 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w)
...
+
+	//lap flag for finished karts
+	if (kart->hasFinishedRace())
+	{
+        video::ITexture *icon_lapflag = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
+        if (icon_lapflag != NULL)
+		{
+            const core::rect<s32> rect(core::position2d<s32>(0, 0),
+                                       icon_lapflag->getSize());
+            const core::rect<s32> pos1(x - 20, y - 10, x + w - 20, y + w - 10);
```
auriamg: many hardcoded values here, have you tested varied resolutions?

src/states_screens/race_gui_base.cpp
Mrxx99: Sorry, I thought I converted everything to space, seems like I missed that line

src/states_screens/race_gui_base.cpp
Mrxx99: First wamted to do that as it is also done that way for the lap flag in the top right corner, but than I saw that in this function every texture is retrieved by name so I made it this way.

src/states_screens/race_gui_base.cpp
Mrxx99: Thought so too, but it's the way these calls are indented in the function and I thought this is done intentionally.

src/states_screens/race_gui_base.cpp
Mrxx99: Same as above, I just copied from the attachments here so I suspect I should work if it works for the attachments.

Alayan-stk-2:
Nice small enhancement.

I have no further remark besides what Auria already noticed.

src/states_screens/race_gui_base.cpp
auriamg: No, that's definitely not right, maybe it's a mistake or a leftover from a merge (merge tools are usually not indentation aware). It would be preferable to fix this in your code, even if there are issues elsewhere

auriamg:
Merged, thanks

src/states_screens/race_gui_base.cpp
```diff
@@ -99,6 +99,7 @@ RaceGUIBase::RaceGUIBase()
     m_gauge_full_bright     = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_full_bright.png"));
     m_gauge_empty           = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_empty.png"));
     m_gauge_goal            = irr_driver->getTexture(file_manager->getAsset(FileManager::GUI_ICON,"gauge_goal.png" ));
+    m_lap_flag              = irr_driver->getTexture(FileManager::GUI_ICON, "lap_flag.png");
```
qwertychouskie: Any reason why this is different than the one above it?

Mrxx99: Yes, I moved that line from race_gui.cpp to here. The method has different overloads, so I guess the lines above could be simplified as well but didn't test that. I thought changing them shouldn't be part off this PR.

PR #3612, opened 2018-11-23
Fix issues with COPYING (see desc.)

qwertychouskie:
See https://softwareengineering.stackexchange.com/questions/281949/how-to-properly-apply-a-license for why the the "How to Apply These Terms to Your New Programs" part was added.

The GPL was copy-pasted from https://www.gnu.org/licenses/gpl-3.0-standalone.html, but with some formatting changed to be like https://www.gnu.org/licenses/gpl.txt (only ASCII characters, etc.)

The top line is based on https://github.com/supertuxkart/stk-code/blob/master/src/main.cpp#L7

Note that Git shows a lot of changed lines because of the whitespace that got changed from the copy-paste.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,18 +1,21 @@
+The SuperTuxKart code is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version.
 
-The SuperTuxKart code is released under GNU GPL (see full text below).
-SuperTuxKart data files (textures, models, sounds, music, etc.) are released under various licenses, see 'license.txt' files
-through the various data subdirectories. Data files are released under a mixture of the following licenses :
+The SuperTuxKart data files (textures, models, sounds, music, etc.) are released under a mixture of licenses including, but not limited to, the following:
 
-Gnu GPL 2.0 and 3.0+
-Creative-Commons-BY(-SA) 3.0
+GNU General Public License (GNU GPL)
+Creative Commons Attribution (CC-BY)
+Creative Commons Attribution-ShareAlike (CC-BY-SA)
 Public Domain
 
-________________________________________________________________________________
+See the 'licenses.txt' files throughout the various data subdirectories for the licenses of each file.
+
+_____________________________________________________________________
+
 GNU GENERAL PUBLIC LICENSE
 
 Version 3, 29 June 2007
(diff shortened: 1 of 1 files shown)


COPYING
```diff
@@ -96,124 +99,162 @@ You may charge any price or no price for each copy that you convey, and you may
...
 
 17. Interpretation of Sections 15 and 16.
 
-    If the disclaimer of warranty and limitation of liability provided above cannot be given local legal effect according to their terms, reviewing courts shall apply local law that most closely approximates an absolute waiver of all civil liability in connection with the Program, unless a warranty or assumption of liability accompanies a copy of the Program in return for a fee.
+If the disclaimer of warranty and limitation of liability provided above cannot be given local legal effect according to their terms, reviewing courts shall apply local law that most closely approximates an absolute waiver of all civil liability in connection with the Program, unless a warranty or assumption of liability accompanies a copy of the Program in return for a fee.
 
 END OF TERMS AND CONDITIONS
+
+How to Apply These Terms to Your New Programs
```
Alayan-stk-2: I don't see the point at all of including this part.

COPYING
```diff
@@ -96,124 +99,162 @@ You may charge any price or no price for each copy that you convey, and you may
...
+    d) If the work has interactive user interfaces, each must display Appropriate Legal Notices; however, if the Program has interactive interfaces that do not display Appropriate Legal Notices, your work need not make them do so.
 
-    A compilation of a covered work with other separate and independent works, which are not by their nature extensions of the covered work, and which are not combined with it such as to form a larger program, in or on a volume of a storage or distribution medium, is called an "aggregate" if the compilation and its resulting copyright are not used to limit the access or legal rights of the compilation's users beyond what the individual works permit. Inclusion of a covered work in an aggregate does not cause this License to apply to the other parts of the aggregate.
```
Alayan-stk-2: Do these trivial formatting change really matter ?

qwertychouskie: No, they just happened when re-copy-pasting the text.  The reason the whole text was  re-copy-pasted was because of https://github.com/supertuxkart/stk-code/pull/3612/files#diff-7116ef0705885343c9e1b2171a06be0eL49, I don't know if there were any other typos in the old version, and better safe than sorry IMO.

hiker:
Thanks!

PR #3631, opened 2018-11-27
Prevent early explosive items

Fouks0:
There are complaints of people getting too much often caked by AIs very early in a race. As everyone is packed during the first ~10 s, explosive homing items tend to create a lot of mess and frustration in this situation.

In this pull request, I propose a simple patch that disables the rubber ball and the cake during the first 15 s by replacing them by bowling balls.

As this is my first pull request here, I hope that I did this right.

### DIFF
--- a/src/items/powerup_manager.cpp
+++ b/src/items/powerup_manager.cpp
@@ -602,6 +602,13 @@ PowerupManager::PowerupType PowerupManager::getRandomPowerup(unsigned int pos,
     }
     else
         *n=1;
+
+    // Prevents early explosive items
+    if (stk_config->ticks2Time(World::getWorld()->getTicksSinceStart()) < 15.)
+    {
+        if (powerup == POWERUP_CAKE || powerup == POWERUP_RUBBERBALL)
+            powerup = POWERUP_BOWLING;
+    }
     return (PowerupType)powerup;
 }   // getRandomPowerup
 


Alayan-stk-2:
I don't think always replacing them by bowling balls is the best solution. Sure, it would only affect one kart instead of 3 or 4 at once like a cake, but when karts are still very close, it is very easy to hit one. In online races as well as races against AIs, when I get first to an item box, I often just fire backwards without checking, and frequently get a hit. Maybe rather trying to pick again a random item would work better.

Fouks0:
Ok so let's reroll the dice until it does not pick a rubber ball nor a cake. Would you then also disable bowling balls or rubber balls + cakes would be enough?

Alayan-stk-2:
> Ok so let's reroll the dice until it does not pick a rubber ball nor a cake.

Put a limit to avoid an infinite loop if someone customizes powerup.xml. Even one reroll would already make those items much less frequent at the beginning, two or three extremely rare. To reroll, applying some fixed prime multiplier (something like 617) to the random number seems a good way. If you just do an addition, the reroll will have skewed results because numbers giving a specific item are clustered.

> Would you then also disable bowling balls or rubber balls + cakes would be enough?

No strong feelings. Putting a bubble gum on the ground is not much friendlier than a bowling ball, neither is a swatter, and we don't want either to have 100% zipper/plunger/switch at the beginning.

To get a better balance overall in the future, we will need some new item. I did a try on this branch, but it is unfinished : https://github.com/supertuxkart/stk-code/tree/GiantKartItem

Fouks0:
Yup I thought about the possible infinite loop. I think that I will just limit to 10 loops, at this point it would almost be impossible that someone gets a basket ball or a rubber ball.

hiker:
Picking a new item should be deterministic, otherwise we have to closely look at rewinds to make sure we get the same behaviour (without saving more state). So a simple 'reroll' won't work (since de-facto there is no random number and you will get the same item). Am happy to just pick a bowling ball (or bubble gum or so). It might be a better idea to look at the tracks and just make sure that you can't collect items that early.

Fouks0:
The code for up to 10 rerolls with a multiplier 617*i for the random_number is ready, I can commit it or open another pull request if wanted.

Replacing by a bowling ball is a very simple solution for this problem, that has for sure no side effect. This is why I did this rather than rerolling.
Rerolling might be more elegant, but also probably comes with complex consequences, and to be honest, I am not sure how the RNG works in this game. At least, if someone edited the powerup.xml to make the bowling balls never appear, this will still be the case with this option.

Alayan-stk-2:
> Picking a new item should be deterministic, otherwise we have to closely look at rewinds to make sure we get the same behaviour (without saving more state). So a simple 'reroll' won't work (since de-facto there is no random number and you will get the same item).

It will work, because the "reroll" multiply the number received as a function parameter by a fixed prime until it gets to another item type or to a fixed limit. The operations applied are meant to give results which are not too skewed, but they are fully deterministic.

>It might be a better idea to look at the tracks and just make sure that you can't collect items that early.

This has been done in Snow Peak (though I asked Benau to move, not to remove), and could be done in other places. But realistically, when a lap is 30s or 40s long, and the phase with many karts still quite close often lasting up to half of the first lap, this is limited.

Benau:
I think soon there will be new beta apk android, so do you want to just use bowling?

and alayan do you want to make the new powerup.xml to a branch so people can test and give feedback

Alayan-stk-2:
I think this should be tested with the _**deterministic**_ "reroll".

I pushed the powerup.xml tweaks to a branch.

hiker:
I am happy to merge this pull request in the way it is (i.e. deterministic just use a ball), but can you make the '15' configurable in our stk_config.xml file? data/stk_config.xml already has a (small) <powerup ...> section, where this would fit in. See src/config/stk_config.?pp for details.

Thanks a lot!
   Joerg

PR #3632, opened 2018-11-28
Fix fire before start

Fouks0:
It was possible to fire items before the race started. This is an issue for Time Trial races, where you could waste your boosters during startup.

This pull request fixes this problem. Also, pushing the fire button no longer induces the Penalty Time, there is no reason to penalize here.


auriamg:
Thanks, merged!

PR #3638, opened 2018-12-01
Move soccer options to track info screen

Mrxx99:
What this pull request does:
It moves the options for setting up a soccer game (goal/time targets) from the screen where you can select your team to the track info screen.

Why?
Because it makes much more sense to have all options on the same place. Right now the settings how many AI players or random item location are on another place like the goal/time targets.

Additional info:
For this the time-enabled spinner is moved to the track info screen, formatted to fit in and renamed to target-type-spinnner, the lap-spinner is reused for both goal and time target and renamed to target-value-spinner.
As a consequence of this they share the same max value of 20 (that was the highest of the three spinners and the one from the lap spinner). Max value could also be set in the code dependent of the spinner, but it does not hurt to share the value (goal target max increases from 15 goals and time target max from 10 minutes; I could apply these again if whished).
When changing the target type the spinner below it changes it's text and value according to the selection, no inactive spinners like before.
Because there was one spinner added all the other widgets move one spot down, but there is still enough space in all the other game modes.

Here is a screenshot:
menu-2018 12 01_15 17 14

### DIFF
--- a/data/gui/screens/soccer_setup.stkgui
+++ b/data/gui/screens/soccer_setup.stkgui
@@ -11,27 +11,6 @@
 
         <spacer height="1%" width="25"/>
 
-        <div layout="horizontal-row" width="100%" height="fit" align="left">
-            <bright width="fit" height="100%"
-                    I18N="In soccer setup screen" text="Soccer game type" text_align="left" />
-            <spacer width="2%" height="25"/>
-            <spinner id="time_enabled" width="30%" wrap_around="true"/>
-        </div>
-
-        <div layout="horizontal-row" width="100%" height="fit" align="left">
-            <bright width="fit" height="100%"
-                    I18N="In soccer setup screen" text="Number of goals to win" text_align="left" />
-            <spacer width="2%" height="25"/>
-            <spinner id="goalamount" width="15%" min_value="1" max_value="10"/>
-        </div>
-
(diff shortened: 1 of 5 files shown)


Mrxx99:
postponing, because of problems on small resolutions, need to fix them first

Mrxx99:
@auriamg do you have any idea how I can collapse widgets so they are invisible and don't take any space?
@Benau suggested setting isVisible to false inside the beforeAddingWidget function but doing so they still are only invisible and take the same space as if visible.

auriamg:
@Mrxx99 It's not as easy as one would hope... Try something like

```
rect_box->setVisible(false);
rect_box->m_properties[GUIEngine::PROP_HEIGHT] = "0";
calculateLayout();
```

example taken from /src/states_screens/online/tracks_screen.cpp
Warning, this will be permanent and the widget will remain hidden next time you visit the screen. So you probably also want an "else" that can revert those changes

Mrxx99:
@auriamg thanks very much, that worked :)

If you have time could you please try it and tell what you think? (Also @Benau @hiker @deveee)
The base implemented here could also be used for adding an option to play FFA offline,
which I want to try to do next.

The layout of the first spinner still could be improved though.

auriamg:
This looks good to me, good work. Since we have a release coming soonish, I wonder if merging this should perhaps wait after the release ( @hiker what do you think? )

PR #3641, opened 2018-12-06
Feature/material based for blender2.8 compatibility

samuncle:
Add support for materials directly from blender, it's done for two major reasons:

* First we need this to be compatible with Blender 2.8 since this upcoming version will remove the ability to add textures without using a material first
* Then it allows more flexibility to have different type of material using the same texture (thus using less  VRAM and less disk space).
* It's also fully compatible backward

How to use it?

In the materials.xml, you can set the name of the material. You need to use the extension .mat
`<material name="sand.mat" tex-layer-0="hd_palmTreeLeaf_a_Albedo.png" normal-map="hd_palmTreeBark_a_Normal.png"/>`
And then you can set the texture for the layer-0

### DIFF
--- a/src/graphics/material.cpp
+++ b/src/graphics/material.cpp
@@ -60,11 +60,23 @@ Material::Material(const XMLNode *node, bool deprecated)
     m_installed = false;
 
     m_sampler_path[0] = "unicolor_white";
-    node->get("name",      &m_texname);
+    node->get("name",      &m_matname);
+    if (m_matname=="")
+    {
+        throw std::runtime_error("[Material] No material name specified "
+                                 "in file\n");
+    }
+
+    node->get("tex-layer-0",      &m_texname);
     if (m_texname=="")
     {
-        throw std::runtime_error("[Material] No texture name specified "
-                                 "in file\n");
+        // We fallback to using the name as the slot 0 texture
+        node->get("name",      &m_texname);
+        if (m_texname=="")
+        {
+            throw std::runtime_error("[Material] No material name specified "
+                                     "in file\n");
+        }
     }
 
     const std::string& relative_path = file_manager->searchTexture(m_texname);
--- a/src/graphics/material.cpp
+++ b/src/graphics/material.cpp
@@ -79,6 +79,9 @@ Material::Material(const XMLNode *node, bool deprecated)
         }
     }
 
+    Log::warn("Material", "debug name: =========== %s",
+    m_matname.c_str());
+
     const std::string& relative_path = file_manager->searchTexture(m_texname);
     if (relative_path.empty())
     {
--- a/src/graphics/sp_mesh_loader.cpp
+++ b/src/graphics/sp_mesh_loader.cpp
@@ -131,10 +131,23 @@ scene::IAnimatedMesh* SPMeshLoader::createMesh(io::IReadFile* f)
                     tex_name_1 = full_path;
                 }
             }
-            sp_mat_map[id] =
-                std::make_tuple(
-                material_manager->getMaterialSPM(tex_name_1, tex_name_2),
-                !tex_name_1.empty(), !tex_name_2.empty());
+
+			// If we detect a material we look for the name rather than the texture
+            if (tex_name_1.find(".mat") != -1)
+            {
+                Log::info("SPM material found", tex_name_1.c_str());
+				sp_mat_map[id] =
+				std::make_tuple(
+					material_manager->getMaterialSPMByName(tex_name_1),
+					!tex_name_1.empty(), !tex_name_2.empty());
+            }
+			else
+			{
+				sp_mat_map[id] =
+				std::make_tuple(
+					material_manager->getMaterialSPM(tex_name_1, tex_name_2),
+					!tex_name_1.empty(), !tex_name_2.empty());
+			}
         }
         else
         {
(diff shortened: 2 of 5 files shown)


src/graphics/material.cpp
```diff
@@ -60,13 +60,28 @@ Material::Material(const XMLNode *node, bool deprecated)
...
+    node->get("name",      &m_matname);
+    if (m_matname=="")
     {
-        throw std::runtime_error("[Material] No texture name specified "
+        throw std::runtime_error("[Material] No material name specified "
                                  "in file\n");
     }
 
+    node->get("tex-layer-0",      &m_texname);
+    if (m_texname=="")
```
Alayan-stk-2: These two lines could be replaced by : 
if (!node->get("tex-layer-0", &m_texname))

It should not be a functional change only a coding style change, using a more compact style already present in other files.

src/graphics/material.cpp
```diff
@@ -60,13 +60,28 @@ Material::Material(const XMLNode *node, bool deprecated)
...
+    {
+        // We fallback to using the name as the slot 0 texture
+        node->get("name",      &m_texname);
+        if (m_texname=="")
+        {
+            throw std::runtime_error("[Material] No material name specified "
+                                     "in file\n");
```
Alayan-stk-2: This error-checking doesn't make much sense. You already try to get the "name" node value for m_matname, and throw an error if there is none, so this part of the code will never be reached.

There is also no need to retrieve again the name node.

You could replace all this by a simple : 
m_texname = m_matname;

src/graphics/material.cpp
```diff
@@ -60,13 +60,28 @@ Material::Material(const XMLNode *node, bool deprecated)
...
+        }
+    }
+
+    Log::warn("Material", "debug name: =========== %s",
```
Alayan-stk-2: Why all those === ?

src/graphics/sp_mesh_loader.cpp
```diff
@@ -131,10 +131,23 @@ scene::IAnimatedMesh* SPMeshLoader::createMesh(io::IReadFile* f)
...
+			// If we detect a material we look for the name rather than the texture
+            if (tex_name_1.find(".mat") != -1)
+            {
+                Log::info("SPM material found", tex_name_1.c_str());
+				sp_mat_map[id] =
+				std::make_tuple(
+					material_manager->getMaterialSPMByName(tex_name_1),
+					!tex_name_1.empty(), !tex_name_2.empty());
+            }
+			else
```
Alayan-stk-2: The else block should be indented similarly to the if block.

qwertychouskie:
@samuncle Now that Blender 2.8 is officially released, can you update this as per @Alayan-stk-2's comments?

src/graphics/material.cpp
Ionga: O="hd_palmTreeLeaf_alAlbedo.png"/>

Andor146:
2019. aug. 6. 18:44 ezt írta ("Ionga" <notifications@github.com>):

*@Ionga* commented on this pull request.
------------------------------

In src/graphics/material.cpp
<https://github.com/supertuxkart/stk-code/pull/3641#discussion_r311190906>:

>                                   "in file\n");
     }

+    node->get("tex-layer-0",      &m_texname);
+    if (m_texname=="")
+    {
+        // We fallback to using the name as the slot 0 texture
+        node->get("name",      &m_texname);
+        if (m_texname=="")
+        {
+            throw std::runtime_error("[Material] No material name specified "
+                                     "in file\n");

O="hd_palmTreeLeaf_alAlbedo.png"/>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/supertuxkart/stk-code/pull/3641?email_source=notifications&email_token=AMW7STMMSGNIBEYDA43PFTDQDGZ77A5CNFSM4GI5Y4B2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCAXOGLY#discussion_r311190906>,
or mute the thread

RQWorldblender:
My version of the Blender scripts at https://github.com/RQWorldblender/stk-blender/ are capable of supporting this new material naming system for sure, as it's just a matter of changing what the material exporter script and the SPM exporter script write.

> Don't use anything other than .png / .jpg (image file) inside .spm file binrary (like .mat), because too many places need to adapt for such change.

The way forward would be to increase the kart and track versions to support this new material naming system, which I can also change in my scripts. The SPM binary versions would also need to be increased for this support. That way, backwards compatibility is preserved with legacy kart and track versions. I can update all of my add-on content to support this system, but I have no control over content published by other people. Such content, if not updated, will continue to be usable, but they won't be able to utilize this system.
- SPM binary version changes from 1 to 2.
- Kart version changes from 3 to 4.
- Track version changes from 7 to 8.

These new versions would then be supported only by STK 1.3 and later, with older versions being unable to use them. Older versions will continue to use usable with older STK versions as expected, as well as this upcoming 1.3 version.

Benau:
i think most artist will find your current blender 2.8x script enough

(btw if it's ready tell me to merge then I will close this pr, I don't want to do massive update to those things)

Benau:
exporter is done, same texture different material may be done in the future if there is market

PR #3651, opened 2018-12-16
Update INSTALL.md

rusty-snake:
Use apt instead of apt-get.

As described [here](https://itsfoss.com/apt-vs-apt-get-difference/ "https://itsfoss.com/apt-vs-apt-get-difference/"), apt is better suited for fronted use and is included in all current apt based distros.


deveee:
I don't see any reason to change it

PR #3659, opened 2018-12-22
Fix build without _IRR_COMPILE_WITH_JOYSTICK_EVENTS_

Partmedia:
This code calls *close*, which requires *unistd.h* which is currently only in the #if block. Fix build when _IRR_COMPILE_WITH_JOYSTICK_EVENTS_ is not set by moving the #include.


deveee:
Merged, thanks!

PR #3660, opened 2018-12-22
Fix build with joystick on FreeBSD

Partmedia:
FreeBSD doesn't have linux/joystick.h. Fix build on FreeBSD when joystick support is enabled (the default).


auriamg:
I can't test since i don't have freeBSD, but looks very simple and straightforward, so I will merge it, thanks

PR #3661, opened 2018-12-23
Use floating trigonometric functions

Partmedia:
Replace calls to double-precision trigonometric functions to their floating-point counterparts. Depending on this platform and standard math library, this can result in less computation.

### DIFF
--- a/src/graphics/camera.cpp
+++ b/src/graphics/camera.cpp
@@ -273,7 +273,7 @@ void Camera::update(float dt)
 
     if (race_manager->getNumLocalPlayers() < 2)
     {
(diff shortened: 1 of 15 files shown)


Partmedia:
I've profiled this and it does no harm. None of these trig functions were in fast paths, so there's not really any performance gain. However, with this patch there's now no implicit float to double conversion and back, and you really are spending a few less cycles computing a result you don't need.

auriamg:
@Partmedia This looks good to me. However, I'd like @hiker 's review too, since he knows physics and networking much better than I do. I think he's currently unavailable during the holidays, so please allow some time for everyone to come back from the holidays

hiker:
Thanks a lot!!

PR #3662, opened 2018-12-23
Add option to disable light scattering

Partmedia:
On old (circa 2011) graphics hardware, I have the sad option of either using fast but ugly graphics with the old pipeline (graphics level 2) or nice-looking but very, very slow (a second in game takes 3 real seconds with very poor control response) render on some maps like Cocoa Temple.

For some context, on my hardware, the solid render at Cocoa Temple takes 15 ms. In the room with the spinning gear, light scatter takes up to 48 ms. That makes for a really crappy experience.

This option, currently not visible in-game, preserving existing behavior by default, adds a configuration setting that disables the light scattering shader in the pipeline. This means that you get the new pipeline, with the new point light sources, only without the later light scattering stage. This means that users running hardware like the stuff I run don't have to pick either old pipeline or unplayable graphics.

### DIFF
--- a/src/config/user_config.hpp
+++ b/src/config/user_config.hpp
@@ -851,6 +851,10 @@ namespace UserConfigParams
             PARAM_DEFAULT(BoolUserConfigParam(false,
                            "ssao", &m_graphics_quality,
                            "Enable Screen Space Ambient Occlusion") );
+    PARAM_PREFIX BoolUserConfigParam         m_light_scatter
+            PARAM_DEFAULT(BoolUserConfigParam(true,
+                           "light_scatter", &m_graphics_quality,
+                           "Enable light scattering shaders") );
     PARAM_PREFIX IntUserConfigParam          m_shadows_resolution
             PARAM_DEFAULT( IntUserConfigParam(0,
                            "shadows_resolution", &m_graphics_quality,
--- a/src/graphics/shader_based_renderer.cpp
+++ b/src/graphics/shader_based_renderer.cpp
@@ -297,7 +297,7 @@ void ShaderBasedRenderer::renderSceneDeferred(scene::ICameraSceneNode * const ca
     // Render discrete lights scattering
     if (track && track->isFogEnabled())
     {
-        PROFILER_PUSH_CPU_MARKER("- PointLight Scatter", 0xFF, 0x00, 0x00);
+        PROFILER_PUSH_CPU_MARKER("- Light Scatter", 0xFF, 0x00, 0x00);
         ScopedGPUTimer Timer(irr_driver->getGPUTimer(Q_LIGHTSCATTER));
         m_lighting_passes.renderLightsScatter(m_rtts->getDepthStencilTexture(),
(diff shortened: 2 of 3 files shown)


deveee:
On my computer in mansion track it's 260 fps vs 250 fps (GLES renderer). I will try on my tablet later. But indeed the difference in performance between 2 and 3 graphics level is large, so I like the idea even if it helps a little. Not sure if it's possible to combine it with already existing graphics settings, because there is already a lot of graphics options.

deveee:
Actually you can get more fps just by using less steps in light scatter shader. Visually there is almost no difference (far lights are slightly more visible).

zrzut ekranu z 2018-12-23 21-44-48
zrzut ekranu z 2018-12-23 21-45-00
zrzut ekranu z 2018-12-23 21-47-00
zrzut ekranu z 2018-12-23 21-47-15

Are you ok with using 8 steps instead of 16? If not, then I can use it only for lower graphics levels, but more work little gain.

```
diff --git a/data/shaders/pointlightscatter.frag b/data/shaders/pointlightscatter.frag
index a50f68fdc..2b82125a8 100644
--- a/data/shaders/pointlightscatter.frag
+++ b/data/shaders/pointlightscatter.frag
@@ -28,14 +28,14 @@ void main()
     vec3 closestpoint = - eyedir * (dot(-eyedir, light_pos) - radius);
     if (closestpoint.z < 1.) closestpoint = vec3(0.);
 
-    float stepsize = length(farthestpoint - closestpoint) / 16.;
+    float stepsize = length(farthestpoint - closestpoint) / 8.;
     vec3 fog = vec3(0.);
     vec3 xpos = farthestpoint;
 
-    for (int i = 0; i < 16; i++)
+    for (int i = 0; i < 8; i++)
     {
         float d = distance(light_pos, xpos);
-        float l = (16. - float(i)) * stepsize;
+        float l = (8. - float(i)) * stepsize;
         float att = energy * 20. / (1. + d * d);
         att *= max((radius - d) / radius, 0.);
         fog += density * light_col * att * exp(- density * d) * exp(- density * l) * stepsize;
```

Partmedia:
To clarify: you're suggesting to globally go from 16 to 8 steps? That definitely improves it for me, but my machine still needs light scattering disabled altogether to maintain about ~20 FPS on some maps. Maybe you'll want some comments from people who are actually able to run light scattering.

auriamg:
@deveee Since I am pretty much unable to tell the difference between 8 and 16, (and what little difference there is doesn't look worse, just slightly different), I would agree to use 8 by default. no use in using cycles for something that's essentially invisible

deveee:
Actually I made a bit more testing and 16 steps is minimum to avoid light flickering in oldmine. Or maybe that flickering is caused by something else, but I can't tell what is the reason.

Alayan-stk-2:
I think it would be good to have this made accessible in the graphics option, and likely disabled by default at level 3, so that people just using the presets have a better middle balance available than now.

qwertychouskie:
Agreed.  Also IMO the default level should be 4 since the current version has some optimizations.

PR #3666, opened 2018-12-30
small improvements regarding widget collapsing

Mrxx99:
small improvements regarding widget collapsing


deveee:
Minor thing, but all other variables in this class are initialized in .cpp, so I suggest to do the same to avoid a mess.

PR #3667, opened 2018-12-30
Using a default music if specified music of track was not found

Mrxx99:
Some addon tracks have set a music in their track.xml that can't be found (either because it was removed from the game in the meantime or the author forgot to include that file). When playing these tracks no music is played at all, which just doesn't feel right.

With this pull request a default music will be played in such cases. The default music can be specified in the stk_config.xml where also the title music is specified.

I chose "kart_grand_prix.music" as default because it (at least for me) always felt like the standard song and I think it just fits almost every possible track well. But that can be easily changed.

### DIFF
--- a/data/stk_config.xml
+++ b/data/stk_config.xml
@@ -132,7 +132,7 @@
            solver-mode=""/>
 
   <!-- The title music. -->
-  <music title="main_theme.music"/>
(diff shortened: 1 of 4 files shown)


auriamg:
Looks good, thanks, merged

PR #3679, opened 2019-01-04
Fix bad file reference in README.md

MarioCMFlys:
Correct a reference to a non-existing license file

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ To run SuperTuxKart, make sure that your computer's specifications are equal or 
 * Ideally, you'll want a joystick with at least 6 buttons.
 
 ## License
-The software is released under the GNU General Public License (GPL) which can be found in the file [`COPYING`](/COPYING) in the same directory as this file. Information about the licenses for the artwork is contained in `data/licenses`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ To run SuperTuxKart, make sure that your computer's specifications are equal or
 * Ideally, you'll want a joystick with at least 6 buttons.
 
 ## License
-The software is released under the GNU General Public License (GPL) which can be found in the file [`COPYING`](/COPYING) in the same directory as this file. Information about the licenses for the artwork is contained in `data/licenses`.
+The software is released under the GNU General Public License (GPL) which can be found in the file [`COPYING`](/COPYING) in the same directory as this file. Information about the licenses for the artwork is contained in [`data/CREDITS`](/data/CREDITS).
```
auriamg: This is not actually correct, because the credits do not name any licenses. The original text is correct, licenses can be found in data/licenses, it's just that there may be some confusion because this is talking about the assets data folder (i.e. SVN) and not the code data folder (git)

auriamg: However, when we create a release, svn and git are combined, which means the file will then exist

PR #3680, opened 2019-01-04
A couple of Android-related changes, some misc as well

TacoTheDank:
- Bump compileSdk 26 -> 28
- Android Studio 2.2.3 -> 3.2.1
- Replace an expansion with the result
- Fix some spelling
- Reformat some code
- Remove some unnecessary parentheses


deveee:
Please don't mix unrelated things and please don't mix them with typo fixes and whitespaces modifications. 

I didn't test it, but you changed ```android:icon="@drawable/icon"``` to ```android:icon="@mipmap/icon"```, so that it will try to use "res/mipmap" directory while in make.sh we create "res/drawable" for icons. 

And for gradle plugin (again untested) if I understand correctly 3.2.1 version will require gradle >= 4.4 while in ubuntu 18.04 I still have only 3.4.1. Does it fix something for you?

Target SDK 26 is enough so that google won't complain and it's properly tested. Maybe we can change it after 0.10 release.

TacoTheDank:
This was a lot of unrelated changes (which I realized some time after opening that it was entirely pointless), so I'll close this. I will state the reasons for the changes you mentioned, though.

- Changing drawable to mipmap (from what I remember) allows for adaptable icons (or maybe that's entirely backwards, I'm not sure).
- Updating Android Studio would require 4.6, and it usually is good to use the latest stable version.
- I'm currently on Oreo 8.1, and I am going to get the Pie update sometime this month, so I'm available to test higher targeted APKs if you need testers.

PR #3683, opened 2019-01-06
Show the name of who made the fastest lap

Mrxx99:
olivermath-2019 01 06_12 17 14

This shows who made the fastest lap in the race result screen.


Alayan-stk-2:
Thanks !

PR #3690, opened 2019-01-07
Update config to use the new online servers & remove hardcoded certificate

leyyin:
This is most likely in the wrong branch, feel free to merge it somewhere else.
As discussed with @hiker we decided to move to a more up to date VPS and to use Let's encrypt certificates.

New online/addons location: https://online.supertuxkart.net
New hardware stats location (TEMPORARY): https://new-stats.supertuxkart.net/ 
After the old VPS is disabled it will be moved to stats.supertuxkart.net (old VPS uses it for now).

NOTE:
- The data in the new VPS is from 6th January 2019. When the old VPS is shutdown I will reimport all the new data again.
- There is no syncing between addons.supertuxkart.net and online.supertuxkart.net (they are two separate islands)


qwertychouskie:
So will the new VPS also host at addons.supertuxkart.net with the old cert for compatibility with old STK versions?

auriamg:
@leyyin I am experiencing issues with the new server, I'm unable to log in... Will probably rollback the pull request until this is resolved

```
[info   ] HTTPRequest: Sending username=Auria&password=*********&save-session=true to https://online.supertuxkart.net/api/v2/user/connect/
[error  ] XMLRequest::afterOperation: curl_easy_perform() failed: Peer certificate cannot be authenticated with given CA certificates
```

qwertychouskie:
Was that error with the compiled version, or an old version with only config.xml changed?

auriamg:
@qwertychouskie That was with the latest version compiled AND a fresh config file

qwertychouskie:
Oddly enough, updating the config.xml file with the new servers but not using the new code is working?

leyyin:
@auriamg Can I assume you are on a mac?
Maybe because of this line https://github.com/leyyin/stk-code/blob/9e79269dfb30a3021b20796725a4523bbf9ccd51/src/online/http_request.cpp#L189 I was going to ask why is it disabled on apple devices.

auriamg:
@leyyin No it was on a windows 10 computer

leyyin:
I think I managed to fix it, it works for me in linux and in windows 10, you can check my branch now. Also can you please also test in OS X :).

auriamg:
@leyyin I can help test on mac, just not right now as I'm busy with other stuff (like building the beta). Maybe open a new PR so we don't forget?

PR #3692, opened 2019-01-09
Fix build on FreeBSD

Partmedia:


deveee:
The ```linux/input.h``` is needed for key names, so it should be outside of ```#if defined _IRR_COMPILE_WITH_JOYSTICK_EVENTS_```.

Partmedia:
You're right. How does this look?

deveee:
Merged, thanks!

PR #3703, opened 2019-01-13
Update config to use the new online servers & remove hardcoded certificate

leyyin:
This is the fixed version of https://github.com/supertuxkart/stk-code/pull/3690
I fixed it by adding the The Mozilla CA certificate store.

New online/addons location: https://online.supertuxkart.net
New hardware stats location (TEMPORARY): https://new-stats.supertuxkart.net/ 
After the old VPS is disabled it will be moved to stats.supertuxkart.net (old VPS uses it for now).

CA Bundle with all the root certificates is from here: https://curl.haxx.se/docs/caextract.html
It contains the public keys for all major certificate authorities, it should be good for about ~10-15 years.

NOTE:
- The data in the new VPS is from 6th January 2019. When the old VPS is shutdown I will reimport all the new data again.
- There is no syncing between addons.supertuxkart.net and online.supertuxkart.net (they are two separate islands)

If someone with a mac can test this it would be awesome.


src/config/user_config.hpp
```diff
@@ -978,7 +979,7 @@ namespace UserConfigParams
                                           "Everything related to online play.") );
 
     PARAM_PREFIX StringUserConfigParam      m_server_multiplayer
-            PARAM_DEFAULT( StringUserConfigParam(   "https://addons.supertuxkart.net/api/",
+            PARAM_DEFAULT( StringUserConfigParam(   "https://online.supertuxkart.net/api/",
```
auriamg: This is problematic because it only changes the *default* value. If the user already has a config file with the previous value, it will not update. Same for the other similar config keys. We need a way to make sure everyone is using the latest server.

A quick and dirty fix, I guess, would be to just rename the config keys. There may be other better ways?

qwertychouskie: IMO these should be in `stk_config.xml`, not the user config.  See e.g. the Donate URL.

Benau:
when u change config directory to 0.10 (without -git) it will be fixed anyway

deveee:
Not really, because we don't import settings from old profile, people may copy files from old directory manually and will wonder why it doesn't work. I would copy at least input.xml myself to avoid reconfigure everything.

And I think stk_config.xml has a sense, because typically people shouldn't modify these links. And if it's really needed, then it's still easy to edit text file.

auriamg:
Also, I believe the beta, already released, will place its config files in the 0.10 config folder, so people who played the beta would not get the new url

leyyin:
@hiker @auriamg I moved  the config into `stk_config.xml` please review my updated changes.

auriamg:
Thank you a lot for your work! I've also done a few tests and everything seems to be working properly

PR #3708, opened 2019-01-13
FFA for offline play

Mrxx99:
This pull request enables to play the Free-For-All mode in offline single- and multiplayer.
It can also be played with AI (thanks to @Benau).

The Free-For-All mode is an option in the current 3 strikes battle mode which got renamed to just battle.
In the Free-For-All mode the user can choose the duration but no hit limit because there is no space for another spinner.


Alayan-stk-2:
Didn't test yet, but the code looks ok.

src/modes/free_for_all.cpp
```diff
@@ -186,9 +188,12 @@ bool FreeForAll::isRaceOver()
...
     if (!getKartAtPosition(1))
         return false;
-    int top_id = getKartAtPosition(1)->getWorldKartId();
+
+    const int top_id = getKartAtPosition(1)->getWorldKartId();
+    const int hit_capture_limit = race_manager->getHitCaptureLimit();
+
     return (m_count_down_reached_zero && race_manager->hasTimeTarget()) ||
-        m_scores[top_id] >= race_manager->getHitCaptureLimit();
+        (hit_capture_limit != 0 && m_scores[top_id] >= hit_capture_limit);
```
auriamg: I wonder in which case hit_capture_limit can be 0 ?

src/states_screens/race_gui.cpp
```diff
@@ -1164,7 +1164,8 @@ void RaceGUI::drawLap(const AbstractKart* kart,
 
     static video::SColor color = video::SColor(255, 255, 255, 255);
     int hit_capture_limit =
-        race_manager->getHitCaptureLimit() != std::numeric_limits<int>::max()
+        (race_manager->getHitCaptureLimit() != std::numeric_limits<int>::max()
+         && race_manager->getHitCaptureLimit() != 0)
```
auriamg: I wonder in which case hit_capture_limit can be 0 ?

src/states_screens/track_info_screen.cpp
```diff
@@ -419,28 +447,35 @@ void TrackInfoScreen::onEnterPressedInternal()
...
-        UserConfigParams::m_num_karts_per_gamemode[race_manager->getMinorMode()] = local_players + num_ai;
-    }
-
     const int selected_target_type = m_target_type_spinner->getValue();
     const int selected_target_value = m_target_value_spinner->getValue();
 
-    if (race_manager->isSoccerMode())
+    const bool enable_ffa = m_show_ffa_spinner && selected_target_type != 0;
+
+	if (enable_ffa)
```
auriamg: Be careful, you are using tabs around here to indent. Please indent everything with spaces

src/states_screens/track_info_screen.cpp
```diff
@@ -463,32 +498,57 @@ void TrackInfoScreen::eventCallback(Widget* widget, const std::string& name,
...
         }
-        else
+        else if (m_show_ffa_spinner)
         {
-            m_target_value_label->setText(_("Number of goals to win"), false);
-            m_target_value_spinner->setValue(UserConfigParams::m_num_goals);
+            const bool enable_ffa = target_value != 0;
+            UserConfigParams::m_use_ffa_mode = enable_ffa;
 
+            if (enable_ffa)
```
auriamg: Minor: this could be simplified into

m_target_value_label->setVisible(enable_ffa);
m_target_value_spinner->setVisible(enable_ffa);

src/states_screens/track_info_screen.hpp
```diff
@@ -46,6 +46,10 @@ class TrackInfoScreen : public GUIEngine::Screen,
 
     bool m_record_this_race;
 
+    bool m_is_soccer;
```
auriamg: These new variables should be initialized with a default value in the constructor, yet unless I missed something that does not seem to be the case

src/modes/free_for_all.cpp
Mrxx99: Now you can set the hit capture limit to zero to only use a time limit, like it is with soccer for the goals.
For offline play I set the hit capture limit to zero

src/states_screens/race_gui.cpp
Mrxx99: The comment before, here used to not show tje trophy symbol that is used to display the hit capture limit if the limit is 0

src/states_screens/track_info_screen.cpp
Mrxx99: Oh sorry, I don't know where a tab came from I set my IDE to use spaces. Will check everything again for this

auriamg:
Thanks, this looks good to me. However, we have started the release process and this PR is coming a bit late as we now try to keep master stable leading up to the release. Unless @hiker agrees to an exception, this PR will likely be merged after the current release is completed

qwertychouskie:
AFAIK there hasn't been a feature freeze yet?

Alayan-stk-2:
The feature freeze wasn't discussed within the team, and I don't think it's a good idea. Features got added to the beta until the last day, we still have time here before a RC.

A feature like FFA for offline play help to polish the release imho and shouldn't be delayed by months until the next point release.

Also, Ravenbridge Mansion has several issues needing to be addressed before 0.10, so I don't think delaying the merge of a feature like this will help to get the version out sooner.

hiker:
Yes, we declared the feature freeze internally, so please don't merge this to master for now. Or create a 'dev' (or whatever) branch - but that would mean that we need to take care that we merge all changes for the release into 'dev'.

Thanks for your understanding!

Alayan-stk-2:
@hiker I have not been very available these last days, so I may have missed something, but I have not been involved in any discussion about a feature freeze ? I saw the "pretty much declared a feature freeze" in the blog post, but that's it, and that itself wasn't discussed ?

We will still have a good bunch of issues to resolve before 0.10 and will have to go through a RC, what's the point of a feature freeze ? I would not want to have something merged which affect e.g. physics because it may require to redo standard replays or may introduce a risky regression.

So when it comes to features, I think we should make a difference between the kind of changes which may induce significant additional work (for polish, bugfix or regression fixes), and those which won't. I think this PR, which is about converting an existing and well-tested mode to work in offline mode with an AI slightly changed from another existing and well-tested mode (3-strikes battle), fall in the second category.

qwertychouskie:
Agreed, IMO should be in 0.10.

qwertychouskie:
Since 0.10 has it's own branch, can we merge this to master and discuss later whether to include it in 0.10?

Issue #3344, opened 2018-07-15
Multiple GPLv3 violations


qwertychouskie:
(See https://github.com/supertuxkart/stk-code/pull/2647 for the original discussion.)

From https://www.gnu.org/licenses/gpl-faq.en.html#WindowsRuntimeAndGPL:

>You may link your program to these libraries, and distribute the compiled program to others. When you do this, the runtime libraries are “System Libraries” as GPLv3 defines them. That means that you don't need to worry about including their source code with the program's Corresponding Source. GPLv2 provides a similar exception in section 3.
>
> You may not distribute these libraries in compiled DLL form with the program. To prevent unscrupulous distributors from trying to use the System Library exception as a loophole, the GPL says that libraries can only qualify as System Libraries as long as they're not distributed with the program itself. If you distribute the DLLs with the program, they won't be eligible for this exception anymore; then the only way to comply with the GPL would be to provide their source code, which you are unable to do.

It seems linking against the versioned DLL is a violation of the GPL, as they are not technically a part of the OS.

There are a few potential options here:

1. Find a way to force VS to link against the unversioned DLL, which comes with Windows.  This may be hard because it is officially frowned upon. See https://stackoverflow.com/questions/10166412/how-to-link-against-msvcrt-dll-instead-of-msvcr100-dll-in-vc-10-0 and https://blogs.msdn.microsoft.com/oldnewthing/20140411-00/?p=1273

2. Only build with MinGW, which links against the unversioned DLL (see https://news.ycombinator.com/item?id=11964589).

3. Use the LGPL-licensed DLL from Wine/ReactOS: <http://source.winehq.org/git/wine.git/tree/HEAD:/dlls/msvcrt>  As long as STK has a copy of the DLL's code in the source, everything should be legal.

4. Something else that I'm not thinking of ;)

qwertychouskie:
Uh oh, OpenSSL is also a problem: https://www.gnu.org/licenses/license-list.en.html#OpenSSL

qwertychouskie:
I just found https://www.openssl.org/blog/blog/2017/03/22/license/ but this seems to have never been completed (the `LICENSE` file in the latest source is still the old one).

hiker:
According to:
https://license.openssl.org/trying-to-find
it looks like they have found all contributors - so maybe they are getting close, and openssl becomes a none issue in the very near future??

deveee:
[...]

PR #3715, opened 2019-01-17
Allow crash dialog for MinGW build

qwertychouskie:


qwertychouskie:
capture

This was on a RelWithDebInfo cross-compiled MinGW 64-bit test build.  `cv2pdb -C  supertuxkart.exe` was used to make the PDB.

qwertychouskie:
The only sorta-problem I had was that it took about a minute after STK freezing for the crash dialog to pop up.

qwertychouskie:
Any thoughts on this?  The main STK builds need to be switched to MinGW builds because of https://github.com/supertuxkart/stk-code/issues/3344, and it would be nice to still have crash dialogs.

qwertychouskie:
Here's a test build: https://jacobspctuneup.tk/STK/stk-code-mingw-crash-handling.zip

In this build LinkHelper::openURL is rigged to crash, the donate button in the credits works for triggering it.

@deveee You have a MinGW setup, right?  Any thoughts on this PR?

deveee:
I wonder if handler for memory allocation error should be totally disabled. Because in this case we should free some memory and not alloc more to display window with crash report. Especially that if there are some memory leaks or not enough ram, then it will probably crash in random place anyway. And I'm not sure if VS function is smart enough, so that it will throw exception.

There is std::set_new_handler() that works with mingw, but they explicitly say that you must free some resources or abort app.

And if everything else works, then I will merge this evening.

deveee:
Tbh. I'm not convinced... Because mingw is typically used for self-compilation. And you need additional external tools like cv2pdb to create pdb file. So that most people will just get wrong crash reports.

qwertychouskie:
The reason for this is https://github.com/supertuxkart/stk-code/issues/3344#issuecomment-455364398.  Basically for the official Windows builds we either have to switch to MinGW builds or direct users to MS's website to download MSVCRT (which would be very bad UX).  Or maybe it's possible to hex edit the MSVC-compiled binary to point to the unversioned MSVCRT dll?

qwertychouskie:
I just force-pushed to remove a small string change that was unneeded, since STK is now in string freeze.  Since 0.10 may be built with MinGW (https://github.com/supertuxkart/stk-code/issues/3344#issuecomment-476488642), it would be good to have this merged.

Also, the diff shows some lines as "changed" that aren't, not sure why. (line ending weirdness?)  Anyways, it won't harm anything.

Benau:
i don't feel like merging pr with fixme

Benau:
The slowness probably because the pdb file generated is not as good as msvc

So we either keep current way or use mingw build with accepting the slowness

deveee:
For mingw you don't need pdb files. You should be able to do the same as on linux - build with debug symbols, strip symbols for the release package, and then addr2line to decode the backtrace.

qwertychouskie:
Closing this PR as it is obsoleted by https://github.com/supertuxkart/stk-code/commit/c710057db1efd7f71f4f8de52368234f8ef25890.

Issue #3691, opened 2019-01-09
More server states required


hiker:
Long term we want to support the following 5 server 'states' (states might not be the best word):
- official, ranked: a server we run, which is ranked
- official, unranked: a server which we run, which is not ranked
- ranked: a trusted server not provided by us that is ranked
- trusted: a truster server, i.e. a server we believe has not been patched 
- none (default): none of the above.

The support for 'official' server is mostly long term, when we add new states it might just be easiest to add it all at the same time.

We should avoid the term 'official' for all ranked server (which we have atm afaik), since we can't really do anything if a server run by someone else has a problem.

qwertychouskie:
Maybe replace "none" with "community". (sounds nicer :)

jpenguin:
If server has
    <server-configurable value="true" />

It would be nice to mention that in the s4rver list

jpenguin:
If you call the servers run by core STK devs "trusted" it kinda implies that all other servers are hacked.  Maybe just Official/Volunteer/Modded

jpenguin:
I agree, I'm saying the average steam user won't realize that

hiker:
With #3745 waiting to be merged that fixes the name the user can see, the rest of this can be postponed till the next release.

v1993:
Now server owners are:

1. `-`
2. `Trusted`
3. `cddepppp256server`

I think this looks pretty confising because

1. Most servers have no owners
2. `Trusted` is not an owner, it is state

Benau:
option 3, we use ☆★STK★☆ (like ghost replay) for the owner field, and change the help to "Some of them are recommended servers with optionally ranked races."

this is less "weird" and translator can translate better.

and i have no intention to distinguish so-called official or trusted, esp there will never be pay-to-play ranked servers, or let people who only donate can join official servers is something that you want

Benau:
@jpenguin we update the server code now it can listen to ipv6 address, have time to update play stk? (Visit irc for details)

jpenguin:
I think I'll have time today, thanks.

Benau:
not sure @jpenguin can update his servers for 1.1 (and addons list!)... (btw 1 playstk server seem to be gone, do you have coredumpctl gdb supertuxkart if it crashed?)

also do u see stk dual license mail.... ![](https://lihkg.com/assets/faces/pig/dance.gif)

jpenguin:
I did get the license mail.  I just got home, I'll see about updating

[...]

PR #3745, opened 2019-02-07
For now rename 'official' servers to 'trusted'

hiker:
... to make it more obvious what the STK team does not operate those servers.

This is a fix for #3691 good enough for the next release, when a proper solution can be implemented. 

Note that some strings modified in this pull request are temporary only (i.e. to make sure we keep already translated strings that we will need later). 

@Benau, can you check that I didn't break anything (especially I found it hard to understand that m_server_owner_lower_case_name was "Official" (capitalised), I used all lower case now, please check that I didn't break anything ;) ).

### DIFF
--- a/src/network/server.cpp
+++ b/src/network/server.cpp
@@ -114,7 +114,6 @@ Server::Server(const XMLNode& server_info) : m_supports_encrytion(true)
         // not operated by us). #3691 will introduce the various states
         // required to properly distinguish between trusted, ranked,
         // and official servers.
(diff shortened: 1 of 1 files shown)


Benau:
// (currently) 'official' servers 'trusted', since we 
        m_server_owner_name = _("Trusted");

since we = ?

Alayan-stk-2:
@hiker iirc, the lower case name is used for search and sorting ? It may behave weirdly with translations as is.

Benau:
alayan no irc!

Issue #3378, opened 2018-08-09
Make sure we are GDPR-compliant


Arthur-D:
We should make sure we are GDPR-compliant, as that should help us avoid legal repercussions or at the very least drama resulting from not following them as best as we can.

Our current [Terms of Service](https://supertuxkart.net/Terms) and [Privacy Policy](https://supertuxkart.net/Privacy) were written a while ago and it would in any case be prudent to look at them again in time for any wider release of networked multiplayer.

- [ ] Find out exactly what data we are and will be collecting on people having an account on [addons.supertuxkart.net](https://addons.supertuxkart.net), people simply playing the game and people visiting or using our websites. This may include things like the wiki accounts on the main site, or other things not originally envisioned when we wrote our current terms and privacy policy.

- [ ] Try as best as we can to make sure we cover the relevant sections of [the GDPR](https://gdpr-info.eu). Especially important as it relates to giving explicit consent, us being able to prove we have such consent, withdrawal of consent, and people under the age of 16 since we need explicit consent by his or her parents.

- [ ] Make sure our terms and privacy policy is as easy to find and as clear as possible.

qwertychouskie:
Also check the whole "Donators" list in the credits and how that might be affected.

Wuzzy2:
This is possibly relevant:

If you tick on the checkbox “Connect to the Internet”, a checkbox appears with the text “Collect hardware statistics”, which is ticked ON by default. There is no indication that this checkbox will appear, and enables itself automatically.
The only warning is at the very first start of STK, but this text can be easily forgotten.

This is opt-out, i.e. wrong.

This checkbox should be OFF by default after you enable Internet access.

hiker:
STK should be all opt-in-only, i.e. all internet and hw stats option default to off, and user needs to opt-in (after #3748 is merged). Note that I left chat enabled by default, since imho the user actively opts-in by typing a message and pressing return.

I've updated our privacy policy to include details about the use of stun servers, and the handling of donators (also added to the donation page).

I think we still need to update the wiki page, and how it uses cookies (though we have a section about cookies in the privacy policy). This outstanding work seems not to be directly required for the next release, so postponing this till later.

MinIsMin:
Also you really need to mention the usage of piwik on supertuxkart.net and request consent as the current usage clearly violates GDPR.

See https://matomo.org/blog/2018/04/how-to-make-matomo-gdpr-compliant-in-12-steps/ for more info.

[...]

Issue #2549, opened 2016-07-02
More options in first run dialog box


qwertychouskie:
Right now, in the first run dialog, you only have the option to either turn on both internet and statistics sharing, or turn them both off.  It would be nice to be able to control them separatly from the dialog, similar to VLC:

![image](http://www.lochnessproductions.com/software/vlcwarper/vlw_install_scr01.png)

PR #3748, opened 2019-02-08
Gdpr compliance

hiker:
First fix for #3378: sending hardware stats is now opt-in (no default to send hardware stats anymore). Fixes also incorrect help message and incorrect status display if hw-stats are reported or not, and made the chat option (in)visible same as hw-stats.

Note that the order of entries in the 'internet' option is strange (the 'always show login' option is not strictly an internet-only option, since it also affects local-only accounts), but I am not sure why this order was chosen, so I left this unchanged.

Combining internet-access and hw-stats with just one option is bad now that stk is gdpr compliant, since it must now default to internet-off :( But I can't see an easy and convenient location - best option imho: remove the HW statistic, since we are actually not using the stats anymore.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1814,7 +1814,7 @@ void askForInternetPermission()
         "Please read our privacy policy at http://privacy.supertuxkart.net. "
         "Would you like this feature to be enabled? (To change this setting "
         "at a later time, go to options, select tab "
-        "'User Interface', and edit \"Connect to the "
+        "'General', and edit \"Connect to the "
         "Internet\" and \"Send anonymous HW statistics\")."),
         MessageDialog::MESSAGE_DIALOG_YESNO,
         new ConfirmServer(), true, true, 0.7f, 0.7f);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1807,7 +1807,7 @@ void askForInternetPermission()
         }   // onCancel
     };   // ConfirmServer
 
-    GUIEngine::ModalDialog *dialog =
+    MessageDialog *dialog =
     new MessageDialog(_("SuperTuxKart may connect to a server "
         "to download add-ons and notify you of updates. We also collect "
         "anonymous hardware statistics to help with the development of STK. "
@@ -1818,6 +1818,10 @@ void askForInternetPermission()
         "Internet\" and \"Send anonymous HW statistics\")."),
         MessageDialog::MESSAGE_DIALOG_YESNO,
         new ConfirmServer(), true, true, 0.7f, 0.7f);
+
+    // Changes the default focus to be 'cancel', which is not
+    // GDPR compliant, see #3378
+    dialog->setFocusCancel();
     GUIEngine::DialogQueue::get()->pushDialog(dialog, false);
 }   // askForInternetPermission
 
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -1807,17 +1807,21 @@ void askForInternetPermission()
...
         "Please read our privacy policy at http://privacy.supertuxkart.net. "
         "Would you like this feature to be enabled? (To change this setting "
         "at a later time, go to options, select tab "
-        "'User Interface', and edit \"Connect to the "
+        "'General', and edit \"Connect to the "
         "Internet\" and \"Send anonymous HW statistics\")."),
         MessageDialog::MESSAGE_DIALOG_YESNO,
         new ConfirmServer(), true, true, 0.7f, 0.7f);
+
+    // Changes the default focus to be 'cancel', which is not
```
Alayan-stk-2: You meant that the current "ok" default focus is not compliant ?

qwertychouskie:
Also see https://github.com/supertuxkart/stk-code/issues/2549, this would fix the internet disabled by default issue by allowing independent control.

src/main.cpp
hiker: I believe so - gdpr  requires a deliberate choice by the user to opt-in. With 'ok' being the default, and assuming that just pressing enter is not a deliberate choice (given that opt-out requires a click and pressing enter). IANAL, and if anyone can point out a counter example I am happy to leave the default, I don't like this.
We could add e.g. internet-permission as part of the installation process, but that's a lot of work and only covers a part of our user base :(
TBH, if we agree that we don't need the HW stats (has anyone looked at them in the last two years??), taking this out would be a really good choice imho.

qwertychouskie: @leyyin had disabled stats viewing because it was OOMing the old VPS, but I think since then IIRC the VPS got more available RAM.  @leyyin Can you re-enable stats viewing?

deveee:
Merged. And I think that we don't need HW stats too.

src/main.cpp
Alayan-stk-2: @hiker The comment is wrong, it says the new default is not compliant

leyyin: Unfortunately I can't because the generation scrip is badly optimized it needs more than 1 GB of ram.

qwertychouskie: @leyyin I thought the VPS got 2GB RAM a little while back as the provider upgraded the plans?  Also is the script in a github repo so someone can work on it?  Stats are useless if we can't use them, this needs to be fixed.

Or maybe just buy a VPS with more RAM, AFAIK we're not short cash for servers.

Alayan-stk-2:
We don't make use of hw stats and can't access them because of server issues, but it could definitely come handy. For example, the share of old processors and old igpu used to run STK can help to guide hw requirements.

Also, some opt-in for general gameplay data collection would be useful so axing hw stats wouldn't make a lot of sense in that context imho.

src/main.cpp
leyyin: It got an upgrade to only 1 GB of ram.

As for the scripts, see:
- https://github.com/supertuxkart/stk-stats
- https://github.com/supertuxkart/stk-stats/issues/7

Buying more servers won't be a solution, fixing the code is.
Fix software not buy more hardware ;)

hiker:
@Alayan-stk-2 good catch with the comment, thanks. I've just pushed the fix to master.

The other discussion: I would recommend to put this in a separate issue to track the decision about HW stats if required :)

PR #3755, opened 2019-02-18
Made AppData and desktop entry translatable.

enz:
Currently, the AppData file of STK is not translated, which causes the common software installation GUIs on Linux (like GNOME Software) show the information about STK only in English. The desktop-entry file is also not really translated (it contains a few hard-coded translations but only for a small subset of the languages supported by STK).

This patch makes both files fully translatable, such that they can be translated on Transifex together with the other string resources.

* This adds a new optional build-time dependency on gettext. To avoid that developers on platforms that don't really use these files (like on Windows) need to install gettext, the merging of the translations into the files will be skipped if gettext was not found.
* I verified that running update_pot.sh and adding translations in the po files works as expected, but this patch does not yet update the pot or any po files.
* The hardcoded translations in the desktop entry were removed to avoid conflicts once translations from the po files are merged in.

I have tested this patch on Ubuntu 18.10. I would appreciate if developers using other platforms could check that it does not break anything. Please tell me if you have questions or want me to change something.


enz:
After seeing the CI build failures, I realized that gettext can handle AppData files only if an appropiate ITS file is installed in /usr/share/gettext, which requires the package appstream on Ubuntu/Fedora. I changed the condition for merging the translations from GETTEXT_FOUND to UNIX AND NOT APPLE (the desktop files are really only used on Unix anyway) and added a test if the ITS file is installed in CMakeLists.txt.

enz:
I'm not sure how to tell the CI on Travis that the package appstream is needed, so I changed the condition for translating the Appdata and desktop entry from UNIX AND NOT APPLE to GETTEXT_FOUND AND METAINFO_ITS.

Benau:
not sure if possible to use data/localized_name.txt as input!

enz:
@Benau Could you clarify what exactly you mean with your comment?

deveee:
Probably he meant that you duplicated languages in data/po/LINGUAS. Maybe it's possible to use data/localized_name.txt, especially that you use "REGEX REPLACE" anyway. And I even wonder if you really need gettext, because simple bash script could generate that file, but it's not my work :p

enz:
Yes, the LINGUAS is unfortunately needed by msgfmt. It could be auto-generated by simply listing all po files, no need to parse localized_name.txt. I can do that if it is wanted. It would require copying all the po files to the build directory because msgfmt expects LINGUAS in the same directory.

I don't understand your comment about needing gettext. It is needed to extract translatable strings from the XML-based AppData file and to merge the translations back in at build time. I doubt you can use a simple bash script for that. And update_pot.sh already depended on gettext anyway. So the only real new dependency is the appstream package, which provides the ITS rules for gettext. Having gettext/appstream at build time is an optional dependency. On Windows (or generally, if gettext/appstream are not found), it will simply not merge in the translations, which is not worse than the situation now.

Nadjaya:
I don’t know if it’s possible to use data/localized_name.txt as input

qwertychouskie:
@enz Why was this closed?  IMO having these translatable would be great.

enz:
I closed it because I will be too busy in the near future to maintain this patch and it hadn't been merged after such a long time. And honestly: after Nadjaya asked literally the exact same question that Benau asked some months ago and that I already had tried to answer, I see a communication problem here that is discouraging.

auriamg:
FYI "Nadjaya" is not a member of the STK team, and may actually be a spambot

enz:
I see. Anyway, I won't work on this anymore. If someone else wants to pick up the patch, feel free to reopen.

PR #3770, opened 2019-02-25
Improve warning

qwertychouskie:
It's more clear now that this is non-fatal.


auriamg:
As discussed on IRC, we should say something like "the game might still run" as opposed to "the game will still run" since in those cases we can't promise things will go well

auriamg:
While at it, perhaps add an // I18N: comment clarifying what %s will be

PR #3789, opened 2019-03-04
Feature/auto lo d computation

samuncle:
This simplify the process of creating maps by computing LoD distance automatically

The distance Is based on a few different parameters
1) the size of the object on the screen. The smaller it is the more likely it will be culled
2) the amount of objects in the map (in a desert with not many objects the lod will be reduced to prevent any popping)
3) the parameter of the user. If you set your geometry detail to low then the Lod will cull more aggressively
4) in the future I want to tweak it so if a map is very high in geometry it tries to cull first small objects

Currently it is fully automated. During my tests I didn't saw any major issues or any visible differences between the automatic and manual lod. I think it it's good enough to have it automated but if you find a scenario that should be manual let me know

### DIFF
--- a/src/graphics/lod_node.cpp
+++ b/src/graphics/lod_node.cpp
@@ -241,8 +241,11 @@ void LODNode::OnRegisterSceneNode()
     scene::ISceneNode::OnRegisterSceneNode();
 }
 
-void LODNode::autoComputeLevel()
+void LODNode::autoComputeLevel(float scale)
 {
+    printf("Scale2 %f\n", scale);
+    m_volume *= scale;
+    printf("Factor %f\n", m_volume);
 
     // This will be set based on the amount of objects in a scene.
     float agressivity = 1.0;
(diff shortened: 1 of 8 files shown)


src/graphics/lod_node.cpp
```diff
@@ -144,98 +153,82 @@ void LODNode::OnAnimate(u32 timeMs)
...
-            scene::IMesh* mesh;
-
-            if (m_nodes[0]->getType() == scene::ESNT_MESH)
+            if (i == m_current_level || i == m_previous_level)
             {
-                scene::IMeshSceneNode* node = (scene::IMeshSceneNode*)(m_nodes[0]);
-                mesh = node->getMesh();
+                m_nodes[i]->setVisible(true);
+                float size = m_timer / 20.f;
+                /*
```
Alayan-stk-2: Commented out code, should probably be removed ?

src/graphics/lod_node.cpp
```diff
@@ -249,8 +242,43 @@ void LODNode::OnRegisterSceneNode()
     scene::ISceneNode::OnRegisterSceneNode();
 }
 
+void LODNode::autoComputeLevel(float scale)
+{
+    m_volume *= scale;
```
Alayan-stk-2: Rename to m_area, here and elsewhere

src/graphics/lod_node.cpp
```diff
@@ -249,8 +242,43 @@ void LODNode::OnRegisterSceneNode()
...
+    if(UserConfigParams::m_geometry_level == 1) agressivity = 0.75;
+    if(UserConfigParams::m_geometry_level == 2) agressivity = 0.5;
+
+    // First we try to estimate how far away we need to draw
+    float max_draw = 0.0;
+    max_draw = sqrtf((0.5 * m_volume + 10) * 200) - 10;
+    // If the draw distance is too big we artificially reduce it
+    if(max_draw > 250)
+    {
+        max_draw = 250 + (max_draw * 0.05);
```
Alayan-stk-2: This formula will transform 251 into 250 + 0.05*251 = 263.55

You should tweak the formula to have a smooth transition.

For example : max_draw = 235 + (max_draw*0.06)

PR #3791, opened 2019-03-04
Fix >= 2k scaling

dumaosen:
Much better in 2k or 4k

### DIFF
--- a/src/font/font_with_face.cpp
+++ b/src/font/font_with_face.cpp
@@ -362,11 +362,16 @@ void FontWithFace::setDPI()
     {
         float scale = std::max(0, screen_width - 640) / 564.0f;
         
-        // fix >= 2k scaling
+        // fix 2k and 4k scaling
         if (screen_height > 1300)
             scale = std::max(0, screen_width - 800) / 650.0f;
         if (screen_height > 2000)
-            scale = std::max(0, screen_width - 900) / 750.0f;
+            scale = std::max(0, screen_width - 1050) / 850.0f;
+        // not sure if these are OK
+        if (screen_height > 2800)
+            scale = std::max(0, screen_width - 1300) / 1050.0f;
+        if (screen_height > 3800)
+            scale = std::max(0, screen_width - 1800) / 1500.0f;
         // attempt to compensate for small screens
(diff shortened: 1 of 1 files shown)


deveee:
Indeed looking at the screenshots it's better. I will try it later.

samuncle:
I have access to a 4k screen I can try :)

deveee:
@dumaosen What about something like this?
```
diff --git a/src/font/font_with_face.cpp b/src/font/font_with_face.cpp
index 2af164d36..179fe9d23 100644
--- a/src/font/font_with_face.cpp
+++ b/src/font/font_with_face.cpp
@@ -358,6 +358,11 @@ void FontWithFace::setDPI()
         float scale = screen_height / 480.0f;
         m_face_dpi = int(getScalingFactorTwo() * getScalingFactorOne() * scale);
     }
+    else if (screen_height > 1300)
+    {
+        float scale = screen_height / 480.0f * 0.45f;
+        m_face_dpi = int(getScalingFactorTwo() * (getScalingFactorOne() + 0.3f) * scale);
+    }
     else
     {
         float scale = std::max(0, screen_width - 640) / 564.0f;
```
In this way:
- title font has reasonable size
- normal font is slightly bigger, so it's better for high resolution screens that may have higher dpi too
- it's proportional to window size, so that it works even with huge resolutions

Or if you prefer your way as an user of 4k monitor, then I will merge. And for next-next release we need proper hidpi support, because screen resolutions won't become smaller in future...

dumaosen:
@deveee your way looks like this:

So still big font here...

qwertychouskie:
I think the speedometer needs to take the absolute font size into account rather than assuming that the font is always the same size, which it isn't.  @deveee could you look into this?

dumaosen:
Just merge it. Everything is good enough for me now.

Alayan-stk-2:
I don't like adding a bunch of ifs like this. Surely one generalized formula could be used instead.

deveee:
The rank number is not too big anymore and I eventually made a commit with my patch, because it's simpler and IMO slightly larger font is better (there is enough space for it anyway). Other issues on 4k screen are not directly related to font size.

PR #3798, opened 2019-03-08
Disable internet-dependent checkboxes rather than hiding them

qwertychouskie:
This feels much more natural.


Alayan-stk-2:
I assume the removal of the two labels variable is because with the disabling method, labels always remain active and don't need to be hidden ?

qwertychouskie:
@Alayan-stk-2 Correct.  This is the same way things are done in the graphics options dialog when the advanced pipeline is disabled.

UnMe0w:
noice

PR #3800, opened 2019-03-09
Fix and improve various scaling in high resolution

dumaosen:


deveee:
Music info looks ok as far as I tested. But fps rectangle is too small even on high resolutions:
Zrzut ekranu z 2019-03-09 22-03-04
You can get font height using ```font->getDimension(L"X").Height```, so that rectangle height could be computed. And maybe rectangle width should be 100% of screen width or maybe 90% with some small margins?

deveee:
I had a quick look and now in addons screen it's worse on 1280x720 resolution (too little space for updated/rating):
Zrzut ekranu z 2019-03-13 20-59-01

and fps bar is too small:
Zrzut ekranu z 2019-03-13 21-02-54

Btw. you can also see how it looks with hidpi_enabled="true" in config.xml, so it will work on android.

I'm also not convinced to ```w->setScrollSpeed(irr_driver->getActualScreenSize().Width / 128);```. IMO currently with fixed scroll speed it's easier to read text on different resolutions. What matters is how fast people can read, so maybe it should have consistent number of letters per second.

For music description I'm not sure why you mix fixed values with proportional size, but as far as I tested visually it looks ok.

For buttons size with 3000x2000 it's just few pixels and I barely see any difference. But it shouldn't break anything.

dumaosen:
I have found a better solution so I'll close it and open a new one.

PR #3804, opened 2019-03-11
Feature/raytracer reflexion

samuncle:
The feature add realtime space screen raycasted reflexion. It basically uses the normal + depth buffer to raycast in the screen and lookup the previous frame of the game to compute reflections.

The technique isn't perfect but it gives the illusion of reflecting the surrounding. If the ray doesn't find an intersection it will simply fallback to the skybox.

https://docs.unrealengine.com/en-us/Engine/Rendering/PostProcessEffects/ScreenSpaceReflection

* It's fully integrated with the current pipeline and nothing needs to be changed for assets to work with. As soon as you add a pbr roughness map it will use it.
* Sadly currently many of the ground textures aren't using roughness, so the only part where it's visible is the space ship.
* Many karts will benefit from this and reflect their environment, since they use the roughness/pbr shader

Performance wise it's a bit heavy (because of the ray tracing). Probably I would like to have it as an option so users can turn it off.

Feedback is welcome (especially on the shader side, I feel the quality could be improved).
reflective_kart

### DIFF
--- a/data/shaders/IBL.frag
+++ b/data/shaders/IBL.frag
@@ -29,6 +29,28 @@ float makeLinear(float f, float n, float z)
     return (2 * n) / (f + n - z * (f - n));
 }
 
+vec3 fastBlur(vec2 uv)
+{
+    vec4 sum = vec4(0.0);
+    float X = uv.x;
+    float Y = uv.y;
+    vec2 pixel = vec2(0.001);
+/*
+    sum += textureLod(albedo, vec2(X, Y - 3.0 * pixel.y),    1.0) * 0.03125;
+    sum += textureLod(albedo, vec2(X, Y - 1.3333 * pixel.y), 1.0) * 0.328125;
+    sum += textureLod(albedo, vec2(X, Y),                    1.0) * 0.273438;
+    sum += textureLod(albedo, vec2(X, Y + 1.3333 * pixel.y), 1.0) * 0.328125;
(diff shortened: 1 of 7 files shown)


qwertychouskie:
I posted this on IRC, but also posting here.

It seems the reflection has some sort of "memory" in that after turning and stopping, the reflection is offset:

Look at the reflection of the window line.  Looking closely at the image, it actually seems that the right part of the screen is offset left, the left part offset right, and the middle not offset.  Actually, looking at it again, it seems that the offset is not connected to the lag, the lag is a temporary offset when turning the kart and is also quite noticeable.

Also, the reflection on Tux's spoiler looks buggy.

All in all, I think the feature is great, but needs a bit more work.

Benau:
There is 
```
    linear_depth_framebuffer.bind();
    LinearizeDepthShader::getInstance()->render(depth_stencil_texture);
```
for u to get linear depth in PostProcessing::renderSSAO

loper12:
sounds pretty neat

Benau:
BTW sam seems that the 0.04 is changed so at least either we adjust the red channel for all pbr data or there is some fallback shader for old assets, before we can use this feature! (Adjust red channel seems more easier)

samuncle:
Okay, for now the gloss part is computed like in the legacy pipeline. Will be enabled with proper computation once assets are up to date!

PR #3813, opened 2019-03-29
Added documtation for CentOS/RHEL/Mageia/OpenSUSE

jpenguin:
Just added dependency instructions

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -47,7 +47,7 @@ OpenSUSE command:
 sudo zypper install gcc-c++ cmake openssl-devel libcurl-devel \
 freetype-devel fribidi-devel libogg-devel openal-soft-devel libpng-devel \
 libvorbis-devel libXrandr-devel pkgconf zlib-devel enet-devel glew-devel \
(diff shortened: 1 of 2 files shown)


Benau:
btw you can idle on our irc channel to know when to update your servers! (you are tuxpeng?)

jpenguin:
Thanks

Yes, I'm going to just keep and update my PLAYSTK servers (AWS), probably shutdown the others

Benau:
Deve wanna merge? Maybe not sure glu-devel is needed...

jpenguin:
It used to compile without it. When I tried to compile on Friday, it couldn't find glew without it

deveee:
I will look at this in next few days. Glu shouldn't be needed because we explicitly want OPENGL_gl_LIBRARY in cmake and our static package is compiled without it. But having it installed doesn't hurt anyway.

jpenguin:
I have opensuse tumbleweed installed on my computer, cmake kept saying it couldn't find glew- I made sure the development packages were installed, then I installed flu & it found glew

PR #3817, opened 2019-04-05
CMake: Add option to use system WiiUse

akien-mga:
Adapted from Fedora patch by @ignatenkobrain:
https://src.fedoraproject.org/rpms/supertuxkart/blob/be85787d3485c477d23419b19c382b2186020016/f/0002-Unbundle-WiiUse.patch

Tested successfully on Mageia 7 with `-DUSE_SYSTEM_WIIUSE=ON`, currently building with `=OFF` to confirm that the original system still works as expected.


deveee:
Thanks for the patch. I would make it off by default, because I don't have possibility to test different versions or debug if there will be some issues. And you have ```if(ENET_FOUND)```. Everything else seems ok.

akien-mga:
Good points, I'll make it opt-in and fix the wrong copy pasted `if` tomorrow.

akien-mga:
Made it opt-in and fixed the incorrect `ENET_FOUND`.

deveee:
Merged, thanks.

qwertychouskie:
@deveee Should this also go on the 0.10 branch?

PR #3860, opened 2019-04-20
Clear separation between information for readers/players and contribu…

marioortizmanero:
…tors


Alayan-stk-2:
These changes are really trivial.

Style enhancements, typo fixes, etc. are not bad, but one at a time is really too little.

PR #3863, opened 2019-04-21
Swedish spelling

Zitrax:
hårtvare -> hårdvaru


Alayan-stk-2:
Ask to join the swedish translation team at Transifex : https://www.transifex.com/supertuxkart/supertuxkart/

Translations are pulled from there, a manual change to the file would be overwritten later on.

PR #3885, opened 2019-04-28
Detect and use ccache

Hi-Angel:
One feature that meson does but CMake doesn't is [automatically detecting and using ccache](https://mesonbuild.com/Feature-autodetection.html). It's a great feature, and greatly speeds up recompilations that often caused by common development activity, such as rebasing branches, checking out stuff, moving commits around, etc. Let's fix this by wiring up ccache autodetection *(based on [this stackoverflow answer](https://stackoverflow.com/a/24305849/2388257#how-to-use-ccache-with-cmake))*.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,12 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "STKRelease")
 endif()
 
+find_program(CCACHE_FOUND "ccache")
+if(CCACHE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "ccache")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "ccache")
+endif(CCACHE_FOUND)
+
 option(SERVER_ONLY "Create a server only (i.e. no graphics or sound)" OFF)
 option(CHECK_ASSETS "Check if assets are installed in ../stk-assets" ON)
 option(USE_SYSTEM_ANGELSCRIPT "Use system angelscript instead of built-in angelscript. If you enable this option, make sure to use a compatible version." OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,8 +43,8 @@ endif()
 
 if((UNIX AND NOT APPLE) AND NOT SERVER_ONLY)
     option(ENABLE_WAYLAND_DEVICE "Enable Wayland device for linux build" ON)
-    
-    if((${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm") OR 
+
+    if((${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm") OR
        (${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64"))
         option(USE_GLES2 "Use OpenGL ES2 renderer" ON)
     else()
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -67,18 +67,18 @@ Compilation instruction is explained there. If you don't need this feature, pass
 ### Compiling
 
 To compile SuperTuxKart, run the following commands inside `stk-code` directory
- 
+
 ```bash
 # go into the stk-code directory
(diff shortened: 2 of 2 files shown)


Alayan-stk-2:
@deveee @Benau Does one of you intend to review this ?

deveee:
Tbh. I always use a script with something like
```
		export CC="ccache gcc"
		export CXX="ccache g++"
```
and I prefer that using ccache is user's decision rather than do it automatically. But anyway no strong opinion...

Hi-Angel:
@deveee but for this people need to remember that the particular build system supports variables, and to know about existence of ccache *(one could have occasionally installed it, e.g. while experimenting with stuff)*. Someone may have STK among dozens of other FOSS projects they sometimes work on *(or just compile, e.g. if a user simply likes having self-built STK)*, and just not to care enough to set the explicit variable.

It's just a pleasant trivia that improves user and developer experience.

Hi-Angel:
Btw, note that this is port of the according functional in Meson. And Meson nowadays is widely deployed, many core projects such as Mesa, XServer, libinput are using Meson, so you can say such autodetection feature is widely tested as well.

qwertychouskie:
If ccache speeds up builds and doesn't cause bugs, I vote to merge it.

CMakeLists.txt
```diff
@@ -18,6 +18,12 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "STKRelease")
 endif()
 
+find_program(CCACHE_FOUND "ccache")
+if(CCACHE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "ccache")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "ccache")
+endif(CCACHE_FOUND)
+
 option(SERVER_ONLY "Create a server only (i.e. no graphics or sound)" OFF)
```
Chujinda: CMakeLists.txt

james165527:
Thanks for the CODES dude !

*Yours sincerely,*
Mark James Agripo.

Hi-Angel:
Thanks for reviews everyone! Note, I don't have write access to the repository, could please someone press the merge button for me?

Benau:
With your way user with ccache installed can never disable it.

What about you just update install.md saying how to export ccache like deve said so everyone is happy?

Hi-Angel:
> With your way user with ccache installed can never disable it.

I can add an option to disable it, something like --no-ccache, would that be okay with you?
 
> What about you just update install.md saying how to export ccache like deve said so everyone is happy?

As I mentioned in [this comment](https://github.com/supertuxkart/stk-code/pull/3885#issuecomment-493255377) it's just pleasant trivia for developers and users who build the game. Presence of ccache is usually transparent, a should not affect build output, or anything, except the speed of rebuilds. But if you think there can be corner-cases, where someone does not want ccache, I can add an option to disable it.

Benau:
Because ccache is a you-know-what-you-are-doing thing.....

Hi-Angel:
> Because ccache is a you-know-what-you-are-doing thing.....

It's really not. As I mentioned at the top post, meson-build based systems all use such autodetection simply because that's how meson works. And right now there's quite a lot of them, because many projects migrated from older autotools to meson. Including core ones, like Mesa and xserver for example.

Hi-Angel:
@Benau for reference, here for example Mesa explicitly enable it in CI https://gitlab.freedesktop.org/mesa/mesa/merge_requests/240

dumaosen:
Just merge it at first. I think U can make it optional later.
I dont think it's good to make a good pr unmerged in a long time because it may cause conflict.

Hi-Angel:
FWIW, I had a similar code [merged to Konsole](https://invent.kde.org/kde/konsole/blob/ef13a09e0178317e40076905ef9cef6efbc99598/CMakeLists.txt#L20), except it's a bit more advanced: in particular there's an option to disable it. If there's an interest, I can update this MR with that version of code.

The relevant discussion https://invent.kde.org/kde/konsole/merge_requests/26

PR #3881, opened 2019-04-25
A better solution for those scaling issues in high or creepy resolutions 

dumaosen:
They are my fixes of the scaling issues appear in the game.
Here are some pictures:
stk1
stk2
stk3
stk4
stk5
stk6
stk7
stk8
stk9
And more!

PR #3910, opened 2019-05-12
A better solution for those scaling issues in high or creepy resolutions gen II

dumaosen:
Another https://github.com/supertuxkart/stk-code/pull/3881.
If it's pushed, it will be good if someone make a gui for font_size...

### DIFF
--- a/data/CREDITS
+++ b/data/CREDITS
@@ -93,6 +93,7 @@ Bug fixes & misc contributions
 - Ward Muylaert
 - Fantasmos
 - QwertyChouskie
+- Dumaosen
 
 = Visual art =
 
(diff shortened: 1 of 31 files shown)


Benau:
deve you are ok to remove m_hidpi? and can u check if it's ok on your android.....

deveee:
I had very quick look. The spacers size that depends on font size is good. Maybe it could even allow floating point values, so you could use eg. "0.5f" between two lines. Can be improved later.

And the "hidpi" is still there and it's even better, because you can choose font size in config. I would be more aggressive on android and use font size = 5. Atm. there is too little space for it in some places, but maybe it could be easily fixed.

And I'm not sure, but maybe font size should depend on desktop resolution? Because 800x600 in windowed mode has different physical size than 800x600 in fullscreen. Especially on your high-resolution monitor. Previously on low resolutions fonts were larger compared to window size. Now it's proportional in all resolutions.

deveee:
Merged, thanks!

I changed max font scale limit to 6 (it's only in config file, so there is no reason to rectrict it that much) and used 4.5 on android.

And maybe it should be added in GUI, so that user could choose best font size manually and it would fix the problem with windowed mode.

PR #3912, opened 2019-05-12
3780 scene optimisation

yang980:
มันเป็นของผม

### DIFF
--- a/lib/irrlicht/include/ISceneNode.h
+++ b/lib/irrlicht/include/ISceneNode.h
@@ -41,6 +41,8 @@ namespace scene
 	{
 	public:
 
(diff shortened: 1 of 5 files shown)


Benau:
Thanks for your submission.

It appears that you've created a pull request using one of our repository's branches. Since this is
almost always a mistake, we're going to go ahead and close this. If it was intentional, please
let us know what you were intending and we can see about reopening it.

Thanks again!

auriamg:
We previously tried similar optimisations, however we need to very vareful as they usually have unintended consequences. For instance, previous similar attempts broke LOD and split-screen, those would need to be tested very well

PR #3914, opened 2019-05-13
thai translate(th.po)

annopnod:
thai translate is not complete (work in process)

### DIFF
--- a/data/po/th.po
+++ b/data/po/th.po
@@ -1,0 +1,5546 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the supertuxkart package.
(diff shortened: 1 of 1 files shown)


Benau:
https://www.transifex.com/supertuxkart/supertuxkart/language/th/
Click supertuxkart.pot

And then click upload file

Benau:
you actually leave transifex thai team... why you are doing that?

Alayan-stk-2:
Translations are handled through STK's page at Transifex. Directly merged translation changes would be overwritten later on when translations are pulled from there.

annopnod:
> you actually leave transifex thai team... why you are doing that?

sorry i clicked wrong button.i rejoin it
if font not work use th-sarabun is public license

Benau:
not sure deve can readd annopnod as thai member instantly lol

PR #3917, opened 2019-05-14
GenIII for scaling issues fix

dumaosen:
They are minor

### DIFF
--- a/data/gui/screens/ghost_replay_selection.stkgui
+++ b/data/gui/screens/ghost_replay_selection.stkgui
@@ -13,7 +13,7 @@
             <list id="replay_list" x="0" y="0" width="100%" height="100%" alternate_bg="true"/>
         </box>
 
-        <tabs id="race_mode" height="1f" width="98%" align="center">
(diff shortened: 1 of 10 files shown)


deveee:
Don't you think that now some tabs with categories are too big? Eg. for font size 4 in ghost selection screen:
Zrzut ekranu z 2019-05-14 22-09-55
it's ~2x larger than all/standard/add-ons in tracks screen:
Zrzut ekranu z 2019-05-14 22-10-05

And you can try to make the lists looking reasonably fine with font size 5, because I can make hiscores in tracks info scrollable if not enough space and then I think that we can add it to GUI options.

Benau:
seems that server selection screen list are too large...

btw @auriamg how can I make the Server Location: xxx the top like the left column
text_align="top" doesn't work... (you need to test on server with players and game started to make it more visible)

Benau:
@dumaosen can u use     int row_height = GUIEngine::getFontHeight() * 4 / 2;
for server selection screen?

PR #3919, opened 2019-05-15
Add background

LeandroStanger:
The Background is used in the App Center of Endless OS
An example image
Captura de tela de 2019-05-11 20-02-07


Benau:
Any preview for stk icon in the App Center?

Alayan-stk-2:
Your code uses a github link in the apddata.xml, this is most definitely incorrect, the picture should be read from the local STK installation.

qwertychouskie:
@Alayan-stk-2 Are you sure?  Look at e.g. the screenshot entries, they point to `supertuxkart.net`.

Alayan-stk-2:
Pointing to supertuxkart.net might be correct (though I don't see why it's necessary at all if the pictures are shipped in the install), but towards github is certainly not what we want.

qwertychouskie:
It's necessary because the game is likely not installed at this point... 



And I don't see why linking to GitHub is worse than supertuxkart.net, but it could be easily changed.

Alayan-stk-2:
> It's necessary because the game is likely not installed at this point...

Mmm, that's a good point. I wasn't fully familiar with how this is used.

I don't know why the PR author closed it if it works as expected, though.

jpgqs:
> 这是必要的，因为此时可能没有安装游戏......
> 
> 
> 
> 我不明白为什么链接到GitHub比supertuxkart.net更糟糕，但它可以很容易地改变。

秀呀老哥good

AsciiWolf:
@LeandroStanger Feel free to send this as AppData file patch for the [STK Flatpak](https://github.com/flathub/net.supertuxkart.SuperTuxKart). Sending this upstream does not make much sense since this is only used in the Flatpak version.

PR #3933, opened 2019-05-21
Using xml.dom.minidom.parse is known to be vulnerable to XML attacks.

ghostslayer989:
changed xml to defusedxml

### DIFF
--- a/data/po/extract_strings_from_XML.py
+++ b/data/po/extract_strings_from_XML.py
@@ -86,7 +86,7 @@ for file in filenames:
         isAchievements = True
         
     try:
-        doc = xml.dom.minidom.parse(file)
+        doc = defusedxml.minidom.parse(file)
     except Exception as ex:
         print "============================================"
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
@ghostslayer989 This code is only used when we want to update the strings for translation. Not only is this never run by end users, it is not run against some external untrusted data, but against the data files that come along with the code and are just as trusted.

Any reason to nonetheless bother with this change ?

qwertychouskie:
I don't see any reason to not merge it...  (People copy code from weird places sometimes, if someone copies the code it should be secure code.)

Edm571:
@ghostslayer989 You have just been invited to creation club! Go to https://creation-club.github.io-org to get started.

ghostslayer989:
> @ghostslayer989 You have just been invited to creation club! Go to [creation-club.github.io-org](https://creation-club.github.io-org) to get started.

???

hamidnegahbanaghdami:
Hi

PR #3944, opened 2019-05-28
Use icons in general_text_field

dumaosen:
Screenshot_20190528_205304


dumaosen:
I think we should merge button and icon buttton and just render them differently. Like this:
<button> -> <button>
<icon_button> -> <button icon="true">

PR #3947, opened 2019-05-29
improved indentation

anishagg17:
some spaces were missing ,which mis-matched the entire pattern of indentation .So, I fixed it along with it i simplified a comparison which will make it easier for a user to understand.


Benau:
You should open this in upstream angelscript instead

kittlerbilly:
Thank you. Somehow I've got to learn this computer talk. I don't understand
a lot of what is directed. If love to be All In but reading domes makes me
feel undeserving! Thanks again. Reaper

PR #3949, opened 2019-06-02
UI fix bundle VII

dumaosen:
Sorry this time online editing since i used samuncle's rt branch (and i dont want to reset).
This time mostly dialogs

### DIFF
--- a/data/CREDITS
+++ b/data/CREDITS
@@ -78,6 +78,9 @@ Google Summer of Code Students
 - Lauri Kasanen
 - Daniel Butum
 - Bart Cools
+
+
+Google Summer of Code Students
 - Csaba Molnar
 - Sachith Seneviratne
 - Akshay Singhvi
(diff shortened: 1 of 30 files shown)


Benau:
Not sure alayan mimd maosen change credits lol

deveee:
Title in every credits screen is IMO good. One donations screen has 7 lines while other screens have 6, not sure if it's intended.

And at least challenge selection dialog is now worse for large font (much harder to click):
Zrzut ekranu z 2019-06-04 23-15-02
compared to previously:
Zrzut ekranu z 2019-06-04 23-17-10

And confirm dialog could have a bit larger buttons too.

dumaosen:
These are untested

deveee:
Assuming that the pull request is finished, the icons in challenges selection are still very small. If you think that there is too little space, maybe do something like this?
challenges

Everything else seems ok.

dumaosen:
> Assuming that the pull request is finished, the icons in challenges selection are still very small. If you think that there is too little space, maybe do something like this?
> challenges
> 
> Everything else seems ok.

Yes it is really good. But i think we should make the dialog smaller if so.

PR #3970, opened 2019-06-25
Feature/enable lod autocompute

samuncle:
Add the auto computation for LoD

The exporter can now be simplified (3d artists don't need to specify the distance anymore)

### DIFF
--- a/src/graphics/lod_node.cpp
+++ b/src/graphics/lod_node.cpp
@@ -249,6 +249,39 @@ void LODNode::OnRegisterSceneNode()
     scene::ISceneNode::OnRegisterSceneNode();
 }
 
+
+void LODNode::autoComputeLevel(float scale)
+{
+    m_area *= scale;
+
+    // Amount of details based on user's input
+    float agressivity = 1.0;
+    if(UserConfigParams::m_geometry_level == 0) agressivity = 1.25;
+    if(UserConfigParams::m_geometry_level == 1) agressivity = 1.0;
+    if(UserConfigParams::m_geometry_level == 2) agressivity = 0.75;
+
+    // First we try to estimate how far away we need to draw
+    float max_draw = 0.0;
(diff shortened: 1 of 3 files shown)


src/graphics/lod_node.cpp
```diff
@@ -249,8 +250,44 @@ void LODNode::OnRegisterSceneNode()
...
+
+    // Amount of details based on user's input
+    float agressivity = 1.0;
+    if(UserConfigParams::m_geometry_level == 0) agressivity = 1.25;
+    if(UserConfigParams::m_geometry_level == 1) agressivity = 1.0;
+    if(UserConfigParams::m_geometry_level == 2) agressivity = 0.75;
+
+    // First we try to estimate how far away we need to draw
+    float max_draw = 0.0;
+    max_draw = sqrtf((0.5 * m_area + 10) * 200) - 10;
```
auriamg: sqrt is known to be quite slow, is it possible to avoid calling it so often?

chogyangzhao:
is test

Chujinda:
ช่งยหน่อย

james165527:
I can't approved my system is block !
*Yours sincerely,*
Mark James Agripo.

james165527:
send later !
*Yours sincerely,*
Mark James Agripo.

src/graphics/lod_node.cpp
dumaosen: // First we try to ESTIMATE how far away we need to draw
It's just an estimate value. If u think it has a low performance, just change it instead of leaving an old, invalid PR.

qwertychouskie:
@Benau Could you take a look and see if this can be merged?  It would be rather nice for tracks like Shifting Sands.

qwertychouskie:
For the record, I tested this branch rebased on master about 1-2 weeks ago, and everything was working with observed no bugs.  (Also rebasing needed no manual intervention.)

PR #3983, opened 2019-07-01
More bottom bar color

dumaosen:


auriamg:
This PR affects 199 files, though most of them I checked seem to have no change to them. Surely there is a way to issue a PR that only changes the files that need changing?

(for the records, I'm also not convinced the bottom bar needs color changing)

Nadjaya:
No, leave it

Alayan-stk-2:
This PR wrongly change dozens of file permissions by making data files executable. It's unmerge-able with this issue.

I'm not sold either on the core idea of changing the bottom bar color (the coal theme uses a custom one mostly for contrast ; otherwise orange is the main STK color which matches the "SuperTuxKart" text of the main menu).

PR #3997, opened 2019-07-11
Fix AngelScript build on FreeBSD non-x86

myfreeweb:


Benau:
Do you think you can sumbit it to upstream angelscript also?

myfreeweb:
I don't see a pull request type thing, but the website mentions emailing the author with suggestions. Doing that now :)

myfreeweb:
update: this will be included upstream.

Benau:
good and merged

PR #4006, opened 2019-07-15
Limit sscanf %s fields to prevent possible crash.

Quipyowert2:
[Cppcheck 1.88](https://github.com/danmar/cppcheck) gives the following warnings about the `sscanf()` calls in the method `ReplayPlay::addReplayFile()`. I changed the sscanf calls from `%s` to `%1023s` because `s1` is a 1024 byte buffer and Cppcheck recommends to make the field width limit for `%s` one less than the buffer size to leave space for the null byte which ends a string. Without the field width limits in the format strings, STK probably won't crash unless the replay files are corrupted somehow (e.g. with `zzuf` or `afl` fuzzers).

Cppcheck warnings:
```
[replay\replay_play.cpp:145]: (warning) sscanf() without field width limits can crash with huge input data.
[replay\replay_play.cpp:165]: (warning) sscanf() without field width limits can crash with huge input data.
[replay\replay_play.cpp:227]: (warning) sscanf() without field width limits can crash with huge input data.
[replay\replay_play.cpp:241]: (warning) sscanf() without field width limits can crash with huge input data.
```

### DIFF
--- a/src/replay/replay_play.cpp
+++ b/src/replay/replay_play.cpp
@@ -142,7 +142,7 @@ bool ReplayPlay::addReplayFile(const std::string& fn, bool custom_replay, int ca
     if (version >= 4)
     {
         fgets(s, 1023, fd);
-        if(sscanf(s, "stk_version: %s", s1) != 1)
+        if(sscanf(s, "stk_version: %1023s", s1) != 1)
         {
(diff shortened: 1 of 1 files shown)


Quipyowert2:
Force pushed to add a limit to the `%[^\n]` field also.

PR #4025, opened 2019-07-26
Disable story mode if assets is non-full

dumaosen:
I'd like to give user a chance to reinstall the assets when click the button.


dumaosen:
As Benau says, overworld is a good playground for newbies.
The PR is old now. Close it.

PR #4027, opened 2019-07-30
Show everyone's items with the left characters icons

Fouks0:
Feature that shows the items of every player in the icons at left, similar to some Mario Karts.

Personally, I really liked that feature, and maybe others Stk players might too. It would also make races a bit more exciting and strategic by having a way to anticipate the others' actions. And I can finally see who is sending me all these Basket Balls Lol.

More importantly, this is currently a way to cheat by having this feature while others don't, and I don't see any real cons to see what items others have (other than "items should be secret", but again, as we can cheat easily...).

Sorry, I can't test this online for now so hopefully that works for it too, but I don't see why it should not work.
Should work in all modes, though there is no icons at left in Soccer, but I think that this feature does not have much point here anyway and would not help a cheater much.
The icons should properly scale depending on the number of players and the resolution.

ShowEveryonesItems1
ShowEveryonesItems2
ShowEveryonesItems3

### DIFF
--- a/src/states_screens/race_gui_base.cpp
+++ b/src/states_screens/race_gui_base.cpp
@@ -1069,6 +1069,27 @@ void RaceGUIBase::drawPlayerIcon(AbstractKart *kart, int x, int y, int w,
         draw2DImage(icon, pos4, rect4, NULL, NULL, true);
     }
 
+    // Current item(s) and how many if > 1
+    const Powerup* powerup = kart->getPowerup();
+    if (powerup->getType() != PowerupManager::POWERUP_NOTHING && !kart->hasFinishedRace())
+    {
+        int numberItems = kart->getPowerup()->getNum();
+        video::ITexture *iconItem = powerup->getIcon()->getTexture();
+        assert(iconItem);
+
+        const core::rect<s32> rect(core::position2d<s32>(0,0), iconItem->getSize());
+        const core::rect<s32> posItem(x + w/2, y + w/2, x + 5*w/4, y + 5*w/4);
+        draw2DImage(iconItem, posItem, rect, NULL, NULL, true);
+
+        if (numberItems > 1)
+        {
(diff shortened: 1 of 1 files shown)


Benau:
you need to wait and see if alayan likes this idea...

deveee:
It's possible to print even more useful information on the screen. Personally I wouldn't touch current race gui, but I would add separate race gui designed for speedrunners. It could be activated eg. by changing a param in config file.

AndetSTK:
> you need to wait and see if alayan likes this idea...

(A comment Alayan made on YouTube): "Yes, I actually had this very thought. It's currently possible to cheat online if you mod STK to display the items others players have, making it visible by default would avoid the concern."

AndetSTK:
> It's possible to print even more useful information on the screen. Personally I wouldn't touch current race gui, but I would add separate race gui designed for speedrunners. It could be activated eg. by changing a param in config file.

If it were in the config but not an in-game setting, wouldn't it make it easier for players to cheat in this way?

deveee:
If you really want to cheat, you can for example spectate other players (something like splitscreen in local multiplayer), show people speed, their best time, automatically use switch item when best player wants to collect gift, automatically use shield when someone wants to hit you and really much more.

I'm rather thinking about race gui that provides more information for people that care about very good lap times.

michaelcoleD14:
Nice

james165527:
thanks men !

*Yours sincerely,*
Mark James Agripo.

dumaosen:
> It shouldn't be implemented for 1.1 though of course, since it'd be unfair for 1.0 players. :P

if you say so, the PR will never be merged lol

risostk:
If included in 1.1, we can force players to update to the new version to get an advantage.

dumaosen:
Also i think u should make it optional because maybe there's someone who doesnt like so many infos on screen

risostk:
@dumaosen, yes, we need to add a setting.

PR #4057, opened 2019-09-08
Add widget extra size

dumaosen:
Agreement

A brand new widget extra size
Before we made font size, some users complained about the widgets are too small in Android and now in very big font size is still the same.
However, the widget borders are attached with fonts.
QwertyChouskie inspired me that we can make some extra blank space around widgets.


Benau:
You missed MPL dual license statement....

qwertychouskie:
Hmm... not sure if this should be user-configurable or not.  I'll think on it a while. (Also I'll see if I can upstream some smaller changes from my branch like music selection today, to pave the way for more infra upstreamed.

Benau:
outdated now

PR #4060, opened 2019-09-15
Link against libatomic if required

reinerh:
On some architectures it is required to explicitely link against
libatomic to use e.g. 8 byte atomics.
Check during configuration if it compiles without libatomic. If not,
add it to target_link_libraries.

Noticed on Debian build servers: https://bugs.debian.org/934799

## Agreement
```

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -513,6 +513,21 @@ else()
     target_link_libraries(supertuxkart ${PTHREAD_LIBRARY})
 endif()
 
+# check if linking against libatomic is required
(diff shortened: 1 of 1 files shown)


Benau:
Do you think we need to disable it on windows?

reinerh:
I'm not very familiar with windows, but I would expect that on windows the test program compiles without problems. And as far as I know windows does not run on these architectures that don't support 8 byte atomics (like armel, mips, mipsel, m68k, powerpc, sh4).

Issue #3903, opened 2019-05-04
Bad AI in soccer mode


usbhubOtto:
## The AI of the karts in the soccer mode are very bad

The AIs of the karts in the soccer mode are very bad, so it is very easy to win (in all levels of difficulty).

## Steps to reproduce

Simply start the soccer mode with difficulty "Supertux" and try it out, it is very easy to win 10:0.

## Configuration

STK release version: 0.9.3

STK Source (PPA, distribution package, official binary, etc.): 
Distribution package
System: 
Linux Mint 19 Cinnamon
Graphics card: 
Nvidia Quadro 2000
CPU: 
Intel Xeon 4 x 3.2 Ghz

Alayan-stk-2:
The soccer AI is weak (especially in 1v1) but it works and allows the game mode to not be locked to multiplayer, which is its primary purpose.

Anard:
Maybe could play soccer 1 vs 8 to make it a little harder vs AI ?

PR #4061, opened 2019-09-15
Add options to distribute AI in single-player soccer

risostk:
I add an option for single-player soccer mode, as shown below, to have additional flexibility to control the distribution of AI among teams. There are three options:
Balanced - this is what used in stk before
All AI blue - make all AI go to the blue team
All AI red - make all AI go to the red team

This will be useful for people who want to play against more AI karts or need more help from AI. I know not everyone needs this, but they can always play with the default option: balanced.

Screenshot from 2019-09-15 16-54-50
Screenshot from 2019-09-15 16-55-11

## Agreement
```

```

### DIFF
--- a/data/gui/screens/track_info.stkgui
+++ b/data/gui/screens/track_info.stkgui
@@ -66,6 +66,15 @@
                     <spacer width="1" height="1%"/>
                     <div width="100%" height="fit" layout="horizontal-row" >
                         <div proportion="1" height="fit" layout="horizontal-row">
+                            <spinner id="ai-mix-spinner" width="100%" min_value="1" max_value="20" align="center"
+                                     wrap_around="true" />
+                        </div>
+                        <spacer width="3%"/>
+                        <label id="ai-mix-text" proportion="3" I18N="In the track info screen" text="AI team distribution" text_align="left" align="center"/>
+                    </div>
+                    <spacer width="1" height="1%"/>
+                    <div width="100%" height="fit" layout="horizontal-row" >
+                        <div proportion="1" height="fit" layout="horizontal-row">
                             <div width="100%" height="fit" text-align="center" layout="vertical-row" >
                                 <checkbox id="option" align="center"/>
                             </div>
--- a/src/config/user_config.hpp
+++ b/src/config/user_config.hpp
@@ -449,6 +449,9 @@ namespace UserConfigParams
     PARAM_PREFIX StringUserConfigParam m_last_used_kart_group
             PARAM_DEFAULT( StringUserConfigParam("all", "last_kart_group",
                                                  "Last selected kart group") );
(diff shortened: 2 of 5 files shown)


Benau:
alayan you ok to close #3903 after this pr?

Mrxx99:
You need to be aware that adding a widget more may fit in the space for the resolution you are currently using but maybe not for lower resolution. Remembering from my work on the soccer config UI it most certainly will not. Did you test with lower resolutions?

dumaosen:
@Mrxx99: See this commit in my pr https://github.com/supertuxkart/stk-code/pull/4057/commits/541d2cca4d27e10af8324b818c0ccd9e73c54b84. It gives a lot more space after that.

dumaosen:
i think we can replace it with Red team AI num and blue team AI num, because it's more advanced.

risostk:
@Mrxx99 Thanks! I tested it on low resolutions, and it does have a problem. I asked Benau, and it may not be a problem on my side: the spinner is expected to scale according to the text. Replacing with Red/Blue AI num. as maosen suggested can solve this problem.

@dumaosen Yes, I also thought about that. The current implementation is just an easy solution. To have Red AI num. and Blue AI num., I need to change the maximum selectable number in response to the user's choice. For example, let's say I am red, and I choose 3 red and 4 blue. After I increase red to 4, the blue should automatically change to 3, as blue_ai+red_ai<=7. I guess this can be done in TrackInfoScreen::eventCallback? I think it is doable and will try to implement that.

risostk:
@dumaosen Just thought more about it: I think people may not bother to select the number of AI each time, most of the time they just want to play with a balanced AI number. 
One solution is to set the default Red/Blue AI number balanced.
Another solution is to keep the spinner, but with 2 options:
- Balanced
- Custom
If 'Balanced' is selected, the user just selects the total number of AI. This is the default option.
If 'Custom' is selected, the user can select the Red AI num. and Blue AI num. separately.

I personally prefer the first solution because it has a clean UI. Any ideas?

qwertychouskie:
Maybe have two spinners: number of AI and distribution, e.g. if 7 AI are selected, show 3/2 in the distro spinner, and the spinner would be of the progress bar type, so it would go between 0/7 and 7/0.

dumaosen:
Another idea is adding a "allies number", which determines how many allies will the player get.
It is good for players who doesnt care which team he is in, so "allies number" is better than "red number" or "blue number"

risostk:
Thanks!
@qwertychouskie Local split-screen players can make this more complex.
@dumaosen Local split-screen players with both red/blue team will confuse the meaning of allies.

deveee:
It's just too small for that many options
Zrzut ekranu z 2019-09-23 18-35-38

Benau:
maybe someone can move options to above div (soccer has no highscore)...

PR #4063, opened 2019-09-17
Add options to control the number of red/blue AI in soccer mode

risostk:
Hi, 
This is an improved version compared to the previous PR (#4061). Now we can set the number of Red/Blue AI separately. It also works for split-screen. Having 8 split-screen will disable the spinners related to the AI. If increasing the number of red AI to num_red_ai+num_blue_ai>max_ai, the blue AI number will decrease by 1 automatically. The upper bound of each spinner is also taken care of.

I still added the spinner with two options: Balanced and Custom. This simplifies the implementation.

The spinner does not scale properly on low resolutions. I believe it can be solved with maosen's code.

The code may need to be cleaned a bit.

Screen Shot 2019-09-17 at 4 56 40 PM

Screen Shot 2019-09-17 at 5 02 08 PM

## Agreement
```

```

### DIFF
--- a/data/gui/screens/track_info.stkgui
+++ b/data/gui/screens/track_info.stkgui
@@ -57,15 +57,6 @@
                     <spacer width="1" height="1%"/>
                     <div width="100%" height="fit" layout="horizontal-row" >
                         <div proportion="1" height="fit" layout="horizontal-row">
-                            <spinner id="ai-mix-spinner" width="100%" min_value="1" max_value="20" align="center"
-                                     wrap_around="true" />
-                        </div>
-                        <spacer width="3%"/>
-                        <label id="ai-mix-text" proportion="3" I18N="In the track info screen" text="AI team distribution" text_align="left" align="center"/>
-                    </div>
-                    <spacer width="1" height="1%"/>
-                    <div width="100%" height="fit" layout="horizontal-row" >
-                        <div proportion="1" height="fit" layout="horizontal-row">
                             <spinner id="ai-spinner" width="100%" min_value="1" max_value="20" align="center"
                                      wrap_around="true" />
                         </div>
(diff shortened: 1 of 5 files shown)


Benau:
maybe a button open an additonal dialog?

Benau:
mabe you can default it to balanced (and shows the corresponding number to 2 spinners), when use press spinners it goes custom? (so u don't need balanced / custom box) wine

risostk:
> mabe you can default it to balanced (and shows the corresponding number to 2 spinners), when use press spinners it goes custom? (so u don't need balanced / custom box) 

Then the question is what is the default AI number? I could set it as the max_arena_players-local_players. I assume the balanced full AI is mostly used, then it won't be a big problem that the user loses the flexibility of controlling the AI number in a balanced distribution.

risostk:
Hi, I am pretty satisfied with the latest commit (169a111). There is no problem at even the lowest resolution. The UI is also clean.

The new spinner causes extra space line in normal race and FFA, see the following. Would it possible to hide the line completely?

Screen Shot 2019-09-18 at 5 37 25 PM

Screen Shot 2019-09-18 at 5 35 33 PM
Screen Shot 2019-09-18 at 5 39 08 PM

Mrxx99:
You can collapse widgets with the setCollapsed method. It is already used in the race config screen so you can look for it

risostk:
@Mrxx99 Thanks! This solved the problem.

risostk:
Hi, I am satisfied with the latest commit (671cee5). Please check the video:
https://www.youtube.com/watch?v=6LpFQfL8l3s
or
https://www.bilibili.com/video/av68342163

dumaosen:
There's a problem: in very large font size the spinners are out-of-box.
Don't worry i can fix it so just merge it

Issue #3444, opened 2018-09-15
When assigning keyboard key, game tells me I have to press the key “fully”


Wuzzy2:
## Description
When I am about to assign a keyboard key, the game tells me I have to press the key “fully”, which is funny. Probably the keyboard is confused with a controller?

## Steps to reproduce
* Go to control settings
* Configure keyboard keys
* Click any control to make the key config window pop up
* Read what it says
* Have a mild chuckle

## Configuration
Default.

STK release version: 32d192460e213fe221e6526d7f07364695a5d33c

System: GNU/Linux 64bit

auriamg:
That's just because we use the same popup and text for keyboards and gamepads

doutsh:
It's an easy fix, but it requires to add the new text to all the translation files (actually no ALL, as most of them don't translate this line for some reason)

nebbii:
In which translation files has/hasn't this been fixed?

auriamg:
@nebbii I don't understand your question. Nothing has been changed in the original version yet so translations are not affected as of now.

123survesh:
Hi I have made a PR for this issue #4075 . please do see if the approach I took is ok and tell me if its not I will try to work on it.

PR #4075, opened 2019-10-03
Fix for #3444 - Change in display text while assigning keys, based on…

123survesh:
I have added a new dialog file for keyboard (data/gui/dialogs/press_a_key_dialog.stkgui) and renamed the existing one to data/gui/dialogs/press_a_button_dialog.stkgui. Also I have updated the translation files to accommodate the above changes.

I didn't know how to manipulate the content of the dialog files after they were rendered and hence I have taken this approach. Please do suggest if there is any other way to do this, I will update my PR.

## Agreement
```

```


Benau:
you can use setText() on the "title" widget instead

PR #4076, opened 2019-10-04
Fix for #3444

123survesh:
* Added a check for the keyboard input device and changed the text of the "title" widget using **Widget->setText("id")** method
* Have hard coded the string **"Press any key..."** in **src/states_screens/dialogs/press_a_key_dialog.cpp** 

****
## Agreement
```

```

### DIFF
--- a/src/states_screens/dialogs/press_a_key_dialog.cpp
+++ b/src/states_screens/dialogs/press_a_key_dialog.cpp
@@ -28,10 +28,15 @@ using namespace irr::gui;
 
 // ------------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


Benau:
you don't need a03ddd8 as it should be run by data/po/update_pot.sh

123survesh:
@Benau I accidentally closed the last PR, I did a hard reset to revert the changes. I saw in the translation files that some strings were hard coded in the code, so I just hard coded my string as well and added an entry for that string in the translation files :sweat_smile:

Benau:
no they are all added with data/po/update_pot.sh

123survesh:
Ah alright :+1:

PR #4087, opened 2019-10-08
Add an option to put the minimap on the center of screen.

luffah:
This PR add an option "Centered" to place minimap on the center of the screen.

~~The minimap is not perfectly centered  because of the computing of minimap width which doesn't concern this change.~~

EDIT: The minimap has its left side on the center of the screen. That allows to read the sceen in Z (what is in front;  kart; map;  speed)

## Agreement
```

```


Benau:
Why do you want to see minimap in the center of screen?

And why do you change our ttfs?

luffah:
Sorry for the ttf. It was not intended. Just removed it.

I want to see the minimap in the center on the screen in order to understand fastly where is the ball in soccer mode.

It allow me to switch 2d gameplay (which is more practical for an overview of the situation)  and 3d gameplay (which allow more precision and immersion).

Benau:
Alayan!

Benau:
test

Alayan-stk-2:
I'm not convinced about it. This is just an option, and not even the default one, so I'm open to configurations I would not like myself if they can be useful for some users.

However, there is an expected level of quality for options we're officially supplying, and I'm wondering if it's good enough and useful enough to be in the core code.

Could you please provide some screenshots of your personal settings ?

Also, don't update the .po file manually in your PR, this will get overwritten later on anyway because translations are done on transifex.

luffah:
After a week of testing, i feel it was more comfortable that the map shall have its left side exactly on the center of the screen.  
But finally it looks like a slight offset change on 'Right' position when the size of the minimap is 360; and just put from right to bottom when the size is 180 (standard).

# Soccer mode
**Centered with standard minimap size** :
Capture d’écran_2019-10-12_02-23-56


**Centered** :
Capture d’écran_2019-10-12_02-01-46
**On right side**:
Capture d’écran_2019-10-12_02-02-53

# Race mode
**Centered** :
Capture d’écran_2019-10-12_01-56-11

**On right side**:
Capture d’écran_2019-10-12_01-57-47

luffah:
Additionnal note: in almost every mainstream soccer game, there is a minimap (or an indicator of where is the ball) in bottom center of the screen.

However i'm not completely happy with the render of the map in current state.

Alayan-stk-2:
Thank you for your work. The option seems good enough to deserve merging. I made a slight adjustment, keeping a small bottom margin is important.

PR #4092, opened 2019-10-12
fix linking order: libpng before libz

risostk:
See title.
## Agreement
```

```


deveee:
Does it fix some real issues? I know, linking order matters, but it was always working for me as it is.

risostk:
@deveee Yes. For static linking, the correct order should be -lfreetype -lpng -lz. I tried to compile harfbuzz on Linux with gcc. I didn't try to compile for Android, just following this script. The linking order does matter for me: the compiler complained of undefined references. The problem was gone by switching the order.

Benau:
Ok deve lets android apk

Issue #3620, opened 2018-11-24
Player spawn in 3-strikes battle is always the same


Wuzzy2:
## Description
The position in which the player spawns in 3-strikes battle is always the same (for the same number of AI karts), but I think it should be random.
Might become a bit repetitive if you play 3-strikes battle a lot.

## Steps to reproduce
1. Start 3-strikes battle with X karts
2. See where you spawned
3. Leave to main menu
4. Repeat 1.-3. until you get bored. Do not change the number of karts

## Configuration

STK release version: 459119075a724ab4aec81d6b86aa7f931386d479

PR #4093, opened 2019-10-14
Randomizing spawn points for battle mode #3620

Aleman778:
This PR changes the kart status struct to include a spawn id that can be used when setting the start transform. 

This fixes the issue #3620 for playing offline but it may still remain for online play. I wasn't able to fix this issue for online play.

## Agreement
```

```

### DIFF
--- a/android/make.sh
+++ b/android/make.sh
@@ -350,7 +350,7 @@ if [ ! -f "$DIRNAME/obj/harfbuzz.stamp" ]; then
 
     cd "$DIRNAME/obj/harfbuzz"
     FREETYPE_CFLAGS="-I$DIRNAME/obj/freetype/include" \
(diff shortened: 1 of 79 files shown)


Benau:
Maybe std shuffle starting transform for offline battle?

Aleman778:
> 
> 
> Maybe std shuffle starting transform for offline battle?

Yes that would work I didn't think about that! Any ideas for implementing this on the server?

Benau:
You dont need to, starting position is already randomized in online games

Aleman778:
I have updated the code to shuffle the starting transforms and removed the code for the previous approach. I was not able to check if the given world was online or not so I had to check if the first player has a global id of -1 to ensure that shuffle is not done online.

Benau:
NetworkConfig::get()->isNetworking()?

Aleman778:
That works!

Benau:
also can u remove newline changes for race_manager.cpp?

Aleman778:
Ops! Forgot to do that before.

PR #4136, opened 2019-11-12
Add font thickness support

dumaosen:
## Agreement
```

```
New font thickness support.
It is useful for skins like Qwerty's new one.


deveee:
Any particular reason to use
```
                if(node->get("regular_outline", &m_regular_outline))
                {
                    m_font = true;
                }
                if(node->get("bold_outline", &m_bold_outline))
                {
                    m_font = true;
                }
```
?

dumaosen:
@deveee In UI options, changing a different font will trigger skin manager update.
A font with different thickness needs an update too, so we make m_font=true to mark it as a new font, so a skin update will be triggered in the occasion.

Benau:
Please move it to only stk_config

PR #4138, opened 2019-11-15
UI fix bundle XIII

dumaosen:
## Agreement
```

```


Benau:
why do u need to change src/states_screens/grand_prix_win.cpp?

dumaosen:
because the texts are badly placed in high resolution

qwertychouskie:
It looks like you moved the item icons to the icons folder, nice!  Been wanting that for a bit :)  If you could also adjust the race GUI code to use these icons, so that the icon theme can change these icons, that would be awesome :)

dumaosen:
No it's not Qwerty. In my opinion race GUI should be apart from the other GUI. We can change the icons in the help menu in order to make it fit to the skin, but it won't fit the style of the in-game GUI if it can also influence the in-game powerup icons.

Benau:
i dont want to duplicate power icons png in both models and data/gui/icons, especially now they are the same

PR #4140, opened 2019-11-15
Some small fixes

psypherium:
## Agreement
```

```


Alayan-stk-2:
Changes to how the server configuration xml looks like should also be applied to the xml file itself no ?

PR #4143, opened 2019-11-23
Add a function to use UTF-32 emojis

risostk:
Add a function to use UTF-32 emojis, e.g. getUtf32Emoji(0x1f947).
I used this function to add the medal emoji for players that finished a race. As that may not be of interest to many players, let me just push this function right now. It can be used in many other places.

## Agreement
```

```

### DIFF
--- a/src/utils/string_utils.hpp
+++ b/src/utils/string_utils.hpp
@@ -276,6 +276,34 @@ namespace StringUtils
     }   // getCountryFlag
 
     // ------------------------------------------------------------------------
+    /** Return string for UTF32 emoji. */
+    inline irr::core::stringw getUtf32Emoji(const uint32_t code)
+    {
+        uint32_t emoji[2] =
+        {
+            code,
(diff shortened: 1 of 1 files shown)


gsbhasin123:
Yes please merge is very great

Benau:
some emoji can be "combined" for example 🏳️‍🌈 (The Rainbow Flag emoji is a sequence of the 🏳 Waving White Flag and 🌈 Rainbow emojis. These are combined using a zero width joiner between each character and display as a single emoji on supported platforms.)

maybe just use utf32towide?

risostk:
hmmm. Do we already have the utf32towide function in stk? I did a quick search but didn't find it. Or you mean it is a C++ function? So benau, you want a more general function than this one that can handle the case of combined emoji?

risostk:
plz use StringUtils::utf32ToWide({0x1f947}), so close this pr.

PR #4135, opened 2019-11-10
Initial works on tips manager

dumaosen:
tips

## Agreement
```
By creating a pull request in stk-code, you hearby agree to dual-license your contribution as
GNU General Public License version 3 or any later version and
Mozilla Public License version 2 or any later version.

This includes your previous contribution under the same name of contributor.

Keep the above statement in the pull request comment for agreement.

```
### WARNING: big PR
The initial work on tips manager.
Things to do, which should be in different PRs:
A lightbulb icon;
More useful tips and guide of tips.xml;

PR #4146, opened 2019-11-26
Initial works on simple tips manager

dumaosen:
## Agreement
```

```
A simple version of #4135


dumaosen:
Update pot after merged plz!

Alayan-stk-2:
I tested it. Loading screens often don't last long enough to comfortably read the tip, but it is displayed fine.

The end-race messages using the messaging system work fine.

I'll rework the text of the tips, and I expect that there will be many improvements to be done later on to the tips system, but this works and I didn't notice anything horrendous when looking at the code.

Thank you for your contribution !

PR #4155, opened 2019-12-03
use /listaddon in network lobby

risostk:
Hi, I add a function /listaddons in the network lobby to list all (local) addons' id (installed and non-installed). The player can further install addons with /installaddon id
/listaddons will list all addons
/listaddons str will list addons whose name matches str

For example: /listaddons island will give results:
kapman-island-resort (n), mystery-island_1 (n), mystic-isl (n), supertux-island-resort (n), the-isl (y), the-old-isl (n), volcan-isl (n), the-old-island_2 (y),
Here, (n) means this addon is not installed, (y) means this addon is installed.

Currently I output the result to the terminal. I am thinking to change it to the text box above the chat box. But I am not sure if it is a good idea (possible spam).

Screen Shot 2019-12-02 at 7 00 11 PM

## Agreement
```

```


Benau:
Sorry I've already done that (just unpushed)

PR #4156, opened 2019-12-03
Add left and right side camera views on kart controls

luffah:
Add left and right side camera views on kart controls :
- add 2 keys for keyboard  to look at left and at right
- add 1 button on gamepad to look at left (imcomplete choice, to discuss)
- no influence (currently) on shooting items 

## Agreement
```

```

### DIFF
--- a/src/network/protocols/server_lobby.cpp
+++ b/src/network/protocols/server_lobby.cpp
@@ -3020,9 +3020,58 @@ bool ServerLobby::handleAssets(const NetworkString& ns, STKPeer* peer) const
         return false;
     }
 
+    std::array<int, AS_TOTAL> addons_scores = {{ -1, -1, -1, -1 }};
+    size_t addon_kart = 0;
+    size_t addon_track = 0;
+    size_t addon_arena = 0;
+    size_t addon_soccer = 0;
+
+    for (auto& kart : m_addon_kts.first)
+    {
+        if (client_karts.find(kart) != client_karts.end())
+            addon_kart++;
+    }
+    for (auto& track : m_addon_kts.second)
+    {
+        if (client_tracks.find(track) != client_tracks.end())
+            addon_track++;
+    }
+    for (auto& arena : m_addon_arenas)
+    {
+        if (client_tracks.find(arena) != client_tracks.end())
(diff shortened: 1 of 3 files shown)


Benau:
You can keep only the camera mode changes only and use some F1-12 key to control camera object instead, see debug.cpp

luffah:
I freeze of some days, maybe 2 weeks.

I intend to change the code, agreeing to :
- this experimental (new) feature shall not be usable without modifying the user config file
- network logic shall not be impacted

Given, i can't align change with artist_debug_mode because of existing keys and features that are too much for a simple camera addition, I'll look to put that as a new hidden option in user config.

Too if the keys can't be set, this PR is useless.
Therefore i'll propose keybinds only if the hidden option is on.

Benau:
i'm not even sure we want this feature... (android user can't do the same)

Benau:
Or maybe just use higher fov for arena (soccer included) modes? (like 100, try change in stk_config)

I assume your pr purpose is only that?

luffah:
> I assume your pr purpose is only that (better soccer gameplay!)!?
yes, it is for semi-hardcore gamers :
- in soccer mode, it allows the goalkeeper to follow the ball in any position
- in arena mode, it allows to check every body around, and have a better hide and seek strategy
- in race mode, it allows to see very tight turn and help to find shortcuts

Benau:
Lol you cannot even drive sideway for gatekeeper

luffah:
> Lol you cannot even drive sideway for gatekeeper

Oh good idea... not a camera change but add the possibility to drive sideway like a space ship :p

luffah:
I didn't found a minimal way that allow to add the keys without adding m_look_left and m_look_right in kart_control.hpp . It seems to be required to (given the camera shall go back to normal when the key is released - in kart.cpp)  keep the view active while the key is pressed (in local_player_controller.cpp).

luffah:
Tested for some hours. Finally, this change is more aesthetical. I'm not satisfied :(

luffah:
New path to the feature : https://github.com/luffah/stk-code/tree/side-camera-normal

PR #4161, opened 2019-12-09
fix #4123

### LINKED ISSUE #4123: Story Mode Timer is not set to zero for a new account after previous player
## Description
When there are several offline accounts for a speedrun, switching between them and quitting/launching can make the timer value invalid.

### First way

Create an offline account A and start the story mode from it, wait for some time (e.g. 30 seconds). Then exit to the main menu and create a new offline account B, then quit the app. When you relaunch, try to start the story mode from B, the timer will show time starting from nonzero value (specifically, the time you waited with A).

### Second way

Create new offline accounts A and B and start the story mode from A, wait for some time. Then exit to the main menu and try to launch story mode for B, the timer will display 99:59:99 and will not change.

Also if the timer is not enabled from the settings before the run starts, you probably can see no fractional parts of a second (.561).

## Configuration

STK release version: git updated on Nov 1, 2019, commit 770d02b

STK Source (PPA, distribution package, official binary, etc.): compiled from source

System: Ubuntu 19.04

Graphics card: 

CPU: 

Gamepads/keyboards models if related to the issue: 

## Additional information

stdout.log:
```

```

input.xml:
```xml

```

Error message:
```

```

GuillaumeBft:
## Agreement
```

```

### DIFF
--- a/src/challenges/story_mode_status.cpp
+++ b/src/challenges/story_mode_status.cpp
@@ -345,8 +345,14 @@ void StoryModeStatus::save(UTFWriter &out, bool current_player)
             story_mode_timer->pauseTimer(/*loading*/ false);
         }
 
-        m_speedrun_milliseconds = story_mode_timer->getSpeedrunTime();
-        m_story_mode_milliseconds = story_mode_timer->getStoryModeTime();
+        if(m_first_time){
+            m_speedrun_milliseconds = 0;
+            m_story_mode_milliseconds = 0;    
+        }
+        else{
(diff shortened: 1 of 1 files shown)


Benau:
We put { in the next line, also space after if

GuillaumeBft:
Sorry, I won't forget for the next one

Benau:
@kimden can you see if your bug is fixed after this patch?

kimden:
Yes, it's fixed. The only thing that remains is the last paragraph of my issue (about milliseconds), but I guess I should have made a separate issue for that since it's more related to having two timers.

Issue #3376, opened 2018-08-08
Allow to filter for installed/uninstalled addons in the addons menu


Alayan-stk-2:
There is a visual indicator in the list of addons telling the player if he has already installed or not an addon, but it's not currently possible to separate addons based on this.

This could be useful when searching for a new interesting addon, or when trying to uninstall an addon, as displaying addons of the unwanted type would clutter the screen.

PR #4162, opened 2019-12-09
Show only possessed checkbox in addons screen #3376

GuillaumeBft:
## Agreement
```

```


Benau:
you can keep the widget pointer inside addonscreen to avoid getwidget each time

PR #4164, opened 2019-12-10
Debug input refactoring

luffah:
## Agreement
```

```
Hello, as indicated in input_manager.cpp , debug actions had to be moved in debug.cpp.
Here my proposal of refactoring:
- key with no associated action are parsed by Debug::handleStaticAction
- given they are already computed : the key, its value and modifiers, and current world are forwarded to the Debug method 
- in order to make the static shortcut more understandable, they are all in the same file (new file keyboard_static_keys.hpp)
- added a lock in debug.cpp to avoid multiple events on 1 key stroke


luffah:
Note : If the numbers of commit is a problem (due to the merge). I'll redo a new one from last version.

Benau:
```
#ifndef HEADER_KEYBOARD_STATIC_KEYS_HPP
#define HEADER_KEYBOARD_STATIC_KEYS_CONFIG_HPP
```

luffah:
(done)

luffah:
Shall this PR be closed (and converted to a future release issue) or be reworked as soon as possible ?

Benau:
...because it looks more complicated after refactoring

luffah:
Not convinced by the argument, cause there is less code repeat. The code referencing semantic instead of the concrete key. is easier to read IMHO.
But indeed, there is many lines in that change on a point that is only about debug mode.

Ok. let's forget this.

PR #4165, opened 2019-12-10
Super Speedometer

CodingJellyfish:
## Agreement
```

```
An improvement of #4099 
It adds total kart & speed evaluation 
Todos:
Make rank somewhere else possible

### DIFF
--- a/data/gui/screens/options_ui.stkgui
+++ b/data/gui/screens/options_ui.stkgui
@@ -52,6 +52,17 @@
                 </div>
 
                 <spacer width="5" height="2%"/>
+                
+                <div width="100%" height="fit" layout="horizontal-row">
+                    <div layout="vertical-row" width="10f" height="fit">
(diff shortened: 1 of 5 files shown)


luffah:
@dumaosen  Nice job. But it still can be better. If i can comment
- for rank display the rendering of the digits looks poor.
- for speed view, it shall be smoothed (rounding speed value and maybe limit update to 4 time per second max) 
- i don't to choose between speed and rank...

Fouks0:
I don't think that it needs an update frequency limit but else I agree with everything
We could maybe move the rank near the lap counts, with the "Podium" logo...
Also if we use km/h (which is appropriate), the speedometer picture might have to be changed to have the marks matching km/h and not m/s...

Benau:
outdated.........

CodingJellyfish:
hmm so should I revamp this or just throw it? I’ve found that Benau doesn’t like adding a speedometer option…

Issue #3893, opened 2019-05-01
Players can spam the chat with very long messages, making lobby and gameplay unusable


Wuzzy2:
## Description
Players can spam/flood the chat with very long messages. This has the following effects:

- In lobby, chat becomes unreadable. Others are basically deprived of their speech
- In game, the chat overlay becomes very large and covers half of the screen. The game becomes very difficult, it's like having a plunger in your face, but for the entire race.

## Steps to reproduce
- Play online
- Write very long messages
- Congratulations! You are now a master troll and hated by everyone

## Workaround

Immediately kick players that abuse the chat.

## Configuration

STK release version: 1.0.0

## Possible fixes

1. Limit effective chat message length, both on server and client side.
2. Limit the size of the chat overlay in the GUI. Make sure it only covers a small part of the screen and that GUI does not break down, no matter the input. Maximum size be two lines. Cut off any text beyond that.
3. ???

UnMe0w:
Yes, the layout is completly broken when it mess up with this: 
`testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttest`
![Capture du 2019-05-01 18-35-32](https://user-images.githubusercontent.com/30691126/57028692-07700780-6c40-11e9-8a33-ca2b6865c18d.png)

BlueCatYellowCat:
Maybe instead of trying to autoban users for long messages or frequent messages, the chat feature could have a character limit and a frequency limit (the server could enforce the same, in case some troll wants to replay chat packets to circumvent client-side limiting)

BlueCatYellowCat:
It might also be worth (if not done already) limiting characters so newlines can't be spammed, etc..

sgreene570:
Would it be possible to word wrap text without spaces, say if you were sending a url or other string in chat?

dumaosen:
See https://github.com/supertuxkart/stk-code/issues/3781

UnMe0w:
Or allow scrolling to previous messages

Wuzzy2:
This issue is !!!NOT!!! just about word wrapping! Word wrapping is just a part of it.
It is also about that players can send very loooong messages that cover the entire screen. Just try repeating the same word over and over again.

god286:
I think there could be a system that detects if the text is touching the edge of the chat. Then,  if so, open a new line.

BlueCatYellowCat:
[...]

PR #4168, opened 2019-12-12
add message freguency limit in chat #3893

GuillaumeBft:
## Agreement
```

```

In chat, ensure that for a person writing more than 3 consecutive messages (less than 5s between 2 messages) prevents display an other message from him while the next 5 seconds.

### DIFF
--- a/src/network/protocols/server_lobby.cpp
+++ b/src/network/protocols/server_lobby.cpp
@@ -698,6 +698,20 @@ void ServerLobby::handleChat(Event* event)
     // Update so that the peer is not kicked
     event->getPeer()->updateLastActivity();
     const bool sender_in_game = event->getPeer()->isWaitingForGame();
+
+    int last_message = event->getPeer()->getLastMessage();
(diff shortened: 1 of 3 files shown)


dumaosen:
please make it configurable in server.xml,  or i can help you after merge.

kimden:
This is too strict in my opinion. I would prefer character limit per some time interval like 5s, but not that strict. And yes, probably the best option is making it configurable on a server.

GuillaumeBft:
I will try to make it configurable in server.xml. By making it configurable on a server @kimden , you mean like an option for the owner for enable or increase/decrease the timer on the server lobby ?

kimden:
I guess making it configurable (and turnable on/off) in server config is sufficient, I don't think this setting changes too often.

GuillaumeBft:
PreventSpam
What do you think about adding a gauge like this ?

Benau:
no we don't even want that in server configuration screen, the server configuration screen is meant to be simple

just add that to server config.xml

Benau:
...let me add this for you is even quicker

PR #4174, opened 2019-12-21
Include linux/sysctl.h rather than sys/sysctl.h on non-glibc systems

### LINKED ISSUE #4172: Alpine Linux sysctl.h build failure
I'm trying to package SuperTuxKart for Alpine Linux. This distribution uses Musl rather than glibc and this seems to cause an issue.

When compiling Irrlicht I get the following:

```
/home/builder/aports/testing/supertuxkart/src/stk-code-1.0/lib/irrlicht/source/Irrlicht/COSOperator.cpp:20:10: fatal error: sys/sysctl.h: No such file or directory
   20 | #include <sys/sysctl.h>
      |          ^~~~~~~~~~~~~~
```

Looking at [the source](https://github.com/supertuxkart/stk-code/blob/master/lib/irrlicht/source/Irrlicht/COSOperator.cpp#L17), it includes `<sys/sysctl.h>` unless the platform is Android. However, at least on Alpine Linux (and I guess this is due to using Musl), it should be `<linux/sysctl.h>` instead, just as on Android.

The following snippet fixes the issue.
```
diff --git a/lib/irrlicht/source/Irrlicht/COSOperator.cpp b/lib/irrlicht/source/Irrlicht/COSOperator.cpp
index ba9612fc9..70d6193af 100644
--- a/lib/irrlicht/source/Irrlicht/COSOperator.cpp
+++ b/lib/irrlicht/source/Irrlicht/COSOperator.cpp
@@ -14,7 +14,7 @@
 #if !defined(_IRR_SOLARIS_PLATFORM_) && !defined(__CYGWIN__)
 #include <sys/param.h>
 #include <sys/types.h>
-#ifdef ANDROID
+#if defined(ANDROID) || !defined(__GLIBC__)
 #include <linux/sysctl.h>
 #else
 #include <sys/sysctl.h>
```

I would PR this, but it seems Irrlicht gets auto-copied and modified using some script so I'm not sure what to do here.

PureTryOut:
## Agreement
```

```
I'm not sure if SuperTuxKart is supported on non-Linux and macOS *nix platforms like BSD, as this might break it.

Anyways, this fixes #4172.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,9 @@
 sudo: false
 language: cpp
 os:
-  - osx
+  - linux
 compiler:
+  - gcc
   - clang
 
 matrix:
@@ -15,6 +16,8 @@ matrix:
 
 env:
   matrix:
+    - BUILD_TYPE="Debug" SERVER_ONLY="OFF"
+    - BUILD_TYPE="Debug" SERVER_ONLY="ON"
     - BUILD_TYPE="Release" SERVER_ONLY="OFF"
     - BUILD_TYPE="Release" SERVER_ONLY="ON"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,8 @@
 sudo: false
 language: cpp
 os:
-  - linux
+  - osx
 compiler:
-  - gcc
   - clang
 
 matrix:
@@ -16,8 +15,6 @@ matrix:
 
 env:
   matrix:
-    - BUILD_TYPE="Debug" SERVER_ONLY="OFF"
-    - BUILD_TYPE="Debug" SERVER_ONLY="ON"
     - BUILD_TYPE="Release" SERVER_ONLY="OFF"
     - BUILD_TYPE="Release" SERVER_ONLY="ON"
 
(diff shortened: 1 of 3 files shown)


Benau:
```
/Users/travis/build/supertuxkart/stk-code/lib/irrlicht/source/Irrlicht/COSOperator.cpp:18:10: fatal error: 'linux/sysctl.h' file not found
#include <linux/sysctl.h>
```
On OSX travis 

maybe there is some define for musl only?

PureTryOut:
There isn't sadly. Then again, it shouldn't have to mark itself as different. Code that works on Musl should also just work on say uclibc or any other POSIX compatible C std.

Maybe we can do the check other way around then? Like `if unix and macOS then <do stuff> else if unix and glibc then <do stuff> else <do android and non-glibc stuff>`?

Benau:
check using cmake script?

PureTryOut:
I pushed a new change, now checking if the platform is Linux and `__GLIBC__` _hasn't_ been set. Hopefully should work better?

That said, it seems glibc is deprecating `<sys/sysctl.h>` and in case of Linux `<linux/sysctl.h>` should be used anyway? Source https://github.com/dotnet/coreclr/pull/27048

PR #4177, opened 2019-12-24
Always show times in ms

Fouks0:
There is no reason to have centiseconds here and milliseconds there...

The only drawback is that most of the time, the time in the finish screen is different of a few ms from the one shown at the left icon due to the fact that one is based on an interpolation and the other on the Tick number (or sometimes due to a bug, https://www.youtube.com/watch?v=MMbPQWepM8k ).
But that is an issue that should be fixed, and we should eventually have consistent times and precision everywhere anyway...

The Pull Request also changes the time string in the Ghosts List/Dialog to have something like "1:12:346" rather than "72.3456s", or "1:23.000" rather than "83s".

Pr1
Pr2
Pr0
Pr3

There should be a fixed width font for times...

## Agreement
Well if I could I would release my Pull Requests to the Public Domain... 68747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f7069672f64616e63652e67696668747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f706967786d2f64616e63652e67696668747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f7069672f64616e63652e67696668747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f706967786d2f64616e63652e67696668747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f7069672f64616e63652e67696668747470733a2f2f6c69686b672e636f6d2f6173736574732f66616365732f706967786d2f64616e63652e676966
```

```

### DIFF
--- a/src/race/race_manager.hpp
+++ b/src/race/race_manager.hpp
@@ -741,10 +741,7 @@ public:
      /** \brief Returns the number of second's decimals to display */
     int currentModeTimePrecision() const
     {
-        if (isEggHuntMode() || isTimeTrialMode())
-            return 3;//display milliseconds
-
-        return 2;//display centiseconds
+        return 3; // display milliseconds
     }   // currentModeTimePrecision
     // ----------------------------------------------------------------------------------------
     /** \brief Returns true if the current mode has laps. */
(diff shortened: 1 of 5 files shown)


Benau:
lol why you deleted stk online account...

PR #4182, opened 2019-12-25
Expose getKartType(int kartId) to scripting

BlueCatYellowCat:
This change attempts to expose the KartType (local player, networked player, AI, etc.) to scripting.
This is useful if a track creator wants an action to only occur if triggered by a real player, or only if triggered by a non-networked player (e.g. alternate route music in Frozen Drive unofficial add-on).

I have rushed this in to try and get it added in STK 1.1 and I have not yet been able to configure and build STK from source, **so this change is currently untested!** If there is still time before 1.1, I can try building myself but otherwise I may not be able to.

I have placed this function in the Track section because it uses the race manager. Please let me know if the Kart section is more appropriate.

## Agreement
```

```

### DIFF
--- a/src/scriptengine/script_track.cpp
+++ b/src/scriptengine/script_track.cpp
@@ -165,7 +165,16 @@ namespace Scripting
         {
             return race_manager->getNumLocalPlayers();
         }
-
+        
+        /**
+          * Gets the kart type, such as local player, networked player, AI, etc.
+          * @return A KartType enum as defined in race_manager.hpp, implicitly casted to an int
+          */
+        int getKartType(int kartId)
+        {
+            return race_manager->getKartType(kartId);
+        }
+        
         bool isTrackReverse()
         {
             return race_manager->getReverseTrack();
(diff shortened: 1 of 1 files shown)


Benau:
[supertuxkart.zip](https://github.com/supertuxkart/stk-code/files/4000612/supertuxkart.zip)
Assume you are using 64bit windows you can replace the exe with the installed one and see if it works for your track?

BlueCatYellowCat:
I tested with that .exe, and it appears to work perfectly. My kart returns 0 (KT_PLAYER), AI return 2 (KT_AI) and my ghost returns 4 (KT_GHOST).

Thanks for the help :)

Benau:
[supertuxkart.zip](https://github.com/supertuxkart/stk-code/files/4018814/supertuxkart.zip)
btw you can try try #if STK_VERSION > 10000099 in scripting now with this zip, can you test?

10000099 is 1.0.0 btw, you can:
```
    string test = Utils::versionToInt("1.0.0");
    string test1 = Utils::versionToInt(Utils::getSTKVersion());
    Utils::logInfo(test + " " + test1);
```
to print current and any version string to integer in scripting.

And I suggest you just use #if STK_VERSION > 10000099 to define the whole scripting.as, then the 1.0 stk will just ignore this scripting.as completely

I think support old stk version is not very important as STK is always "bleeding edge", also i don't think mobile stk user can even edit the scripting.as file in an easy way...

deveee:
Not sure how this function works:
https://www.angelcode.com/angelscript/sdk/docs/manual/doc_global_import.html
but maybe it's possible to check version in runtime and if > 1.0 then load module.

Or write minimum version in filename, eg "angelscript-1.1.as" or in first line in the file. Or just increase track version number?

PR #4190, opened 2019-12-29
Update wording in Tip manager

DoggoOfSpeed:
## Agreement
```

```
Updates Some tips to be clearer.


Benau:
I find the qwerty one better because he doesn't remove https (which can be clickable in the future)

PR #4195, opened 2020-01-04
Skin padding implementation

qwertychouskie:
Needed for my new theme.  Tested with all existing themes, and all should be well.

## Agreement
```

```


CodingJellyfish:
Ok i'll close mine.
(Looks like it's Qwerty's biggest PR ever!)

Benau:
image
now it has gaps between tabs

qwertychouskie:
If the gaps bother you, they can be removed by setting a negative `h_margin`, e.g. `-8`.

Benau:
if any negative values don't matter then it should not be a -8 (too arbitrary!), maybe all those padding values just need a fixed set of values for old skin

Benau:
image
Still problemtic.... 
LHS your branch, RHS original, seems that your branch doesn't seem to use available space in a good way...

src/guiengine/screen_keyboard.cpp
```diff
@@ -225,17 +223,21 @@ void ScreenKeyboard::createButtons()
...
         for (int j = 0; j < cols_num; j++)
         {
+            ButtonWidget* button = new ButtonWidget();
+
             float width = (float)total_width * layout_proportions[i][j] 
                                              / total_proportions - margin;
             
             char width_str[100];
-            sprintf(width_str, "%i", (int)roundf(width));
+            sprintf(width_str, "%i", (int)roundf(width / (SkinConfig::getHorizontalInnerPadding(button->getType(), button)+1.0f)));
```
auriamg: why +1.0f ?

Benau:
image
still bad space usage

src/guiengine/screen_keyboard.cpp
Benau: qwerty said in irc btw:
```
1.0f is the height of the button with no inner padding, so (SkinConfig::getHorizontalInnerPadding(button->getType(), button)+1.0f) is the real height of the button
```

PR #4196, opened 2020-01-04
Missing header for network struct, vital for FreeBSD.

devnexen:


Benau:
We cannot merge your pr without agreement..... (and i think you recieved the dual-licensing mail?)

PR #4197, opened 2020-01-04
Missing header for network struct, vital for FreeBSD.

devnexen:
## Agreement
```

```


Benau:
Thank you, now let me test on apple later to make sure it works everywhere....

PR #4198, opened 2020-01-04
Revert "Use openssl default after its relicensing to apache 2.0"

qwertychouskie:
This reverts commit 173e3907556070f07e1c77ad5ae20d56dad757c2.
OpenSSL 3.0, the first version to be licensed under the Apache 2.0 license, won't be coming for a while, as per https://www.openssl.org/blog/blog/2019/11/07/3.0-update/

## Agreement
```

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ option(USE_SYSTEM_GLEW "Use system GLEW instead of the built-in version, when av
 option(USE_SYSTEM_WIIUSE "Use system WiiUse instead of the built-in version, when available." OFF)
 option(USE_SQLITE3 "Use sqlite to manage server stats and ban list." ON)
 
-option(USE_CRYPTO_OPENSSL "Use OpenSSL instead of Nettle for cryptography in STK." ON)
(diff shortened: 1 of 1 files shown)


Benau:
well i guess distro who care about this matter will auto change the flag anyway.... (and we now dual license stk to mpl anyway)?

qwertychouskie:
It's not just about distros, it's about everyone who has contributed code to STK under the GPL license.  The re-licensing is good, but will likely take a while.  Is there any advantage to defaulting to OpenSSL?  AFAIR, Nettle support was added because of the GPL, and that reason is still valid until OpenSSL 3.0 or MPL STK.

Benau:
because some stk fans seem to have problem finding nettle in arch linux....

qwertychouskie:
As per `CMakeLists.txt`, STK should automatically fall back to OpenSSL if Nettle is not present: https://github.com/supertuxkart/stk-code/blob/master/CMakeLists.txt#L562-L564

Also, it seems STK can only use OpenSSL on macOS and Windows MinGW builds, but that's another issue for another time.

Benau:
qwerty in the future all deps that will be provided by stk team will use only mbedtls, so i will  this PR

Issue #4187, opened 2019-12-28
Improving the Tutorial


CodedOre:
While the tutorial is not bad and show most of the things you need to know about playing the game, I see a few issues here:
- [x] At first, a new player will likely not find it, since it is not at an prominent spot in the menu. So it can’t help people who don’t find it. Therefore I would suggest to show an dialog asking if you want to start it when a new profile is made.
- [x] Secondly, the current state of showing text boxes with the tutorial hints disturbs the game flow, especially on mobile devices. It also don’t make a good job on mobile because it can’t point on the on-screen buttons. So, i would suggest doing it like the tutorial in [Horizon Chase](https://youtu.be/diMMEOrUvc8) by slow down the game, highlight the button (or showing the keyboard key as icon) and showing the text and confirm it by doing the action. 
- [x] Lastly, the tutorial don’t show the option for fire backwards. So, instead of shooting the crates I would suggest this:
  After picking up the power up, we could drive in an AI-Kart to shoot at. After this, a second kart would appear behind the player, and the hints would show “Fire while looking back to fire backwards” (or something similar). The AI-Karts could come from an side street on the track, which would need an adapted map, but I think they should not appear out of nowhere.

Benau:
2 is a bit complicated and not so interesting.....

For 3 maybe just after passing the shooting boxes scene tell player to shoot the behind boxes again? (We can use some scripting to reset the boxes to orginal place when the dialog is shown)

![](https://lihkg.com/assets/faces/lm2xm/nosee.gif)

CodedOre:
I think especially 2 would be important to improve. It is kinda annoying to have the a constant stop and go with the action. Like I said, it disrupts the game flow and lowers the quality of the tutorial.
However I know my suggestion could be complicated to implement, since it required a lot of code not used by anywhere else, so maybe there is an better way.

I know that the tutorial is something you just opens one time and then visit „never“ again, but this is the moment a new player will (or should) see at first, so I think it would be good not to scare them away with an bad presentation.

Benau:
or use messagequeue (the thing you see for ingame online chat) to push message and only dismiss the message after the require task is acomplished? ![](https://lihkg.com/assets/faces/pigxm/snowhat.gif)![](https://lihkg.com/assets/faces/pigxm/dance.gif)

CodedOre:
MessageQueue could do it. As i see it, we have to create a MessageQueue-Implementation for the Scripting-Engine and adapt the Script of the Map.

[...]

PR #4201, opened 2020-01-05
Enabling MessageQueue for Scripting

CodedOre:
As @Benau suggested to use MessageQueue for *making the Game Flow of the Tutorial smoother* (#4187), this Pull Request enable MessageQueue for Track-Scripts.

For me it works well with MessageQueue, however i think we could make the Font of a Message a little bit bigger, since it is quite small, especially on mobile.

## Agreement
```

```

### DIFF
--- a/src/scriptengine/script_gui.cpp
+++ b/src/scriptengine/script_gui.cpp
@@ -30,6 +30,7 @@
 #include "tracks/track_object_manager.hpp"
 #include "utils/string_utils.hpp"
 #include "utils/translation.hpp"
+#include "guiengine/message_queue.hpp"
 
 #include <angelscript.h>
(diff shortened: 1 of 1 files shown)


CodedOre:
Just to add here, right now the Messages are dismissed after a few seconds, so I will try to add the option to dismiss it only after the wanted action is done.

Benau:
btw for "dismiss message after accomplished" I suggest you use 1 more trigger after each tutorial message, so that we assume the player finish the task if it passed the next n+1 trigger (to avoid adding too much code for checking each specific task)

you can add some function to scripting so it "bind" the trigger as an object and capture isTriggered to decide if the message should be dismissed.

PR #4204, opened 2020-01-07
Adapting MessageQueue for Tutorial

CodedOre:
This PR prepares the MessageQueue for Implementation in the Tutorial (#4187).

## Agreement
```

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dependencies*/
 CMakeFiles/
 stk-editor/
 .vscode/
+tags.*
 
 .config/
 supertuxkart-64
--- a/src/guiengine/message_queue.cpp
+++ b/src/guiengine/message_queue.cpp
@@ -457,7 +457,6 @@ void discardStatic()
 void clear()
 {
 #ifndef SERVER_ONLY
-	g_static_message->discard();
 	g_static_message = 0;
     g_all_messages.lock();
     while (!g_all_messages.getData().empty())
--- a/src/guiengine/message_queue.cpp
+++ b/src/guiengine/message_queue.cpp
@@ -457,6 +457,8 @@ void discardStatic()
 void clear()
 {
 #ifndef SERVER_ONLY
+	g_static_message->discard();
+	g_static_message = 0;
     g_all_messages.lock();
     while (!g_all_messages.getData().empty())
     {
--- a/src/guiengine/message_queue.cpp
+++ b/src/guiengine/message_queue.cpp
@@ -103,14 +103,14 @@ protected:
     /** Drawing rectangle of text layout. */
     core::recti m_text_rect;
 
-	/** A pointer to the Container for the Message */
+    /** A pointer to the Container for the Message */
     SkinWidgetContainer* m_container;
 
 public:
     TextMessage(MessageQueue::MessageType mt, const core::stringw &message) :
         Message(5.0f)
     {
-    	m_container = g_container;
+        m_container = g_container;
         m_message_type = mt;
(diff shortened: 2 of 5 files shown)


CodedOre:
In order to not clutter the GUI-Scripting with too many Functions, the first commit combines "displayGenericMessage" and "displayErrorMessage" in one, with an enum to change between Types.

qwertychouskie:
@CodedOre Full information on the tools needed to create/edit tracks is here: https://supertuxkart.net/Installing_Tools

Note that you'll need Blender 2.79, as the current scripts aren't compatible with Blender 2.8.

CodedOre:
So, i ran in some problems with creating a triggered message.
Since it would be good to use `MessageQueue` for this as well, i started to create the basic frame for it (see the last commit).

My Plan was to create an Class for static TextMessages which resets the timer (like the ProgressBar is doing). Adding is easy, however the problem lies in removing it again, since the now used `std::priority_queue` don´t allow to pick an specific part in the container, which is necessary to securely remove the wanted message.

So, i see these options:

- Prioritizing the static messages, so they can be removed with the existing system.
- Creating an own System to display Static Messages, so the Queue is avoided.
- Changing the Storage to an iterate-able container.

I would choose the last option, because we could also use this to show multiple messages at once, however i would like to hear your opinions as well, especially from @Benau.

Benau:
lol alayan doesn't like showing many messages at once (annoying!) 

you can just use iterate-able container probably

CodedOre:
So, following an advice to keep it simple, i just added an Pointer, so we can add one Static Message at once, overriding an older one. (At the end we only need one, so...).
And instead of triggering it, for now the static Messages are discard by an Function. Should do the job as well. But can still be changed.

This would leave only a few points on my ToDo-List:

- Allowing for static and temporal messages to shown both (since the render on the same spot now).
- Discard any still shown Message when closing the Track (for Example if the Player quits the Tutorial).

and... finally adapting the Track itself...
@qwertychouskie, thanks for the advice on the Scripts! It should now work.

CodedOre:
So, this Pull Request should do it.

So, i am ready for review.

Benau:
we use only 4 spaces, also why do you need to change the enum order in message queue .hpp?

CodedOre:
> we use only 4 spaces

Can you explain this in detail? Don’t understand what you mean with “4 spaces”, so I don’t know what to correct.

> also why do you need to change the enum order in message queue .hpp?

Well, if I understand it right the Order in the Enum defines also the Priority of an Message, and an Error Message should be prioritized over an Generic one in my opinion.

Benau:
that's 4 spaces for indentation 

hmm that probably the same enum order should be used in scripting as well

CodedOre:
> that's 4 spaces for indentation 

Oh, though I have this already, but I check it again... Also for the Enum Order in the Scripting Part.

CodedOre:
Alright, Identation should be fixed now.

Benau:
what is tags.* in gitignore purpose?

CodedOre:
Oh, i use Gnome Builder as IDE, and it produces these tag-files to keep track of functions, classes and these things. So i have set them to ignore for git.

Benau:
maybe delete previous g_static_message before addStaticMessage to avoid memory leak

Benau:
you still need g_static_message = 0 after delete to avoid double delete next time

CodedOre:
g_static_message = 0 **instead or additional** to delete?

Benau:
delete g_static_message;
g_static_message = 0;

PR #4213, opened 2020-01-14
Add goal to race result

### LINKED ISSUE #3877: Display challenge results in race end card
## Introduction
Currently (STK 1.0.0 and earliser), the interface isn't that very informative regarding the victory or failure of a Story Mode challenge. It can sometimes be pretty confusing as it's not clear when you failed.

## The problem
For instance, the end card does not tell the player directly if you failed or lost. You'd have to infer that from the buttons, but this is far from ideal. If you have completed a challenge *again*, the end card gives you no clue at all whether you have won or not, it looks the same. There's a lack of information.

Also, in some of the more difficult challenges it is not enough to finish first, you also must beat the clock. This can be easily missed, forgotten or misunderstood as it might be confusing that you don't unlock anything although you won the race. Improving the information would help here.

## Goal of this issue
To make it clearer to the player what has actually happened at the end of a challenge, whether you failed or not, and why you failed or won.

## Suggestions

I make the following suggestions for interface improvements:

1) In the race end card, show a line of text / title that says *directly* if you won or lost the challenge
2) In the race end card, show a list of all challenge objectives and which ones have been completed
3) 1) and 2) must work also after giving up a race. In this case, all challenge goals are counted as failures, no matter the actual progress

I have made a simple mock-up:
https://satoshiupload.com/images/2do3HV8eXb.png

In this mock-up, this would be the race end card that shows at the end of a nitro challenge which has failed. The icons and text color show the success or failure of each challenge objective. Also note the race records are gone to make space for the challenge results. Race records are not very important in this context and can be looked up elsewhere.

For the Grand Prix challenges, things are a bit different: Obviously, it doesn't make sense to show the challenge results after every race. Only after the final race, it makes sense to show the challenge results. Maybe in a separate screen after the end card of the final race.
However, I don't think this is important for now as all GP challenges are the same (just be 1st) and have no special requirement.
Long story short, it's OK to just leave the GP end carts unchanged for now.
But this is something to keep in mind for later. GP challenges might change in future should there ever be a GP challenge with a special goal (like a total time limit or a score goal).

### LINKED ISSUE #3897: best_while_slower: Make visible in GUI?
STK 1.0.0.

`best_while_slower` is a secret and hidden award (in `players.xml`) and is given if you complete a Story Mode challenge in a non-SuperTux difficulty but beat all the requirements in SuperTux level. It does not work for GPs.

It is completely secret and invisible to the player. The only moment the player sees any effect when the SuperTux difficulty unlocks early. Yet the whole Story Mode has been secretly designed around the idea of `best_while_slower` and has apparently been even balanced towards it.

So STK has a fully balanced, tested set of challenges that the player is completely left in the dark of. Maybe a missed opportunity?

Since `best_while_slower` is an unique challenge, what do you think in actually exposing it in the GUI and making it an actual award the player can get? Visibly, not secret anymore.

Ideas for exposure in GUI:

As icon, use a golden star.
In the minimap, overlay a small golden star above the cup.
In the challenge screen, display the golden star at the top right corner. If you move the cursor above this symbol, show a tooltip that explains what it means.
If the player gets this award the first time, also spawn the “unlock treasure chest screen”. A golden star is in the chest (if modelling the star is not feasible, skip this step).

CodingJellyfish:
## Agreement
```

```
STK
Fixes #3877 
Partly fixes #3897
The best while slower needs test

### DIFF
--- a/src/race/race_manager.cpp
+++ b/src/race/race_manager.cpp
@@ -80,6 +80,7 @@ RaceManager::RaceManager()
     m_track_number       = 0;
     m_coin_target        = 0;
     m_started_from_overworld = false;
+    m_started_challenge_id = "";
     m_have_kart_last_position_on_overworld = false;
     m_num_local_players = 0;
     m_hit_capture_limit = 0;
--- a/src/race/race_manager.hpp
+++ b/src/race/race_manager.hpp
@@ -238,6 +238,7 @@ public:
 private:
 
     bool m_started_from_overworld;
+    std::string m_started_challenge_id;
 
 public:
 
@@ -773,7 +774,17 @@ public:
     {
         return m_started_from_overworld;
     }   // raceWasStartedFromOverworld
-
+    // ----------------------------------------------------------------------------------------
+    std::string getStartedChallengeID() const
+    {
+		assert(m_started_from_overworld);
+		return m_started_challenge_id;
+	}
+	// ----------------------------------------------------------------------------------------
+	void setStartedChallengeID(std::string id)
+	{
+		m_started_challenge_id = id;
+	}
     // ----------------------------------------------------------------------------------------
     /** \name Callbacks from the race classes
      * These methods are to be used by the classes that manage the various
--- a/src/race/race_manager.hpp
+++ b/src/race/race_manager.hpp
@@ -777,14 +777,14 @@ public:
     // ----------------------------------------------------------------------------------------
     std::string getStartedChallengeID() const
     {
-		assert(m_started_from_overworld);
-		return m_started_challenge_id;
-	}
-	// ----------------------------------------------------------------------------------------
-	void setStartedChallengeID(std::string id)
-	{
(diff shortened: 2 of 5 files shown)


Alayan-stk-2:
Please make the code cleanup changes (mostly printLine and printSmallLine) into their own separate PR.

From the screenshot, the suggested solution suffers from contrast issues (the text is not easy to read).

Also, while exposing the secret challenge of beating the SuperTux challenges at a lower difficulty level is something that would be good long term, I'm not convinced about this implementation, which strays from the ideas discussed in the issue.

I know from having done my own PRs in the past that mixing stuff into a big PR is tempting, but it doesn't work out very well.

Wuzzy2:
What's the progress on this?

CodingJellyfish:
> What's the progress on this?

Sorry busy with study so I don't have time.

CodingJellyfish:
Seperate PR as Alayan said

PR #4226, opened 2020-02-05
Build fix on *BSD.

devnexen:
dns resolution belongs to the libc on those systems.
little server connection checking update proposal.

## Agreement
```

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,11 @@ include_directories("${PROJECT_SOURCE_DIR}/lib/bullet/src")
 if(WIN32)
     set(LIBRESOLV_LIBRARY)
 else()
-    find_library(LIBRESOLV_LIBRARY NAMES resolv libresolv)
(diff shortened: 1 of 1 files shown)


Benau:
< 1024 port should be avoid for publicly stk port?

devnexen:
well only privileged services can access those low ports ?

Benau:
btw get_port lambda is meant to return integer

devnexen:
ah true

Benau:
actually stk server document doesn't mention you cannot use < 1024 port at all (it could work if someone sudo supertuxkart --wan-server=test --port=88), so maybe either we allow this or adjust the document

also this is just server port told by dns record, nothing prevent the stk owner type anything there

Benau:
allow this == remove the 1024 check here

Benau:
little server connection checking update proposal. not true now!

Benau:
(commit message i meant)

PR #4230, opened 2020-02-12
Show kart direction by an arrow in soccer

risostk:
I think it is useful or helpful to show the direction of the kart in soccer (can do a precise shooting of the bowling ball during drifts). I remember an issue that asked to rotate the minimap by 180 degrees for blue, and qwertychouskie suggested a (fancy) way to fix it which is similar to what I did here. I cannot find that issue.
Here is what it looks like. I used a larger minimap here.
Screen Shot 2020-02-11 at 7 46 06 PM
Here is a video:
https://www.bilibili.com/video/av88362726/
It is also possible to show the arrow in CTF and other modes. But right now I only show it in soccer.

## Agreement
```

```


Benau:
probably you need to do m_icons_frame = irr_driver->getTexture("icons-frame_arrow.png"); in init, otherwise it would affect normal race gui after game mode is changed

risostk:
Thanks. Emm..., I checked different mode, and it worked. I'll double check it tomorrow. I didn't check online with other players. I'll check online later this week.

Benau:
hmm maybe i was wrong them (race gui is not like the state screens), now let see if alayan likes your idea...

risostk:
Hi Benau, there is no problem changing mode. Online also works, though I didn't check with other players.

Benau:
you need to ask crystal in discord and see if he can change the minimap to vertical...

Benau:
or maybe we can figure out if it's horizontal minimap in stk...

risostk:
Emmm... So the minimap direction is determined by the model? You mean Crystal rotated this map differently?

Benau:
yes he rotated the mnimap himself

Benau:
Maybe you can angle between the starting position (btTransform) of each kart id

risostk:
I don't think the starting position can solve the problem here: the arrow shows rotation, while the starting position gives the rigid body transformation. I think it has something to do with m_minimap_invert_x_z and how the minimap is flipped. I'll dig into it in the weekend. For a horizontal minimap, there is no need to flip it.

Benau:
maybe you can use the starting position rotation as the "base" to determine kart rotation....

risostk:
Fixed a very stupid bug. Now everything works correctly, including online :D

Benau:
maybe you can pre-rotated the .png to avoid ```icons-frame_arrow.png was rotated by 45```

risostk:
I actually tried that first. I need to make the circle smaller, to put in the arrow and keep the rotation center the same as the png's center. That does not look good. That's why I rotated it by 45 degrees :D

PR #4232, opened 2020-02-13
Migration Python2 towards Python3

marjorieB:
## Agreement
```

```


deveee:
I think that there is no need for ```#!/usr/bin/env python3```, because print() should work with python2 too

Issue #3408, opened 2018-08-28
Add the possibility to rename input configurations


Alayan-stk-2:
Along with #3241, this can be useful to quickly remember the purpose of a configuration ; as a custom name can be much more useful than a simple "Keyboard N".

GuillaumeBft:
Is it ok if I create a new dialog in "states_screens/dialogs/rename_config_dialog.hpp" ?

Benau:
can you describle what are going to do with the dialog?

GuillaumeBft:
A dialog with a textbox which contains the current name of the input configuration and below it, 2 buttons : accept and cancel.

Benau:
you can use general textfield dialog for that

GuillaumeBft:
Got it

Benau:
now it's possible

PR #4233, opened 2020-02-17
Add rename config option

GuillaumeBft:
Issue #3408
Add a button for rename a configuration.

RenameConfigOption

## Agreement
```

```

### DIFF
--- a/data/gui/screens/options_device.stkgui
+++ b/data/gui/screens/options_device.stkgui
@@ -53,6 +53,11 @@
 
                          <spacer width="50" height="2%" />
                      </div>
+
+                     <spacer width="2%" height="10" />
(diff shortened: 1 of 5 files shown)


Benau:
You probably need xmlDe/Encode to deal with " character

GuillaumeBft:
Understood, I'm on it

Benau:
maybe m_config_name can be stringw so you don't need to utf8ToWide everytime...

GuillaumeBft:
Yes, I hesitated to do it, I correct it

Benau:
there is a .empty() for string(w) instead of size() != 0

also i think you you just setConfigName without c_str()

PR #4240, opened 2020-03-11
allow 0 player number if there is AI

risostk:
It is not necessary to restrict player number > 0. If there is at least one AI in the team, 0 player should be allowed. We only need to deal with the case of 0 player and 0 AI.

## Agreement
```

```


Benau:
hmm, so your patch allows starting game if 2 splitscreen players red and no red ai, > 1 num of blue?

risostk:
Yes. I think it is natural for splitscreen players play as a team against AIs. 
The main purpose of this change is to save the AI number for the next game, even there's 0 AI.

Issue #4238, opened 2020-03-08
How to implement a new camera type for soccer (with patch)


luffah:
## Description
Many idea of camera improvement exists for soccer mode :
 - mirror to see behind
 - a camera as public for spectator
 - add items (bubblegum is nice in soccer mode) // oups it is not linked to camera
  - ...

Here i got an example of camera suitable to soccer mode that show the ball everytime.
How (where) shall i add this camera ? using subtype of camera normal or designing a new camera model.
And how shall i add the setup of the camera type ? (in a perfect world maybe it shall be switchable during the match... but with which key ... )

```patch
diff --git a/src/graphics/camera_normal.cpp b/src/graphics/camera_normal.cpp
index 9cdcb9592..5a976aefe 100644
--- a/src/graphics/camera_normal.cpp
+++ b/src/graphics/camera_normal.cpp
@@ -24,6 +24,8 @@
 #include "input/device_manager.hpp"
 #include "input/input_manager.hpp"
 #include "input/multitouch_device.hpp"
+#include "modes/world.hpp"
+#include "modes/soccer_world.hpp"
 #include "karts/abstract_kart.hpp"
 #include "karts/explosion_animation.hpp"
 #include "karts/kart.hpp"
@@ -299,13 +301,30 @@ void CameraNormal::update(float dt)
         core::vector3df current_target = (m_kart->getSmoothedXYZ().toIrrVector()
                                        +  core::vector3df(0, above_kart, 0));
         m_camera->setTarget(current_target);
-    }
-    else // no kart animation
+    }     else // no kart animation
     {
         float above_kart, cam_angle, side_way, distance, cam_roll_angle;
         bool  smoothing;
         getCameraSettings(&above_kart, &cam_angle, &side_way, &distance, &smoothing, &cam_roll_angle);
         positionCamera(dt, above_kart, cam_angle, side_way, distance, smoothing, cam_roll_angle);
+
+        World* world = World::getWorld();
+        SoccerWorld *soccer_world = dynamic_cast<SoccerWorld*>(world);
+
+        if (soccer_world) {
+          m_camera->setNearValue(1.0f);
+
+          Vec3 ball_pos = soccer_world->getBallPosition();
+          Vec3 kart_pos = m_kart->getSmoothedXYZ();
+          Vec3 to_target=(kart_pos-ball_pos).normalize();
+          if ((kart_pos-ball_pos).length() < 10) {
+            ball_pos = (kart_pos - (to_target * 10));
[...]

PR #4245, opened 2020-03-12
add a camera mode to spectate soccer in large view that follows the ball

luffah:
This PR, add a camera mode to spectate soccer in large isometric view  that follow the ball (switchable with accel/brake keys).

Related to issue: https://github.com/supertuxkart/stk-code/issues/4238 (no need to check the issue, it is just a technical question with some thougths about UI — a mirror like in _crazyracing kartrider_ —).

## Agreement
```

```

### DIFF
--- a/src/graphics/camera.cpp
+++ b/src/graphics/camera.cpp
@@ -190,6 +190,7 @@ void Camera::setupCamera()
  */
 void Camera::setMode(Mode mode)
 {
+    if (mode == m_mode) return;
     // If we switch from reverse view, move the camera immediately to the
     // correct position.
     if( (m_mode==CM_REVERSE && mode==CM_NORMAL) || 
@@ -204,6 +205,7 @@ void Camera::setMode(Mode mode)
         m_camera->setTarget(target_position.toIrrVector());
     }
 
+    m_previous_mode = m_mode;
     m_mode = mode;
 }   // setMode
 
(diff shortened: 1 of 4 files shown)


Benau:
you should check if dynamic cast fail, also if(x){ should be next line

Benau:
not sure deve wanna add android support or bike do that... 

let hope deve is not isolated because of 🇨🇳 virus....

Benau:
..still if(x) { ....

we use if (x)
{

in stk

luffah:
_China virus ? are you speaking about Tai Chi ?_
Sorry, i'm used to code formatting tools with Python.

Benau:
...if you don't mind can u use 4 spaces for indentation? (there are 2 spaces now for soccer world)

luffah:
.vimrc updated

Benau:
if you don't know the android one then probably let's merge first lol

Benau:
btw why it's if (action == PA_ACCEL || action == PA_BRAKE)?

you only need brake input?

luffah:
Similarly to left/right to change kart, i was thinking about up/down to change camera modes (to possibly add more than 2).

Benau:
then maybe you should change the translation or remove the PA_ACCEL after it's added

luffah:
I prefer first solution… but it gives an ugly message.

Benau:
maybe just use accel button for now?

Benau:
actually if you have just a few camera mode then you switch with only 1 key is quick anyway (and they don't even exist yet)

luffah:
ok only up button for now

Benau:
maybe better english will be
```
Press <%s> or <%s> to change the targeted player, <%s> or <%s> for the camera position.
```

Or alayan / andet!

Benau:
sorry for confusion... 

but "Press <%s> or <%s> to change the targeted player " ", <%s> ..." there will be an extra " " before comma.

qwertychouskie:
@Benau are you saying there should be a space before the comma?  In English there shouldn't be a space before commas, just after them. <- Example

Benau:
...there will be an extra " " before comma which is not good lol

luffah:
Whoaaa faster PR (reviews included) i never seen.

Issue #4154, opened 2019-12-01
camera forward-up-angle in normal camera mode 


luffah:
(very-minor issue) 

Does parameter forward-up-angle works as expected ?
## Description
In kart-characteristics.xml
camera forward-up-angle="60" 
have no effect on CM_NORMAL view.

## Configuration
STK release version: git fe346196525ea2d064911787b4ca107bbed89870

Benau:
This option is ignore if smoothing of camera is enabled (which only useful in CM_REVERSE and CM_CLOSEUP atm)

luffah:
It shall be added in <camera/> description.

luffah:
Too, where smoothing can be desactivated ?

Benau:
...probably static camera is too boring for racing...

you need to hack stk yourself

luffah:
Note : only backward up angle is used for CM_REVERSE. For CM_CLOSEUP, angle is hardcoded.

PR #4249, opened 2020-03-15
camera parameter forward-smoothing  and restore use of forward-up-angle 

luffah:
This PR restore the usage of _forward-up-angle_ which was unused in code because of hardcoded smoothing. It is an attempt to reconcile the hard coded dynamic camera with xml settings. Default settings change nothing to the camera.

Additionnally, smoothing when camera is pointing forward can be disabled with new parameter _forward-smoothing_.

Related to : https://github.com/supertuxkart/stk-code/issues/4154

Impact `data/kart_characteristics.xml`.
```xml 
        <!-- Camera
            Distance between kart and camera.
            forward-up-angle: Angle between camera and plane of kart (pitch)
                when the camera is pointing forward
            forward-smoothing: if true, use smoothing (forward-up-angle become relative to speed) when pointing forward
            backward-up-angle: Angle between camera and plane of kart (pitch)
                when the camera is pointing backwards. This is usually
                larger than the forward-up-angle, since the kart itself
                otherwise obstricts too much of the view. -->
        <camera distance="1.0"
                forward-up-angle="0" forward-smoothing="true"
                backward-up-angle="5" />
```

## Agreement
```

```

### DIFF
--- a/data/kart_characteristics.xml
+++ b/data/kart_characteristics.xml
@@ -180,11 +180,13 @@
             Distance between kart and camera.
             forward-up-angle: Angle between camera and plane of kart (pitch)
                 when the camera is pointing forward
+            forward-smoothing: if true, use smoothing (forward-up-angle become relative to speed) when pointing forward
             backward-up-angle: Angle between camera and plane of kart (pitch)
                 when the camera is pointing backwards. This is usually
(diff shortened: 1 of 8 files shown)


luffah:
@risostk  maybe this PR to setup the camera will interest you.

PR #4264, opened 2020-04-09
Use smoothed position for kart sounds

wb9688:
When spectating someone with a high latency, the sound was really annoying when using headphones. This seems to fix it. Thanks @Benau for telling me where to look.

## Agreement
```

```

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -1774,7 +1774,7 @@ void Kart::update(int ticks)
 
         if (!GUIEngine::isNoGraphics() && !has_animation_before)
         {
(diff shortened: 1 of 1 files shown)


src/karts/kart.cpp
```diff
@@ -1774,7 +1774,7 @@ void Kart::update(int ticks)
 
         if (!GUIEngine::isNoGraphics() && !has_animation_before)
         {
-            HitEffect *effect =  new Explosion(getXYZ(), "jump",
+            HitEffect *effect =  new Explosion(getSmoothedXYZ(), "jump",
```
wb9688, line 1777: Hmm… maybe I shouldn't change it here?

PR #4265, opened 2020-04-09
Modern skin

qwertychouskie:
The culmination of years worth of work!  Thanks to all who helped make this as amazing as possible!
PNGs were ran through `pngquant` then `optimize_data.sh` for maximum smallness.
`213 items, totalling 1.5 MiB (1,576,115 bytes)`


Benau:
Not sure @Alayan-stk-2 is isolated because of chinese virus lol.....

Benau:
data/gui/icons/android.png is not used anymore

Maybe it's time to bring back font thickness code to save ttf to make distro happy

qwertychouskie:
@Benau Icons for Android/Windows/Mac/Linux removed, we can add these extra icons to the media repo later.

qwertychouskie:
`5 items, totalling 618.4 KiB (633,272 bytes)`  As for the TTFs, distros will work their magic as they already do with the existing TTFs, and the size is small enough that Android shouldn't be really affected (especially with app bundle support coming).

Benau:
every bytes matter...

Benau:
...Alayan....

qwertychouskie:
@Alayan-stk-2 ^

@Benau Real pings generally work better ;)

Benau:
Alayan most stk fans and even deve think it's a good idea to include this so i will go ahead merge it...

Let hope the chinese virus doesn't affect you too much...

deveee:
Congratulations Qwerty!

PR #4269, opened 2020-04-16
add option to substitute look back camera by soccer ball camera

luffah:
In user config, add a bool option `reverse-look-use-soccer-cam` to use almost the same camera than spectator camera soccer except angle and distance are the ones of reverse camera mode.

To use it : set the option to "true"; start a soccer game; use rear view mirror that is no more rear view mirror in this mode, but a camera that follows the ball.

On apply to soccer mode, not completely useful (given the minimap make the job), but :
- completely stylish :laughing: 
- familiar to Rocket League users

## Agreement
```

```
## Possible future, but that's not the subject here
In order to be used by non advanced users, this parameter would be located in a section that allows to set all about camera view with presets. (future PR from me ? probably, but i know it will raise some debates about complexity of parameters)

### DIFF
--- a/src/config/user_config.hpp
+++ b/src/config/user_config.hpp
@@ -976,6 +976,11 @@ namespace UserConfigParams
             PARAM_DEFAULT(  FloatUserConfigParam(40.0, "camera-angle", &m_spectator,
                                                   "Angle between ground, kart and camera.") );
 
+    // ---- Special settings for soccer mode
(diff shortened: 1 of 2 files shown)


Benau:
```BoolUserConfigParam(false, "reverse-look-use-soccer-cam"```

luffah:
the :eagle:  eyes

PR #4270, opened 2020-04-19
replace split screen horizontal checkbox by a spinner 

luffah:
Hello !

I propose to replace the checkbox that activate horizontal splits in multiplayer mode by a spinner.

The aim is to clarify this part of the interface, because (IMO) :
 -  the user need to knows explicitly which current split in multiplayer mode.
 -  activating horizontal split is a global rendering option, not an additionnal widget like "display FPS". Therefore this option shall be placed in same group that theming.

Capture d’écran_2020-04-19_15-39-06

## Agreement
```

```

### DIFF
--- a/data/gui/screens/options_ui.stkgui
+++ b/data/gui/screens/options_ui.stkgui
@@ -53,6 +53,17 @@
 
                 <spacer width="5" height="2%"/>
 
+                <div layout="horizontal-row" width="100%" height="fit">
+                    <div layout="vertical-row" width="10f" height="fit">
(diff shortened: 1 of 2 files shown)


data/gui/screens/options_ui.stkgui
```diff
@@ -53,6 +53,17 @@
```
Staszek2312, line 53: Git

luffah:
keep the mask... and don't forget glasses to protect your eyes

Benau:
Untitled

PR #4280, opened 2020-05-02
Make formatting of song title translatable

huftis:
Different languages use different quotation marks and spacing
(e.g., « Song title » in French), so the formatting needs
to be translatable.

## Agreement
```

```

### DIFF
--- a/src/states_screens/race_gui_base.cpp
+++ b/src/states_screens/race_gui_base.cpp
@@ -626,7 +626,8 @@ void RaceGUIBase::drawGlobalMusicDescription()
     const MusicInformation* mi = music_manager->getCurrentMusic();
     if (!mi) return;
 
-    core::stringw thetext = core::stringw(L"\"") + mi->getTitle() + L"\"";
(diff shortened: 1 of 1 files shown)


Benau:
Auria true? 

And you dont mind music title still english?

huftis:
I’m not Auria, but as a translator, I would prefer if the *titles too* were translatable, just like the level names are. They are typically descriptive English titles, e.g.:

Scottish bagpipes
Ravenbridge Mansion
Landing in gran paradiso
SuperTuxKart Theme

But that is a separate issue than having the *formatting* translatable, so I’ll open a proper issue for it.

Benau:
Because Auria speaks french!

huftis:
> Because Auria speaks french!

Better mention @auriamg with a @ comment then, so they get a notification. :)

auriamg:
@Benau yes it is true that punctuation/quotes/etc. vary between languages, at least when considering their ideal format. Using the wrong quotes won't prevent anyone from understanding, but using the right quotes would technically provide a better localisation

PR #4294, opened 2020-05-22
Make some GL/GLES preprocessor checks into if statements

qwertychouskie:
The goal is to eventually allow building with both GL and GLES support in one binary.  (Also, some of the code is arguably cleaner now.)

## Agreement
```

```


Benau:
support gles and gl in 1 binary you probably need to discard glew and load all gl* function with dlopen (or with SDL!), or deve!

lheckemann:
Would dynamic dispatch (admittedly not familiar with STK's coding style/structure) maybe be better than sprinkling `if` statements everywhere? e.g. have two renderer classes (subclasses of a common virtual class) which implement this functionality and not care which is used at the call site?

qwertychouskie:
It looks like replacing GLEW with https://github.com/Dav1dde/glad might allow allow both renderers in the same binary easily, as it supports GL and GLES, unlike GLEW.

Benau:
well from sdl2 source they already have headers for both GL which functions are limited to gl2 only  

so we just need to call SDL_GL_GetProcAddress without linking libGL, or ! 

so when vulkan stk we can just use that (and the new renderer will be VK only) 
 
though the work is not started yet because of the chinese virus

Benau:
not sure  thinks https://github.com/supertuxkart/stk-code/tree/hangkang 

seems that only irrlicht renderer supports gl + gles2 and the same time  
(gles2 is not done yet)

so  will  modern renderer in vulkan soon

btw maybe we can support android 4.2 later if glad in gles works

Benau:
not sure  can  https://github.com/supertuxkart/stk-code/commit/2fc5022459b8b40d2e4b66f944fbd43f2e86c2de in android 4.2 emulator

Benau:
lol deve actually android-16 is "Android 4.1" from sdl2 which is their minimum requirement, not sure why you used android-14 in your branch...

Benau:
not sure  thinks GLAD will work like GLEW in most case...

if so  will  master branch

Issue #4066, opened 2019-09-20
Icons look blurred on a 5K display


risostk:
## Description

See the title and the following image:
![Screen Shot 2019-09-19 at 10 37 51 PM](https://user-images.githubusercontent.com/4726939/65298418-0f6a0e80-db31-11e9-875e-65dc7ae7fd0e.png)

## Steps to reproduce

- Find a 5K display
- Open STK

## Configuration

STK release version: git 63b23d9

STK Source: github.com/supertuxkart/stk-code

System: macOS Mojave 10.14.6

Graphics card: Radeon Pro Vega 56 8GB

CPU: Intel Xeon W

Gamepads/keyboards models if related to the issue: NA

qwertychouskie:
The icons are somewhat low-resolution.  AFAIK some of these have SVG sources and could be exported to a higher res, though it might eat up more valuable space, though now that STK can download it's assets on launch on mobile, maybe it's not an issue.  If @Benau gives the OK, I can make a PR with higher res icons.  The one issue, however, is that high-res icons downscaled have jagged edges, I'm not sure if there's an easy way to fix this or not (different downsampling algorithm?  dedicated HiDPI icon theme?)...

dumaosen:
Yes in 4k also a bit
But u can try the stk 0.9.3 then you will find a little blur == nothing

risostk:
@qwertychouskie Not sure if we can use SVG directly instead of PNG. Generally, SVG should have a smaller file size compared to PNG, and there is no downsampling problem. I guess the compatibility is the issue?
Btw, Do you have the SVG/any type of source files of the tux icons of different difficulty levels? I only find the PNG file:
https://svn.code.sf.net/p/supertuxkart/code/media/trunk/vector_art/tuxes_by_conso.png
The SVG files in the folder are pretty out-of-dated. We also miss icons related to new modes like soccer, ffa, ctf, and egg. Maybe @Benau can help to find out the authors and to collect the source files?

@dumaosen Actually I think the issue is not very big. I could live with it, but it will be nicer to make stk better.

Benau:
svg added

Alayan-stk-2:
SVG support is good, but the icons currently used by the game for difficulty are still blurry.

@CodingJellyfish has done work on new improved .svg based icons so perhaps we can look at merging them (all or at least part of them to begin with) in master to properly solve this.

qwertychouskie:
[...]

PR #4306, opened 2020-06-06
SVG support

risostk:
Hi, this PR provides SVG support for STK, to address issue #4066. As large screens become more and more popular, bitmap icons will look blur on them. Having a higher resolution icon will increase the size of the assets. Also need to deal with the downscaling issue mentioned by qwertychouskie. SVG can provide a solution for this.

In stkskin.xml, you can specify .svg files. Also, you can put SVG files in data/gui/icons, so STK can load the SVG file. Note that the SVG file should have the same name as the PNG file. There is no need to set a specific default size for the SVG, the code will take care of the scaling. You just need to have the correct aspect ratio.

The size of the rasterized image depends on the screen size. I used a phenomenological equation: 0.21*screen_height + 30. This gives icon size for different screen resolutions:
- 720p: 181x181 pixels
- 1080p: 257x257 pixels
- 4K: 484x484 pixels
- 5K: 635x635 pixels
- 8K: 937x937 pixels

I currently don't have access to my 5K screen, so not able to test it on a 5K resolution. I will do the test once I have access. Anyone with a large screen is welcome to test!

I haven't completely tested the SVG version of the modern skin, due to a few missing SVG files. But for the screen without any missing icons, it displayed correctly, see below. I also tested it with a skin made by myself. It worked well. I will test the modern skin once I have all the files. The modern (SVG) is a good test case.

I have tested the compilation on Linux, Android, and MacOS, without any problem. Hope anyone with access to Windows and iOS can help to check the compilation. I will fix the compilation error if there is any.

I used the nanosvg provided in SDL2_Image-2.0.5 as parser and rasterizer. I manually merged several PRs.

I am pretty happy with it. Any comments are welcome! Thanks Benau for helping with my questions!

Screenshot from 2020-06-06 16-12-01
Screenshot from 2020-06-06 16-18-04

## Agreement
```

```


Benau:
There is quite a lot of memory leak (char* x = new and no delete) 

Also i think you better make the image power of 2 for (gles 2.0) legacy pipeline, check CreationParams.legacydevice because old gl doesn't support non power of 2 texture

risostk:
> There is quite a lot of memory leak (char* x = new and no delete) 
> 
> Also i think you better make the image power of 2 for (gles 2.0) legacy pipeline, check CreationParams.legacydevice because old gl doesn't support non power of 2 texture 

Hi, any idea how to get a pointer to CreationParams in CImageLoader?

Benau:
hmm maybe it's better to let me do the legacy support for u

risostk:
Cool, I'll just push the delete. Thanks!

Benau:
new char[x], delete[] data

risostk:
>  new char[x], delete[] data

sorry, my bad

risostk:
Also benau, do you think it is possible to load exactly the required size of icons (non-legacy case), so not simply determined from the screen size? As far as I understand, STKTexture::reload (where images are loaded) is called before RibbonWidget::add (where the optimal size is determined).

qwertychouskie:
> As far as I understand, STKTexture::reload (where images are loaded) is called before RibbonWidget::add (where the optimal size is determined).

Is there any particular reason that this needs to be the order?  Seems like it could be made so that `STKTexture::reload` is inside `RibbonWidget::add`, so RibbonWidget::add` calls `STKTexture::reload` once sizes are determined, but still has the textures available to it when it starts to do stuff with the textures.

risostk:
> > As far as I understand, STKTexture::reload (where images are loaded) is called before RibbonWidget::add (where the optimal size is determined).
> 
> Is there any particular reason that this needs to be the order? Seems like it could be made so that `STKTexture::reload` is inside `RibbonWidget::add`, so RibbonWidget::add`calls`STKTexture::reload` once sizes are determined, but still has the textures available to it when it starts to do stuff with the textures.

I notice that RibbonWidget::add is called once a screen is shown/switched. But STKTexture::reload is called only once for each icon. I guess STK did this because reloading images is time-consuming?

qwertychouskie:
Most icons should only be shown once per screen, though if there are performance issues, keeping a copy of each icon/resolution pair in memory as a "cache" to copy from instead of re-rasterizing duplicate icon/resolution pairs each time could solve it.

Benau:
too many things need to be changed just for it...

Benau:
btw can you give me the modern skin data folder in svg so i can test

risostk:
Hi, you may notice that rating_star.svg is not rendered correctly. That is due to "paint-order:stroke fill markers". Seems this painting order is currently not supported. To have a correct rendering with this implementation, you have to paint stroke after fill. This icon may need some redraw. Sorry for this @qwertychouskie .

risostk:
Hi, I modified the rating_star.svg to have it rendered correctly.
[rating_star.svg.zip](https://github.com/supertuxkart/stk-code/files/4748835/rating_star.svg.zip)

Benau:
,,,where is the complete data/skins/modern.zip with only svg icons...

risostk:
> ,,,where is the complete data/skins/modern.zip with only svg icons...

I don't have the complete data. Several icons (represent different states, i.e. pressed, unpressed) do not have an SVG source. Qwertychouskie got those PNGs with GIMP. Let me just send you what I have. I simply duplicated the existing SVG for the missing icons. We need to wait for qwertychouskie for the complete set, or if I have time, I can fix them, probably in the weekend. Anyway, it should be good enough to test the code.

risostk:
[modern_svg.zip](https://github.com/supertuxkart/stk-code/files/4749824/modern_svg.zip)
Here is the SVG skin. Notice that there are still a few PNGs used.

risostk:
Hi benau and qwertychouskie, I have drawn all the missing icons. There you go.
[modern_svg.zip](https://github.com/supertuxkart/stk-code/files/4755786/modern_svg.zip)


There is still one problem. The button in the options screen is the tab image in the modern (SVG) skin, but in the modern skin it is the button image, see the following figure:

Screenshot from 2020-06-09 20-43-22
Screenshot from 2020-06-09 21-01-39

BTW, why up_down.png and wing_mirror.png are both in the icons/ folder and the icons/android/ folder?

qwertychouskie:
I'll work on this this evening.  I suspect that `resolution_of_the_SVG * upper_power_of_two(round(1080 * monitor_vertical_res))` might give a better result, especially for e.g. box borders, and I'd like to try to get rendering at native resolution working for standard icons (basically anything that doesn't use advanced stretching).

risostk:
Thx! If you try the latest code, I have disabled scaling of the box, so box border looks nicer. We could provide a list of icons that needs stretch in CImageLoaderSVG.cpp. That's a good idea!

risostk:
Updated the SVG skin. Now modern (SVG) and modern should visually be the same.
[modern_svg.zip](https://github.com/supertuxkart/stk-code/files/4767512/modern_svg.zip)
Log: update options_language.svg, tab_vert.svg, tab_vert_focus.svg

7wells:
Hi and thanks a lot for this update! If I want to test/use these changes, how do I do this? Will the above described changes go into the next git versions for all distros?

In general, I can compile STK for Android (have not yet tested compilation for Linux and Windows).

risostk:
> Hi and thanks a lot for this update! If I want to test/use these changes, how do I do this? Will the above described changes go into the next git versions for all distros?
> 
> In general, I can compile STK for Android (have not yet tested compilation for Linux and Windows).

Hey! Thanks for your interest. You can pull the code using:
```
git remote add risostk https://github.com/risostk/stk-code.git
git checkout --track -b svg_support risostk/svg_support
```
Then compile the code as normal. Let me know if you have any problems.

risostk:
Hi qwertychouskie, I do not quite understand your equation. Here is how I get mine. I measured the size of the big icons in the main menu, and did a linear fit to the data. Then I make the size of the icons slightly bigger. Here is the data. I didn't measure the data for larger resolutions. But I guess it won't be a big problem.
icon_screen_data

qwertychouskie:
`resolution_of_the_SVG * upper_power_of_two(round(1080 * monitor_vertical_res))`

The idea of taking `resolution_of_the_SVG` into account is that icons that only display small will not be rendered at a very high resolution, while large icons (main menu, etc.) will.  The current icon sizes are optimized for a 1080p-ish screen, so rendering at these sizes on these screens, and rendering a POT larger on higher-res screens will give a good result for all icons, and keep VRAM consumption under control (tiny icons won't render super high-res).  (And keeping everything POT keeps things simple on e.g. GLES2.)

That being said, I'm currently experimenting with adding preferred h/v resolution hints to the relevant calls (`getTexture()`, `loadImage()`, etc.) so that SVGs can be loaded at their native resolution when possible.

risostk:
Thanks! Now I understand. Then we need to have the default size proportional to the final one. For different screen resolution we just scale them all by a same factor.

PR #4317, opened 2020-06-24
file_manager: Emit a compiler error when no asset dir is defined for mobile platform

dos1:
I'm playing with running various games on a mobile GNU/Linux device. It's often the case that when a game has existing mobile support (Android and/or iOS), it then relies on some preprocessor directives in order to adjust some of its UI, available options etc. to mobile devices - so when I've noticed `MOBILE_STK` across the SuperTuxKart's code, I decided to try compiling it with it to see whether it will bring some benefits. It compiled fine.

However, running it had some unforeseen consequences.

In `src/io/file_manager.cpp`:

```c
#ifdef MOBILE_STK
    m_stk_assets_download_dir = getenv("HOME");
#ifdef IOS_STK
    m_stk_assets_download_dir += "/Library/Application Support/SuperTuxKart/stk-assets/";
#elif defined (ANDROID)
    m_stk_assets_download_dir += "/stk-assets/";
#endif
#else
```

As I was running a mobile STK, but neither on iOS nor Android, the `m_stk_assets_download_dir` was left at `getenv("HOME");`...

**...and turns out `m_stk_assets_download_dir` is passed in a few places straight to `FileManager::removeDirectory`, which then happily removes everything recursively from that directory.**

Maybe this fail-save check will help save some `$HOME`s. Fortunately mine had a recent enough backup.

## Agreement
```

```

### DIFF
--- a/src/io/file_manager.cpp
+++ b/src/io/file_manager.cpp
@@ -221,6 +221,8 @@ FileManager::FileManager()
     m_stk_assets_download_dir += "/Library/Application Support/SuperTuxKart/stk-assets/";
 #elif defined (ANDROID)
     m_stk_assets_download_dir += "/stk-assets/";
+#else
+#error You must set m_stk_assets_download_dir to appropriate place for your platform
 #endif
 
 #else


deveee:
For mobile linux devices you may want to use USE_GLES2=1 in cmake. The MOBILE_STK is atm. used only for android and iOS and it's used internally only. There is no such option in cmake.

Actually there are some functions that could be useful on other platforms too, eg. override_default_params_for_mobile()... But I suppose that each platform would have their own sepcific code.

deveee:
Btw. not sure if current git STK works for you, because it uses SDL2 now, so no more CIrrDeviceWayland.

Benau:
not sure deve ok to merge... 

and after android uses sdl2 later override_default_params_for_mobile can be shared

dos1:
> There is no such option in cmake.

I'm aware, of course I've added it there myself. Still, emitting an error in such case (or at least handling it gracefully) is good code hygiene and it saves time of people who will port it into new platforms in the future. Not even speaking about leaving their /home intact.

> not sure if current git STK works for you, because it uses SDL2 now

It does. I've spent some time working on SDL2's Wayland backend last year to make sure it's in good shape last year, so that's a welcome change. STK's SDL backend still needs to learn about hidpi screens though.

Benau:
by learn about highdpi you meant something like this https://github.com/Benau/stk-code/commit/e2635f594e62c69594ab2c5daba3cbff177dffe3

dos1:
@Benau haven't tested yet, but a quick glance at the code suggests that the answer may be "yes" :)

PR #4320, opened 2020-07-08
Split the num of bots from human players in server selection

risostk:
Hi, I sometimes find it is not straightforward to know if the number of players is Bot or human player by simply looking at the Players column on the Server Selection screen. The following figure shows what the code does. If a server supports Bots, the Players column will be: 
number of human players + number of bots / Max. player

Hope you like it :D

Screenshot from 2020-07-08 10-09-20

## Agreement
```

```

### DIFF
--- a/src/states_screens/online/server_selection.cpp
+++ b/src/states_screens/online/server_selection.cpp
@@ -245,7 +245,15 @@ void ServerSelection::loadList()
         if (t)
             icon = track_manager->getTrackIndexByIdent(t->getIdent()) + 2;
         core::stringw num_players;
-        num_players.append(StringUtils::toWString(server->getCurrentPlayers()));
+        int np = static_cast<int>(server->getPlayers().size()); // number of real players
+        int nc = server->getCurrentPlayers(); // number of current players (includes AI)
+        num_players.append(StringUtils::toWString(np));
+        // if there is AI player: use the format player+ai/max
+        if (nc-np > 0)
+        {
+            num_players.append("+");
+            num_players.append(StringUtils::toWString(nc-np));
+        }
         num_players.append("/");
         num_players.append(StringUtils::toWString(server->getMaxPlayers()));
         std::vector<GUIEngine::ListWidget::ListCell> row;


Benau:
This is not always true because of splitscreen players / duplicated player, you ok?

risostk:
I see. Thx!
Quick question: When I join a server with split-screen, server->getPlayers() return 0 players?

Benau:
players list will only show the online id of player name (1 player) even it's splitscreen with many players

Benau:
so basically if you don't mind sometimes these values may fail to represent the real values then we can

risostk:
Yes, basically it will fail for (1) a single id from multiple devices and (2) multiple ids from a single device with split-screen.
Here is what I tested:
- I logged in to HK3 with the same account from Android and iPad. I got 1+2/10 (what we want is 2 riso + 1 Bot / 10 Max.)
- Then I tried to log in from iPad using two accounts and split-screen. I got 0+2/10 (what we want is 2 players + 1 Bot / 10 Max.) This is a bit weird, coz I would expect server->getPlayers() to return 1 so it shows 1+2/10 for this case. Any idea?

The problem can be solved if we send the Bot number from the server, but compatibility problem...
Let me see if I can find a workaround...

Benau:
there is no workaround for them because if you another stk instance it will discard previous joined account, thats why you will never see duplicated player name in server info dialog

the only way for proper ai number is to let server send no. of ai themselves (need changes in stk-addons and server side)

risostk:
> there is no workaround for them because if you another stk instance it will discard previous joined account, thats why you will never see duplicated player name in server info dialog

ok
> the only way for proper ai number is to let server send no. of ai themselves (need changes in stk-addons and server side)

Then there will be compatibility issues for old STK clients :(

Wuzzy2:
I doubt people will understand this “X+Y/Z” syntax. Most people can't probably guess what the “+” means. Maybe a separate column would to the trick?

risostk:
> I doubt people will understand this “X+Y/Z” syntax. Most people can't probably guess what the “+” means. Maybe a separate column would to the trick?

Agree. This is not the best solution. I will leave it here or close if devs want.

Alayan-stk-2:
Between the instances of incorrect counts being shown and the confusing syntax (I agree with @Wuzzy2 that for people who don't know beforehand what it's supposed to represent it's weird, and even if so knowing if the first or second number is for player/bot isn't obvious)

I don't like much bots being advertised as players in the player count column, but this PR is not the fix we need.

PR #4345, opened 2020-08-13
Fix #3577

### LINKED ISSUE #3577: Allow to specify header font size in GUI files
Currently, the font size of UI text elements is fully determined in code.

It would however be useful to, e.g., create sub-headers, to have the ability to specify some font size information. It wouldn't be an exact number, but something relative with e.g. 3 presets allowing to have appropriate size for page title, subheader, and maybe some other element.

CodingJellyfish:
Support adding different size headers in stkgui.
After merged I have a better idea to solve the kart selection text contrast issue.


Alayan-stk-2:
The code looks good to me.

PR #4358, opened 2020-08-27
Add Render resolution (scale_rtts) slider with auto-mode 

qwertychouskie:
[UPDATED]

Auto mode adjusts based on your display's DPI in order to hit a target DPI.  Currently it will default to Always Full for render resolution, but High Quality or even Quality might make a good default, community testing is needed.

## Agreement
```

```

### DIFF
--- a/data/gui/screens/options_video.stkgui
+++ b/data/gui/screens/options_video.stkgui
@@ -34,7 +34,7 @@
                 <!-- ************ GRAPHICAL EFFECTS SETTINGS ************ -->
                 <div width="100%" height="fit" layout="horizontal-row">
                     <spacer width="3%" height="100%" />
-                    <div layout="vertical-row" width="8f" height="fit">
+                    <div layout="vertical-row" width="10f" height="fit">
                         <gauge id="gfx_level" min_value="1" max_value="8" width="100%" align="center" />
                     </div>
                     <div layout="vertical-row" proportion="1" height="100%">
@@ -46,7 +46,25 @@
 
                 <div width="100%" height="fit" layout="horizontal-row">
                     <spacer width="3%" height="100%" />
-                    <div layout="vertical-row" width="8f" height="fit">
+                    <div layout="vertical-row" width="10f" height="fit">
+                        <gauge id="scale_rtts" min_value="0" max_value="7" width="100%" align="center" />
+                    </div>
+                    <div layout="vertical-row" proportion="1" height="100%">
+                        <label id="scale_rtts_label" height="100%" I18N="In the video settings" text="Render resolution" align="left"/>
+                    </div>
+                    <div proportion="1" layout="horizontal-row" height="fit">
+                        <spacer width="5%" height="100%" />
+                        <checkbox id="scale_rtts_custom"/>
+                        <spacer width="2%" height="100%" />
+                        <label id="scale_rtts_custom_label" height="100%" I18N="In the video settings" text="Custom"/>
+                    </div>
+                </div>
+
+                <spacer width="5" height="1%"/>
+
+                <div width="100%" height="fit" layout="horizontal-row">
+                    <spacer width="3%" height="100%" />
+                    <div layout="vertical-row" width="10f" height="fit">
                         <gauge id="blur_level" min_value="0" max_value="2" width="100%" align="center" />
                     </div>
                     <div layout="vertical-row" proportion="1" height="100%">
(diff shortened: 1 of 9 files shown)


deveee:
Not sure about slider but such option would help a lot. Ideally it could calculate best resolution based on screen size, so that you will get reasonable dpi. On my phone even scale_rtts=0.5 looks ok in game and with lighting/shadows it's way better than non-advanced pipeline. And then maybe graphics level 3 with reduced resolution could be used by default for newer phones, eg. android >= 8.

Or maybe someone with more time can see if it's possible to decrease only lighting/effects resolution while still keep high resolution for models. But it would need to render all vertices twice, eg. first for lighting to rtt and then for models using that lighting texture. So not sure if this would be faster than than drawing everything at once. Currently I'm even not sure how it's done in stk.

Benau:
ok  duda can be  again 

also not sure tarball can also be

deveee:
ok, building linux package now

Benau:
Not sure can also be uploaded here... https://github.com/supertuxkart/stk-code/releases/tag/1.2

deveee:
Updated, duda should be super smart now!

And there is no default lap in ranked because it's not possible to set it in command line, so I have to mess with my script again 

And sourceforge is slow...

Benau:
do i need to upload for you instead? (I download from github)

deveee:
If you are able then sure, because I thought it hanged and now it's just 3%

qwertychouskie:
I definitely think it would be great to have an "Auto" option that is selected by default for mobile, while keeping the slider for those who want manual control.  That being said, I don't currently have an Android dev environment set up, and also I'm not familiar with SDL2's interfaces around DPI, so I'm probably not going to be the one to implement that feature :P

Calinou:
We could likely default to half resolution on mobile unless the screen resolution is lower than or equal to 1280×720.

Benau:
not sure  thinks https://github.com/supertuxkart/stk-code/tree/hangkang 

seems that only irrlicht renderer supports gl + gles2 and the same time  
(gles2 is not done yet)

so  will  modern renderer in vulkan soon

btw maybe we can support android 4.2 later if glad in gles works

qwertychouskie:
Just updated this to have an Auto mode, and to work on both Desktop and Mobile.  Thoughts would be appreciated! :)

Benau:
```
         C:\projects\stk-code\src\states_screens\options\options_screen_video.cpp(858): error C2065: 'scale_rtts': undeclared identifier [C:\projects\stk-code\build\supertuxkart.vcxproj]
```

qwertychouskie:
Should be fixed now.

Benau:
not sure deve thinks this is good enough to merge....

deveee:
Thanks for your work.

From what I noticed:
```
supertuxkart: /mnt/data/Dokumenty/programowanie/supertuxkart/supertuxkart-git/src/states_screens/options/options_screen_video.cpp:194: virtual void OptionsScreenVideo::loadedFromFile(): Assertion `m_scale_rtts_custom_presets.size() == 8' failed.
```
Also print in shader based renderer prints always 0, so I need
```
Log::info("Renderer", "Render scale: %f", (float)UserConfigParams::m_scale_rtts_factor);
```
Also now probably all other CIrrDevice* files will fail to compile. Tbh. I'm not sure if it's still possible to use them since we use SDL2, but adding
```
virtual bool getDisplayDPI(float* ddpi, float* hdpi, float* vdpi) {return false;}
```
to CIrrDeviceStub.h is simple, so better add it.

And maybe add also
```
            if (UserConfigParams::m_scale_rtts_factor < 0.3f)
                UserConfigParams::m_scale_rtts_factor = 0.3f;
```
for safety. IMO 0.3f is really small anyway.

It looks good in general. I'm building android apk now and I will look if my devices have some reasonable dpi values.

qwertychouskie:
@deveee Those issues should be fixed now.

deveee:
Merged, thanks!

Benau:
```
    scale_rtts->addLabel("30%");
    scale_rtts->addLabel("35%");
    scale_rtts->addLabel("40%");
    scale_rtts->addLabel("45%");
    scale_rtts->addLabel("50%");
    scale_rtts->addLabel("55%");
    scale_rtts->addLabel("60%");
    scale_rtts->addLabel("65%");
    scale_rtts->addLabel("70%");
    scale_rtts->addLabel("75%");
    scale_rtts->addLabel("80%");
    scale_rtts->addLabel("85%");
    scale_rtts->addLabel("90%");
    scale_rtts->addLabel("95%");
    scale_rtts->addLabel("100%");
```
It seems that "%%" in pot file may confuse translator....

Hows about the above? Do you really need it to be translatable?

qwertychouskie:
It's fine if not translated, I just assumed that translatable strings would be preferred.

PR #4361, opened 2020-08-28
network_config: use ns_name_uncompress

ericonr:
## Agreement
```

```

This allows the game to build for distros using musl libc, such as Void and Alpine, since musl does not implement `ns_name_ntop`. Does this patch keep the intended behavior for that code? I am opening this PR since it isn't clear whether this function will be added to musl, and it would be best if such changes were discussed with upstream.


Benau:
not sure  thinks cygwin works now

PR #4366, opened 2020-08-29
Debian 10 package for supertuxkart 1.2

Vitexus:
## Agreement
```

```

This is how to make working debian package using current stable Debian 10 Buster.

You can find compiled binaries in my https://repo.vitexsoftware.cz/pool/buster/backports/s/supertuxkart

I use work of Debian Sid authors for supertuxkart 1.1

It would be nice to add /debian directory in the upstream code.

Have nice day and fast cart!


Benau:
you removed too many (required) files in stk-code which is used by non-linux build (squish / mcpp...), also we don't use fribidi and raqm anymore...

and please just let the debian (vincent!) people to do the packaging.....

PR #4377, opened 2020-09-03
Imgbot optimization

Adnan1091:
All files has been compressed losslessly by using default settings.


Benau:
Seems not very reliable... https://github.com/minetest/minetest_game/pull/2735

Also no agreement!

Adnan1091:
@Benau

PR #4397, opened 2020-09-28
Remove separate menu FPS throttling

qwertychouskie:
We should let VSync/m_max_fps (which defaults to 120) do their job.  If a monitor has a refresh rate even slightly higher than 60FPS (e.g. 60.0007FPS), throttling will introduce stutters, even more so at modern high refresh rates (75, 90, 120, etc.)

This also cleans up the code, e.g. the var name `m_throttle_fps` was also being used in `main_loop.[c/h]pp` for a different function, despite having the same var name.

## Agreement
```

```


Benau:
not sure  like this

Issue #4372, opened 2020-08-31
Crashes on entering story mode


AMDmi3:
## Description
Maintainer of FreeBSD STK port here, trying to update it to 1.2.

When trying to enter story mode or run a single race exists with the error:

`[fatal  ] Scripting: Failed to create script engine.`

## Steps to reproduce
Run stk, click on 'Story mode'

## Configuration

STK release version: 1.2 

STK Source (PPA, distribution package, official binary, etc.): **official FreeBSD port being updated to 1.2** 

System: **FreeBSD 12.1 amd64**

Graphics card: **Intel(R) HD Graphics 520 (Skylake GT2)**

CPU: **Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz (2808.10-MHz K8-class CPU)**

## Additional information

stdout.log:
```
[info   ] [FileManager]: Data files will be fetched from: '/usr/local/share/supertuxkart/data/'
[info   ] [FileManager]: User directory is '/home/marakasov/.config/supertuxkart/config-0.10/'.
[info   ] [FileManager]: Addons files will be stored in '/home/marakasov/.local/share/supertuxkart/addons/'.
[info   ] [FileManager]: Screenshots will be stored in '/home/marakasov/.local/share/supertuxkart/screenshots/'.
[info   ] [FileManager]: User-defined grand prix will be stored in '/home/marakasov/.local/share/supertuxkart/grandprix/'.
[info   ] [FileManager]: Asset 0 will be loaded from '/usr/local/share/supertuxkart/data/challenges/'.
[info   ] [FileManager]: Asset 1 will be loaded from '/usr/local/share/supertuxkart/data/gfx/'.
[info   ] [FileManager]: Asset 2 will be loaded from '/usr/local/share/supertuxkart/data/grandprix/'.
[info   ] [FileManager]: Asset 3 will be loaded from '/usr/local/share/supertuxkart/data/gui/icons/'.
[info   ] [FileManager]: Asset 4 will be loaded from '/usr/local/share/supertuxkart/data/gui/screens/'.
[info   ] [FileManager]: Asset 5 will be loaded from '/usr/local/share/supertuxkart/data/gui/dialogs/'.
[info   ] [FileManager]: Asset 6 will be loaded from '/usr/local/share/supertuxkart/data/replay/'.
[info   ] [FileManager]: Asset 7 will be loaded from '/usr/local/share/supertuxkart/data/shaders/'.
[...]

PR #4398, opened 2020-09-28
Fix include path order for all bundled libraries

AMDmi3:
Follow up to #4374 (issue #4372): the simiar problem was encountered with squish, so use BEFORE for all `include_directories` of bundled libraries to avoid clashes with system includes.

## Agreement
```

```


Benau:
You sure there isn't a better way? Because it was always working in the past....

Or deve!

AMDmi3:
Better in which way any why's this not good enough? Include order is undoubtfully broken here as it allows system includes before local ones. Since the project bundles third party libraries which may be (and are) packaged in official repos, this setup immediately breaks as soon as different version of one of these libraries is installed systemwide. I've stepped into angelscript issue myself, and was reported issue with squish a day after.

The solutions are:
- not bundle any libraries (the most correct one, however it may imply extra complications)
- place local includes before all others (`BEFORE` keyword)
- place system includes after all others (`SYSTEM` keyword)

The third one is also applicable here, but it's not mutually exclusive with the second one which I've implemented. May as well implement both of them.

Benau:
where is deve....

Alayan-stk-2:
> The solutions are:
> 
>     * not bundle any libraries (the most correct one, however it may imply extra complications)
> 
Not a good option. Using system libraries instead of bundled libraries has been an endless source of dependency hell in the UNIX ecosystem, and some of our dependencies may also need custom changes that were refused upstream (for instance bullet physics). Nowadays, the disk space savings from including reasonable libraries (not QT monsters) are too small to be a justification.

AMDmi3:
> Not a good option. Using system libraries instead of bundled libraries has been an endless source of dependency hell in the UNIX ecosystem, and some of our dependencies may also need custom changes that were refused upstream (for instance bullet physics). Nowadays, the disk space savings from including reasonable libraries (not QT monsters) are too small to be a justification

From my 20 years experience of working with a lot of different packaging ecosystems, there's no such problem as "dependency hell" and bundling is always problematic, and having non-upstreamable local changes may be the only valid justification for bundling (still should be avoided). Wasted diskspace is not the worst problem it leads to (and you won't have an idea of how big it actually is unless all packages you use start bundling all their depends), the much worse problems are 
- increased maintenance cost (you basically have to deal with fixing third party code when it stops compiling or otherwise breaks for some of your users)
- increased packaging cost (package maintainers have to patch the same code multiple times for each bundled copy - for instance, both bullet, angelscript and squish have local patches in FreeBSD ports tree, and these are not applied to stk copies; investigating which configurations this breaks is a task on its own)
- security problems (the common dependency is patched in a centralized way right after a vulnerability is discovered, but this is not possible for bundled depends, as there's no even public knowledge of what is bundled where, and fixing them may be arbitrary complex as the bundled code may be arbitrary old and modified)
- build problems like this one, caused by clashes with different versions of the same library

Summarizing, bundling is against all principles of responsibility distribution, code reuse and not doing the work twice, and this is widely supported position:

https://wiki.debian.org/UpstreamGuide#No_inclusion_of_third_party_code
https://fedoraproject.org/wiki/Bundled_Libraries

Benau:
basically we try to avoid bundle library unless it not common (squish, angelscript) or really needed to be modified 

for harfbuzz freetype we don't

deveee:
From my experience working on STK itself
- if you build the game with different library versions, you can't guarantee that it works properly and you can't even test if it works as expected, because you don't know the game that well with details
- often it's way easier to apply a patch locally rather than trying to patch it upstream, especially when upstream is not interested in it or has no time to review the patch
- IMO it's more important for users that the game behaves correctly, than some vulnerabilities in libpng are fixed...
- using system libraries takes time too, bcause often newer versions are incompatible, also you have to properly test it, fix bugs etc

Anyway I'm aware that package maintainers prefer no bundling and we try to avoid it.

AMDmi3:
Thanks @deveee!
Would another PR with adding `SYSTEM` to system includes for completeness be useful?

deveee:
And thanks @AMDmi3 for the patch.
I wouldn't add SYSTEM unless it fixes some real issues.

AMDmi3:
It fixes the same issues as this PR, but from the other side. Having both is redundant for the matter of this specific problem, but future proof and consistent in general. SYSTEM has other nice properties such as not producing warnings from third party code (see `-isystem` gcc flag description).

Issue #3779, opened 2019-03-01
Rescue sometimes place the kart just in front of a banana


Alayan-stk-2:
In some tracks, there is a banana right in the middle of the road. This by itself is not an issue, but rescue is completely unaware of any track elements besides the drivegraphs, so when being rescued in some spots, the kart will be placed right in front of the banana,with no way to avoid it besides going backwards for a moment before re-acceleration.

Expected behavior : the rescue placement is made aware of bananas (and chewing gums ?) and avoid placing the kart right in front of them.

Related possibility to consider : give a short immunity after rescue.

BlueCatYellowCat:
Immunity after rescue sounds better to me, since it also help avoid other issues like the golfballs in minigolf or immediately getting hit by another player's weapon.

Benau:
if adding 2 seconds of immunity after getting rescued is ok then maybe we can do it before 2.0 (or even 1.3 if you accept tiny "lag" for server), but not in current master ![](https://cdn.lihkg.com/assets/faces/pig/sick.gif)

PR #4399, opened 2020-10-01
Added 2 seconds of immunity after getting rescued

ahellqui:
## Agreement
```

```

This solves the problem of the rescue occasionally dropping the player in front of the banana, as described in #3779. As was suggested in the first comment of the issue; I went with providing immunity to the player for 2 seconds instead of changing the rescue placement, since that also contributes to fixing further issues such as immediately getting hit by another player.

2 seconds is arbitrary, but I think it hits a nice balance between not being able to drive away, and not making it advantageous to trigger a rescue.


Benau:
this cannot be done without breaking network compatibility 

next time you can visit our irc or discord and ask for feedback before making pr

PR #4435, opened 2020-11-29
add support for gamemode

### LINKED ISSUE #4434: Support for Feral's gamemode
It would be cool if SuperTuxKart could add support for [Feral's gamemode](https://github.com/FeralInteractive/gamemode), it might help to improve the performance on weaker setups a bit.

Currently, support can be added via [3 possible ways](https://github.com/FeralInteractive/gamemode/blob/master/lib/README.md): including `gamemode_client.h` as header (requires `-ldl`), linking dynamically against libgamemodeauto or linking statically against libgamemodeauto. None of these force the user to have gamemode installed, however linking dynamically obviously requires libgamemodeauto to present somewhere (either shipped or on the system).

The easiest way is probably to add a build option (which is disabled by default) which loads the `libgamemodeauto` pkg-config file, i. e. linking against the shared library. Users compiling themselves or distros can then use it and ensure it is actually installed, without any effect on prebuild binaries.

stephanlachnit:
Closes #4434.

## Agreement
```

```


Benau:
seems that no one else in the team has interests for gamemode, sorry about that

Issue #4432, opened 2020-11-27
Mobile control : implementing auto forward


luffah:
Suggestion for moblie from a new player (addicted to other kart games):

Have an option for mobile to auto accelerate when no key is pressed.

~~Given that in STK, you could brake or reverse, i suggest to keep the backward key and disable acceleration when pressing in the blue zone.~~ Edit : forget that

Note : too, the zone to press to accelerate is unclear (wheel control mode),  hard to see where is the maximum acceleration.

luffah:
Replacing release of brake key by accel  key could be enough to experiment the feature.
I see how to implement the thing locally but in network context, it looks more complicated ... 
```diff
diff --git a/src/karts/controller/player_controller.cpp b/src/karts/controller/player_controller.cpp
index 7117d7ec1..6669d7830 100644
--- a/src/karts/controller/player_controller.cpp
+++ b/src/karts/controller/player_controller.cpp
@@ -201,10 +201,13 @@ bool PlayerController::action(PlayerAction action, int value, bool dry_run)
         }
         else
         {
-            SET_OR_TEST_GETTER(Brake, false);
-            SET_OR_TEST_GETTER(Accel, m_prev_accel/32768.0f);
-            // Nitro still depends on whether we're accelerating
-            SET_OR_TEST_GETTER(Nitro, m_prev_nitro && m_prev_accel);
+            PlayerController::action(PA_ACCEL, 65535.0f, false);
+            /* 
+             * SET_OR_TEST_GETTER(Brake, false);
+             * SET_OR_TEST_GETTER(Accel, m_prev_accel/32768.0f);
+             * // Nitro still depends on whether we're accelerating
+             * SET_OR_TEST_GETTER(Nitro, m_prev_nitro && m_prev_accel);
+             */
         }
         break;
     case PA_NITRO:
```

Benau:
actually player who wants auto acceleration can use accelerometer / gryoscope control

the design of steering wheel is not suitable for auto acceleration or will lead to confusing control ![](https://cdn.lihkg.com/assets/faces/pig/sowhat.gif)

luffah:
I got some argument on my side too :p i know you are someone busy.

Gyroscope is really boring when your screen reflect light. Too, if you brake with gyromode , you'll have to press again on accel after releasing the brake key.

I  asking to the person and tried myself.
and gave this change as a test to the person that really prefer this mode (probably because it is  similar to Mario Kart auto forward option).

[...]

PR #4437, opened 2020-12-04
Add an option to lock accel key once the key is pressed

luffah:
Related to issue https://github.com/supertuxkart/stk-code/issues/4432.

What does it change to use it :
On race starting, you still have to press on accel key (for activating the startup boost or just start to go forward).
Once you have pressed the accel key, STK will accelerate by default; you will have to brake to reduce speed or to be static.

Pros :
- for mobile users, ease the pression on accel key at max value, and relax the thumb
- for pc / keyboard users, reduce injury on middle finger due to constant pressure and (on the key pressed :p)

Cons:
- difficult to be static with this option activated
- add ~1 on linear complexity of the program ( a condition check + once you release forward or backward key, another key event is raised ) // in fact no real impact on perf

## Agreement
```

```

### DIFF
--- a/data/gui/screens/options_input.stkgui
+++ b/data/gui/screens/options_input.stkgui
@@ -44,6 +44,15 @@
 
                 <label id="help2" I18N="In the input configuration screen" text="* Which config to use will be inferred from which 'Select' key is pressed to join the game."
                        proportion="2" word_wrap="true"/>
+
+                <div layout="horizontal-row" width="100%" height="fit">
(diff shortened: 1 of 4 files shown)


Benau:
too hacky and not interesting

luffah:
> and may make accelerator / gyroscope control looks weird
isn't it that the case ?

Benau:
probably just one option (auto acceleration) in mobile control, and those handling should only be done in mobile control class (without messing up local player controller class).

messing up local player controller is

PR #4450, opened 2020-12-24
camera settings

risostk:
Hi, I made a few changes on the camera setting:
1. Put the 'ball camera in soccer' back to the custom_camera_settings UI.
2. Add the ability to customize 'camera smoothing' and 'ball camera in soccer' of Standard and Drone chase camera. Gray out the rest settings in the custom_camera_settings UI. The 'Custom...' button will be used for all three cameras.
3. Put the settings of Standard and Drone chase camera in the config file, instead of hard-coded.

Benefit of this PR:
1. The ball camera setting really belongs to the camera setting UI. We can save one line on the option UI.
2. Camera smoothing can also be customizable for Standard and Drone chase camera.
3. For advanced users, they can modify the config file to have better 'Standard' and 'Drone chase' settings for their need. I think it is better than the hard-coded settings.

Hope you like it.

## Agreement
```

```

### DIFF
--- a/data/gui/dialogs/custom_camera_settings.stkgui
+++ b/data/gui/dialogs/custom_camera_settings.stkgui
@@ -64,6 +64,17 @@
                 </div>
             </div>
 
+            <spacer height="2%" width="100%" />
+
+            <div width="100%" height="fit" layout="horizontal-row">
+                <spacer width="3%" height="100%" />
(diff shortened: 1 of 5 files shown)


Benau:
Super!

Benau:
maybe you can add a "Reset" button in the dialog so every preset can be configured and reset later 

luffah seem to want to behaviour

PR #4454, opened 2020-12-25
small fix of kart color slider and kart icon in result screen

risostk:
Hi, some small fix:
1. use default kart color instead of the hardcoded 100
2. deal with the problem of extremely small kart icon on a 4K screen, see following figure:

Screenshot from 2020-12-25 13-42-22

There are various places where the icon size are based on the original size of the icon. This can be a problem for svg icons. It is better to scale according to the text size.

## Agreement
```

```


Benau:
font->getDimension(L"0").Height is quite slow, maybe you can save the value and use that for all

risostk:
I'll find time later to clean up the icon size related code in race result GUI. But this PR is done.

PR #4476, opened 2021-01-31
Add digital speedometer and units option

RQWorldblender:
Restores and revamps the digital speedometer last seen in STK 0.8.1. It is placed above the current speedometer; its original place is taken by rank number. It can be toggled to show metric and/or imperial units, or be disabled entirely. The latter is the default behavior in new config files, preserving the current behavior while keeping the digital speedometer an option for those who want it. There are now vertical lines that clearly separate each unit to make them easier to discern. Additional options will be written to existing config files, with the default value of None. This unit option is currently not used anywhere else. an additional tip is added to remind users of this new feature.
Options:
- None (default)
- Metric
- Imperial
- Both (Metric on top)
- Both (Imperial on top)
    
Also fix a minor issue for the comment with difficulty not showing SuperTux mode in the config.xml file.

## Agreement
```

```

### DIFF
--- a/data/gui/screens/options_ui.stkgui
+++ b/data/gui/screens/options_ui.stkgui
@@ -70,6 +70,13 @@
                 <spacer width="5" height="2%"/>
 
                 <div layout="horizontal-row" width="100%" height="fit">
+                    <spinner id="units" width="10f" align="center" />
+                    <label height="100%" x="0" y="0" I18N="In the ui settings" text="Speed units"/>
+                </div>
+
+                <spacer width="5" height="2%"/>
+
+                <div layout="horizontal-row" width="100%" height="fit">
                     <checkbox id="showfps"/>
                     <spacer width="1%" height="100%" />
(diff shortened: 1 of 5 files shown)


Benau:
I hate options everywhere.....

also i don't even think we want to bring back numeric speedmeter

RQWorldblender:
This is one another one of those things that I was out of the loop on:
Why did the digital speedometer get removed between the 0.8.1 to 0.9.0 transition?

Benau:
lots of thing got removed in 0.9 too like the hit message

Benau:
maybe they are troublesome to be kept

Benau:
i personally don't mind numeric speedometer, just remove all those tips / option... and keep 1 format (km/h) to the code simple

but you need to see if alayan mind

Benau:
actually only fouks speedrunner mod speedometer is acceptable


this no good to be used everywhere......

PR #4478, opened 2021-02-07
Feature/remove old code

samuncle:
Through the years and various we accumulated way too many unused, old, deprecated properties.

## Materials.xml

Not only many of those effects are not used or broken, there is now a much easier and more powerful way to do almost the same with shaders. With the release of blender 2.8 and newer we had to anyway redo the property system and most of those old properties are already removed on the blender side.

In Antarctica I removed a lot of those property but left warning messages so addons creators will know if they are using unsupported properties.

I also quickly removed / replaced most of those properties in the assets. So it should not break (aka crash) the game.

## General properties

Clouds: Was a feature for having moving clouds on the terrain. Never really used and anyway the code was commented. Can be done again in the future with shaders.

Skydome: Was not used anywhere + also desecrated in irrlicht.

## Agreement
```

```

### DIFF
--- a/data/gui/icons/materials.xml
+++ b/data/gui/icons/materials.xml
@@ -1,10 +1,8 @@
 <?xml version="1.0"?>
 <materials>
- 
   <material name="speedback.png"    compositing="test" clampU="Y" clampV="Y" light="N" ignore="Y"/>
   <material name="speedfore.png"    compositing="test" clampU="Y" clampV="Y" light="N" ignore="Y"/>
   <material name="notes.png"             compositing="test" light="N" ignore="Y"/>
-  <material name="soccer_player_red.png"             compositing="blend" light="N" ignore="Y"/>
-  <material name="soccer_player_blue.png"             compositing="blend" light="N" ignore="Y"/>
-
+  <material name="soccer_player_red.png"             shader="alphablend" ignore="Y"/>
+  <material name="soccer_player_blue.png"             shader="alphablend" ignore="Y"/>
 </materials>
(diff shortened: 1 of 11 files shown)


Benau:
the textures/materials.xml still have compositing=xxx

Benau:
actually now that you mention addons, even addons are not officially supported i think it's better to keep them working anyway, and some addons authors are no longer actively contributed

so i suggest you remove the changes to material.cpp so i merge the rest

and with vk stk i will make a new material.cpp anyway

samuncle:
I removed them in stk-assets. They should be clean, make sure you update them.

I know they might be still used in addons, and I will still merge this pull request for a few reasons:

- Most of those changes don't prevent from loading a map. Some effect, notably transparency will not work but the map itself is still playable.
- Those options are marked for a long time as deprecated. Every time we kept them for the same reason for backward compatibility with addons. Since we never remove them there is never an incentive to upgrade the maps to a newer structure.
- They confuse artists, programmers, newcomers into what is supported and the proper way to do maps. The exporter also still support them which increase the complexity. There was in total 5 alternative ways to enable alpha blending / testing.
- If players wanna play old maps, they are welcome and can download easily older version from the game. All binaries are still provided and can easily be downloaded.
- Rewriting material.xml is cool for VK, and this is a good first step to push people into using the newer more modern format / shader system.


Sam

Benau:
your commit causes too many trouble for addons so it need to be reverted, sorry about that.... 

even for new material system I don't see any problem to keep some compatibility, next time please don't merge yourself

PR #4491, opened 2021-02-19
Support for Nintendo Switch

Mstrodl:
## Agreement
```

```

Pretty basic support for the switch, so far only things broken should be online multiplayer (because no AES-128-GCM implementation yet) and audio. Local multiplayer and singleplayer work though!
Untitled


Mstrodl:
Building instructions:
```
mkdir cmake_build
cd cmake_build
cmake -G"Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/opt/devkitpro/switch.cmake \
    -DUSE_SWITCH=ON \
    -DCMAKE_INSTALL_PREFIX=/  \
    ../
make install DESTDIR=./install
cp -r install/usr/share/supertuxkart/data /mnt/stk-data # Assumes /mnt is your switch sd card
elf2nro bin/supertuxkart bin/stk.nro
cp bin/stk.nro /mnt/whatever
```

Mstrodl:
Something to look into is how we're going to deal with compiling all the dependencies not in portlibs. I had to make some minor modifications to some of them (e.g. OpenAL) to get them to build for the switch. I suppose we could vendor them (as we do with other things in `lib/`), but I don't want to do that without first getting other people's thoughts on it.

Benau:
First please don't add unrelated change to repo (heart.png)..... 

Next I have something you may find interesting: 
https://github.com/Benau/stk-code/blob/master/cmake/BuildSTKDeps.cmake
(This can  all deps together with stiatically linked, but actually i prefer that to be available in our supertuxkart/dependencies repo, you can use that as private usage until that happens)

https://github.com/Benau/stk-code/blob/master/src/network/crypto_mbedtls.cpp
Mbedtls should be  easier because it has cmake

And I have replace OpenAL later anyway because it has bad  license

maybe we can merge mbedtls + code first

Benau:
you can try use MOBILE_STK define to make full screen startup easier (need a define for stk-assets directory) 

and i think switch doesn't need nouveau? (or that is for cracked switch linux installed )

Mstrodl:
@Benau What unrelated changes have I made?

Also, I didn't realize mbedtls had a switch build sorry.

Also we do need the `drm_nouveau` as it contains symbols needed by EGL

Mstrodl:
Additionally, I'm unsure that `MOBILE_STK` is something I want to enable on the switch builds since it doesn't seem to bring much benefit? Keep in mind the game can be played in docked (tv) mode

Benau:
You added few addLoadingIcon png......... 

So can you test if mbedtls is possible to be built with switch toolchain?

mobile stk does few things:
always fullscreen
touchscreen control default on
allow downloading hd assets (mobile stk uses 256 texture sizes)

even android tv use the same flag 

But for testing purpose this may be not needed for now 

btw maybe you can compile libresolv yourself https://github.com/aosm/libresolv, without it stk online function will be limited

(though for the record stk cannot be put into official switch repo until  licensed is , if your PR is more interesting then maybe benau will work faster to get rid of old code quicker)

Benau:
you can cd src and grep -rn MOBILE_STK and have a full look, so of those are interesting for a more  switch port

Mstrodl:
@Benau Ah, sorry. These loading icons were for testing on the switch (stdout is difficult to get sometimes). I was planning on removing them.

* Enabling touchscreen by default doesn't work very well because users will not play with touch controls when there are physical buttons. They otherwise take up space.
* HD assets should ship with it anyways, remember that this will be played on a large screen.

This libresolv appears to be taken from apple's osx codebase, I was going to look into compiling musl's libresolv.

There is no "official switch repo", it's homebrew.

Benau:
ok

Benau:
for official switch repo l meant switch equivalent "google play / app store"  

or Nintendo doesn't like it anyway because of mario kart

Mstrodl:
@Benau There is no such thing. This is homebrew.

Benau:
Nintendo eShop?

Mstrodl:
@Benau The Nintendo eShop does not contain homebrew. Homebrew is not officially sanctioned by Nintendo

Benau:
Do you have access to the official switch kit to tell if there will be too much difference between homebrew / official switch port?

like _SWITCH_ define can be shared / filesystem handling

Benau:
oh socketInitializeDefault maybe should be put in enet (look their enet initialize for win32)

Benau:
actually libcurl should already do that too (not fixable by stk-code), i assume that function will do "allow socket connection at first, otherwise all socket function returns -1")

Mstrodl:
@Benau right. It's needed for libcurl. If I had access to the official switch sdk I'm not sure I could say because of NDAs. The real sdk probably wouldn't allow the use of the homebrew sdl2 port.

Also: mbedtls **is** supported on the switch, if you want to make a PR that would be appreciated.

Benau:
then probably the libcurl in switch homebrew should fixed later (I assume you use the prebuilt one in homebrew?) (maybe somewhere in curl initialize like windows curl)

in our windows stk we don't need WSA_***

well i can do the mbedtls for you later to save your time

Mstrodl:
@Benau I think it was a deliberate choice in the case of libcurl. I've talked with another homebrew developer and they seem to agree they would put it in main. Also it's good to have it there because of nxlink debugging

Benau:
ok for main

and mbedtls is used in master so you can adjust your pr now

Mstrodl:
I got mbedtls working it seems, but I think there's some issues with UDP on the switch I still need to sort out.

Mstrodl:
Seems specifically to be that UDP packets for internet games don't get received by the switch.

Benau:
if it works in LAN (192.168.** as in the 1st screenshot) then globally should work too, only difference between them is the encryption (AES) for stk, how do you determined then UDP is not received by switch?

Mstrodl:
@Benau I added printfs around the `recvmsg` and `poll` calls in ENet. I get packets back when running a lan server, but not when running in "online" mode even though my computer is on the same network.

Benau:
maybe you can visit our irc / discord so we don't need to spam people here

Mstrodl:
@Benau Is there a discord link somewhere?

Benau:
https://discord.gg/m272UTf

Mstrodl:
Wanted to make a check list to show what still needs to be done

- [x] Online multiplayer
- [ ] Audio without threading problems (game can freeze while loading maps if audio is enabled)
- [ ] Touch buttons (buttons on menu don't respond to touch for some reason, touch racing controls work though)
- [ ] Buttons matching labels (currently, the switch controllers are using the xbox or playstation button layout, 'A' is labelled as 'B' and so on)
- [ ] Write some build script and figure out how to distribute library patches

Mstrodl:
OpenAL-Soft has landed changes in master meaning we don't need a patch for it anymore. See: kcat/openal-soft#536

Still pending: harfbuzz/harfbuzz#2881

I'm writing a build script for switch builds. Only thing from there is to wait on devkitpro/sdl#71 (follow game controller labels hint)

Benau:
actually not sure some upstream will dislike ```_SWITCH_``` define as it may conflict real switch ndk (nda!)

Mstrodl:
@Benau Fair point. I think in either case, neither of these changes do anything that would harm the switch. In OpenAl-soft's case, this was just to get a reference to the current executable which isn't possible on the switch anyway and on HarfBuzz, it was simply including climits which (as far as I know) is a pretty standard header file.

Mstrodl:
Harfbuzz changes landed in master (See harfbuzz/harfbuzz#2882). I'm going to work on getting a VM set up to test my build script and after that borrow the switch again to verify everything still works. After that, I think it should be safe to merge(?)

PR #4498, opened 2021-03-04
Support broadcast address on switch (and better service calls)

Mstrodl:
## Agreement
```

```


Mstrodl:
Oh shoot. I should just make a separate branch... Oops.

PR #4500, opened 2021-03-07
Discord rich presence

Mstrodl:
## Agreement
```

```
Image of a Discord profile playing SuperTuxKart. Normal Race (medium). Nessie's Pond. 00:51 elapsed


Mstrodl:
I would appreciate if someone could try building this for Android and iOS. I think they should work, but best to be sure :)

Mstrodl:
Marking as draft because we should fix the JSON serialiser. Right now things will break if you have (e.g. a `"` in your username)

PR #4512, opened 2021-04-01
Haiku graphics mode scheduler low latency rather power saving for

devnexen:
insuring best performance.

## Agreement
```

```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -166,6 +166,10 @@
 #include <jni.h>
 #endif
 
+#ifdef __HAIKU__
+#include <kernel/scheduler.h>
+#endif
+
 #ifdef __SWITCH__
 extern "C" {
   #include <sys/iosupport.h>
@@ -2348,6 +2352,9 @@ int main(int argc, char *argv[])
 #ifndef SERVER_ONLY
         if (!GUIEngine::isNoGraphics())
         {
+#ifdef __HAIKU__
+            set_scheduler_mode(SCHEDULER_MODE_LOW_LATENCY);
+#endif
             // Some Android devices have only 320x240 and height >= 480 is bare
             // minimum to make the GUI working as expected.
             if (irr_driver->getActualScreenSize().Height < 480)


Benau:
@panos is it true?

panos:
Wow, I didn't even know about this! I'll give it a check later today.

panos:
LGTM!

pulkomandy:
Hi, I think this change is not correct.

First of all, this is a global setting for the OS scheduler, and not something that apps should be changing without telling the user.

Second, the low latency mode is the default in Haiku. So, this will not change anything, unless the user had put the system in power saving mode. And it seems not right to ignore the choice of the user here (this setting is available in the ProcessController menu).

If you want to improve performance, you may consider changing the priorities of threads. But changing a global system setting like this does not seem right to me.

Benau:
not sure if there are more haiku developers can clarify....

panos:
[I specifically sent an e-mail to the mailing list for more feedback](https://www.freelists.org/post/haiku/SuperTuxKart-Call-for-testing-performance-update-on-Haiku), wasn't aware that this wasn't global OR that SuperTuxKart would now arbitrarily take people off power saving mode, which would explain why I noticed a performance boost. @Benau 

It may be a good idea to revert this change for now.

panos:
I think that telling the system "Hey, it may be a good idea for you to push the performance up, if you can", rather than barging in and being all like "HEY YOU SILLY UNDERLING **I MAkE thE rUles here** WHY ARE YOU PERFORMING SO SLOWLY **I'M HERE** YOU BETTER FIX THIS" is *probably* more correct?

I spotted the following information that may be useful to you, if you're willing to give this another shot, @devnexen.

- [BeBook - The Kernel Kit - Threads And Teams (`suggest_thread_priority())`](https://www.haiku-os.org/legacy-docs/bebook/TheKernelKit_ThreadsAndTeams.html#suggest_thread_priority)
- [SuperTuxKart - `src/main.cpp`](https://github.com/supertuxkart/stk-code/blob/087ad350d7364b7239c44f86e5e0690fc939c6cc/src/main.cpp#L169)

(P.S. Let's not force a max framerate.)

After you make the changes (if you agree to do this), I'll follow and issue a few additional patches for better integration with the system.

Benau:
relax!

Benau:
Lol actually doesnt that require root privillege to overwrite system config!

pulkomandy:
No, in Haiku the default user is root and can do everything.

Setting appropriate thread priorities (if not done already) would be a better idea. Note that this is already done by default for threads that run a display window (they get B_DISPLAY_PRIORITY) but maybe for a game it makes sense to use B_URGENT_DISPLAY_PRIORITY or even B_REAL_TIME_DISPLAY_PRIORITY.

Setting up all threads correctly may require some experimentation and benchmarking to see what's best

pulkomandy:
> [I specifically sent an e-mail to the mailing list for more feedback](https://www.freelists.org/post/haiku/SuperTuxKart-Call-for-testing-performance-update-on-Haiku), wasn't aware that this wasn't global OR that SuperTuxKart would now arbitrarily take people off power saving mode, which would explain why I noticed a performance boost. @Benau
> 
> It may be a good idea to revert this change for now.

I thought I had replied there already but my email was lost or I forgot to send it or something.

panos:
> I thought I had replied there already but my email was lost or I forgot to send it or something.

Either way, two is better than one!

devnexen:
> [I specifically sent an e-mail to the mailing list for more feedback](https://www.freelists.org/post/haiku/SuperTuxKart-Call-for-testing-performance-update-on-Haiku), wasn't aware that this wasn't global OR that SuperTuxKart would now arbitrarily take people off power saving mode, which would explain why I noticed a performance boost. @Benau
> 
I feel terrible for that sorry



> I think that telling the system "Hey, it may be a good idea for you to push the performance up, if you can", rather than barging in and being all like "HEY YOU SILLY UNDERLING **I MAkE thE rUles here** WHY ARE YOU PERFORMING SO SLOWLY **I'M HERE** YOU BETTER FIX THIS" is _probably_ more correct?
> 

A bit emotional response I see :) Sorry for the surprising performance boost ... also did not know it was necessarily in a global scale, a bit unknown/undocumented feature known mostly from haiku os developers. Even you did not know before ;-)
> I spotted the following information that may be useful to you, if you're willing to give this another shot, @devnexen.
> 
> * [BeBook - The Kernel Kit - Threads And Teams (`suggest_thread_priority()`)](https://www.haiku-os.org/legacy-docs/bebook/TheKernelKit_ThreadsAndTeams.html#suggest_thread_priority)
> * [SuperTuxKart - `src/main.cpp`](https://github.com/supertuxkart/stk-code/blob/087ad350d7364b7239c44f86e5e0690fc939c6cc/src/main.cpp#L169)
> 
> (P.S. Let's not force a max framerate on the system's end.)

Good point but as said before me, it might take time to adjust and will it play accordingly for everyone (also people using haiku on real hardware vs haiku on virtual environment ? etc). Some settings mentioned (e.g. B_URGENT_DISPLAY_PRIORITY) comes as what cost ?

I won t be able to look into it for some days if anyone want to start, no issue from me.
> 
> After you make the changes (if you agree to do this), I'll follow and issue a few additional patches for better integration with the system.

pulkomandy:
No problem, from my point of view, all efforts in improving the experience for Haiku users are welcome. If anything it shows that we should improve our documentation on APIs.

Setting the thread priorities (even to realtime ones) should not have any unwanted side effects unless you just set everything to the max priority. In BeOS, the realtime threads would always run first, and only the remaining cpu time after they are all done would be distributed to the other threads. But I think the current scheduler in Haiku isn't as strict in doing that.

The idea is to give the game a priority above other applications and system things, but not so much that it would completely block other important system services. "realtime display" is the lowest realtime priority so it should be fine for this use. Higher priorities are used, for example, for audio processing in the media kit. This is because audio frameskip is way more noticeable than video frameskip.

One thing to check is, what priority are opengl rendering threads using? But I think tthis will be handled by Mesa and out of control from SuperTuxKart?

panos:
> I feel terrible for that sorry

No, it's fine, I just said that as a response as to whether there's more Haiku developers to @Benau, and still asked for more feedback because I'm far from an expert. @pulkomandy jumped in as a result! :D (Excuse me if I inadvertently sounded passive aggressive.)

> One thing to check is, what priority are opengl rendering threads using? But I think this will be handled by Mesa and out of control from SuperTuxKart?

Well, 3D rendering is not the only aspect of the game that's in scope as far as the need to improve performance is concerned, there's also a lot of logic that would operate in a more optimal manner if it possessed more resources to it. I think it makes sense to prioritize the game, considering that there's not quite much that the user would do in the meantime.

pulkomandy:
> Well, 3D rendering is not the only aspect of the game that's in scope as far as the need to improve performance is concerned, there's also a lot of logic that would operate in a more optimal manner if it possessed more resources to it. I think it makes sense to prioritize the game, considering that there's not quite much that the user would do in the meantime.

Priorize the game vs what else? I suspect that if you want good performance, you are already making sure that you did not let a compiler, bitcoin or other CPU-intensive task run in the background, or if you have such things running, that they are set at low priorities already and only use the CPU when everything else is idle.

You could always artificially increase CPU priorities for one specific app but it would only result in an "arms race" where everyone will just set higher and higher priorities to be scheduled first.

It's precisely to avoid this situation that our API defines a few standard constants that give a good general idea of what should be done at each priority level.

Assuming a generally idle CPU, the question is more of which threads inside the SuperTuxCart process should get higher priorities than others,and which should get "realtime" priorities. This is where you can hint the scheduler about which parts are more timing critical, and which can be delayed a little. And this is how you get maximum performance: not by setting arbitrarily higher priorities (that doesn't magically make your CPU able to do more things), but by using relatively lower priorities for the less important tasks.

panos:
I don't remember seeing a *lot* of multi-thread processing last time I checked, unless if the task manager fooled me or if I forgot to enable an option. I'll do some more testing and get back.

PR #4514, opened 2021-04-06
Add Switch to Github actions

Mstrodl:
This should generate [builds for STK for Switch](https://github.com/Mstrodl/stk-code/releases/tag/git20210405) in a Github action

## Agreement
```

```


Benau:
not reply from irc! 

so copying here...

any test reply from park?

also i think you miss ```github.repository_owner != 'supertuxkart``` checking (see main.yml) otherwise i think every forked repo will  a switch package too which may not what they want

btw I would like to use lowercase "switch" instead of "Switch" for consistency (other builds use mac, win... too)

Mstrodl:
@Benau should be okay :)

PR #4524, opened 2021-05-02
haiku elevate the thread priority from B_DISPLAY_PRIORITY to B_URGENT…

devnexen:
…_DISPLAY_PRIORITY attempt

## Agreement
```

```

### DIFF
--- a/src/graphics/sp/sp_texture_manager.cpp
+++ b/src/graphics/sp/sp_texture_manager.cpp
@@ -25,6 +25,10 @@
 #include "utils/string_utils.hpp"
 #include "utils/vs.hpp"
 
+#ifdef __HAIKU__
+#include <kernel/scheduler.h>
+#endif
+
 #include <string>
 
 namespace SP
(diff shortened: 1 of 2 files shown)


devnexen:
@panos

panos:
Wouldn't setting `B_REAL_TIME_DISPLAY_PRIORITY` be more correct?

devnexen:
isn t it more for tasks like audio processing ?

devnexen:
see no real difference in my side but I do not run haiku on real hardware tough.

panos:
[The Be Book](https://www.haiku-os.org/legacy-docs/bebook/BLooper.html) says that it should be used, but *only* the threads that get involved with rendering, which is also something that Adrien also brought up before. Not 100% sure about this.

Benau:
changing texture thread is useless because it's only for loading texture 

tbh i'm not in favour of premature optimization

devnexen:
I also think it changes nothing regardless, the default priority is ok as is.

panos:
> I also think it changes nothing regardless, the default priority is ok as is.

Thing is, how does the system react under heavy load?

devnexen:
I think the best option is someone having haiku installed in real hardware developing it and assessing os reaction under heavy load.

PR #4530, opened 2021-05-12
Fix switch hacks

WinterMute:
## Agreement
```

```

Hi. This is an attempt to fix several issues with the recent switch additions likely to cause conflicts with upstream dependencies provided by devkitPro.

- openal-soft and harfbuzz have been added to the devkitPro portlibs so no longer need to be built here.
- taking scripts from our repositories and applying them manually to a toolchain install is likely to cause issues in the future. I appreciate there's a check to attempt to prevent overwriting but these particular scripts weren't intended for use outside our own package building and we now have a separate cmake package intended for this use case.
- I don't think scripting toolchain installation in projects is a good idea, this is something that should be done once by toolchain users and updated at their convenience. The docker image we provide should be used for CI and ideally left alone.

Unfortunately I'm not familiar ennough with cmake to fix all this properly and I'm struggling to figure out why it appears to be able to find libraries when configuring the build but fails to find some libraries when doing the final link. Any pointers would be greatly appreciated.


switch/make.sh
```diff
@@ -2,105 +2,22 @@
...
-
 echo "Compiling STK"
 
 if [[ ! -d "${STK_DIR}/cmake_build" ]]; then
   mkdir "${STK_DIR}/cmake_build"
 fi
 cd "${STK_DIR}/cmake_build"
 
-cmake -G"Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="${DEVKITPRO}/switch.cmake" \
+"${DEVKITPRO}//portlibs/switch/bin/aarch64-none-elf-cmake" -G"Unix Makefiles" \
```
Mstrodl, line 19: extraneous `/`!

Mstrodl:
@WinterMute With regard to taking the toolchain files from the repo: I took them because at the time there was no cmake package or anything, so I couldn't think of any better way to do it at the time...

I appreciate you taking the time to try and do this. I had toolchain stuff install automatically because I figured most STK fans are not interested in setting up a devkitPro toolchain with the correct dependencies, and because of the devkitPro attitude of "use the repo and nothing else".

One other little thing: the toolchain will only ever be updated if the user does it themselves. This is because we don't ever update the Pacman repos. In other words, devkitPro will always remain at the same version the user has installed.
Also, Pacman is not designed to keep mismatched package versions (unlike apt, there's no supported downgrading) so I'm not sure this should be a problem for anyone.

With regard to linking: I was running into the same issues, hence the hacky thing with compile options. It's really important (as you likely know) that all the libraries are linked in the proper order. `-Wl,--verbose` compiler option might be of use :)

WinterMute:
> @WinterMute With regard to taking the toolchain files from the repo: I took them because at the time there was no cmake package or anything, so I couldn't think of any better way to do it at the time...

From my perspective it's always better to try & talk to me about the best way to do things before you go merging it into a project repo. I get that people just want to get stuff done & move on. Sometimes I wish they'd slow down just a little bit. A lot of things make me go "jfc, don't do that" and it's really, really hard to explain why if you haven't seen what happens when 10s of 1000s of homebrewers suddenly get bitten by technical debt. 

I find that a lot homebrewers have a tendency to just copy stuff they see without bothering to think too much about whether or not it might be the right way to do things. Things propagate really quickly sometimes & I end up spending far too much time trying to get people to go back and fix something they don't think is a problem because "it was working last week". 
 
 I have been looking for an opportunity to try and look at trying to clean up some cmake stuff and try to make sense of it for a while though, so swings & roundabouts.
 
> I appreciate you taking the time to try and do this. I had toolchain stuff install automatically because I figured most STK fans are not interested in setting up a devkitPro toolchain with the correct dependencies, and because of the devkitPro attitude of "use the repo and nothing else".

We generally just try to discourage people from using things supplied by 3rd parties because it unfortunately always seems to be a horrific mess. Primarily we'd like people to submit PRs for the libraries they'd like to use - thanks for openal & harfbuzz btw, it's much appreciated.

> One other little thing: the toolchain will only ever be updated if the user does it themselves. This is because we don't ever update the Pacman repos. In other words, devkitPro will always remain at the same version the user has installed.

Their toolchains and libraries will remain at the same version ... devkitPro is and will always be the organisation that provides the tools :P It is up to them though, much as I'd prefer everyone was up to date all the time it's not reasonable to be forcing updates that they may not yet be ready for. I'd be pretty annoyed if I tried to build something & it started updating my system before it even starts compiling. 

> With regard to linking: I was running into the same issues, hence the hacky thing with compile options. It's really important (as you likely know) that all the libraries are linked in the proper order. `-Wl,--verbose` compiler option might be of use :)

I appear to be almost there. This looks like it's probably an issue with my switch-openal-soft package - I'm not sure readlink is really of much use on a switch given that we really only have access to a FAT filesystem atm.

```
Scanning dependencies of target supertuxkart
[ 44%] Linking CXX executable bin/supertuxkart
/opt/devkitpro/devkitA64/bin/../lib/gcc/aarch64-none-elf/11.1.0/../../../../aarch64-none-elf/bin/ld: /opt/devkitpro/portlibs/switch/lib/libopenal.a(helpers.cpp.obj): in function `GetProcBinary()':
/home/davem/projects/devkitpro/pacman-packages/switch/openal-soft/src/openal-soft-1.21.1/alc/helpers.cpp:260: undefined reference to `readlink'
/opt/devkitpro/devkitA64/bin/../lib/gcc/aarch64-none-elf/11.1.0/../../../../aarch64-none-elf/bin/ld: /home/davem/projects/devkitpro/pacman-packages/switch/openal-soft/src/openal-soft-1.21.1/alc/helpers.cpp:267: undefined reference to `readlink'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/supertuxkart.dir/build.make:6559: bin/supertuxkart] Error 1
make[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/supertuxkart.dir/all] Error 2
make: *** [Makefile:130: all] Error 2

Did you run into this? It's possible devkitA64 r17 has changed something that the configury for openal picks up though. Maybe you were using r16?
```

WinterMute:
Got it
```
[ 56%] Linking CXX executable bin/supertuxkart
[100%] Built target supertuxkart
```

Just needed to patch out some code in openal-soft that will never work on a Switch.

Still not 100% happy with how this works - ideally I'd like to see a standard cmake && make pattern just work™. Needing to run a script first violates the principal of least surprise for users of cmake on other platforms.

@Mstrodl if you poke fincs on Discord again he can give you my discord ID. I can't seem to add you from my side & I don't want to hand out my ID in public.

Mstrodl:
@WinterMute Actually, with regard to your openal-soft patches: it might be better to use [the patches which are already in master](https://github.com/kcat/openal-soft/commit/4179633440cbe418174c8d4a0c359f337995a394) which fix this issue. They aren't in a release version yet, but it probably makes the most sense to follow that scheme (which basically does what you described).

.github/workflows/switch.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           rm stk-assets-full.zip
       - name: Restore modified date
         run: |
-          sudo apt install -y git-restore-mtime
+          sudo apt update && sudo install -y git-restore-mtime
```
DavidBuchanan314, line 32: > `sudo install`

Typo? I think this should have been `sudo apt install`

Mstrodl: You're right

WinterMute: Oops, clearly I wasn't paying attention. I've added git-restore-mtime to the docker image now though so this isn't needed any more.

Benau:
is your PR ready so I can test it?

WinterMute:
> is your PR ready so I can test it? 

@Benau should all build now. I'd still like to get nro building to work via cmake rather than using extra external steps and look at using relative paths to find data on switch. We can defer that to another PR if you like though.

WinterMute:
> @WinterMute Actually, with regard to your openal-soft patches: it might be better to use [the patches which are already in master](https://github.com/kcat/openal-soft/commit/4179633440cbe418174c8d4a0c359f337995a394) which fix this issue. They aren't in a release version yet, but it probably makes the most sense to follow that scheme (which basically does what you described).

Thanks for pointing to that, I've updated the packaging with upstream patch https://github.com/devkitPro/pacman-packages/commit/3697a768e984523cf0813cdf1f8a07e2b0697445

Benau:
```
CMake Error at cmake/FindOggVorbis.cmake:14 (pkg_check_modules):
  Unknown CMake command "pkg_check_modules".
```
 apple failed...

WinterMute:
> ```
> CMake Error at cmake/FindOggVorbis.cmake:14 (pkg_check_modules):
>   Unknown CMake command "pkg_check_modules".
> ```


Yeah, oops, sorry. Is pkg-config available for macOS on the runners or is it better to just not use it there? I know it's not available on vanilla macOS now but I guess people are using brew to build there? I've turned it off with https://github.com/supertuxkart/stk-code/pull/4530/commits/a63c9477f5354b12dd6efb4454dc8af3a8e8f413 but equally I could juse include the module at https://github.com/supertuxkart/stk-code/pull/4530/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR51

Benau:
probably because stk doesn't do brew install pkg-config to save time 

our osx deps were self-compiled

.github/workflows/switch.yml
```diff
@@ -16,46 +16,22 @@ jobs:
...
           rm stk-assets-full.zip
       - name: Restore modified date
         run: |
-          sudo apt install -y git-restore-mtime
           cd stk-code
           git restore-mtime .
-          cd lib/openal
-          git restore-mtime .
-          cd ../../lib/harfbuzz
-          git restore-mtime .
```
Mstrodl, line 57: Why was this removed? This breaks caching (make uses last modified dates) leading to really long compile times

Mstrodl (changes requested):

.github/workflows/switch.yml
Mstrodl: Other than this I don't see any big issues

Mstrodl (approved):

.github/workflows/switch.yml
WinterMute: This is just removing the libs that were being compiled but are now packages in the devkita64 docker. restore mtime is still there for stk-code - https://github.com/WinterMute/stk-code/blob/fix-switch-hacks/.github/workflows/switch.yml#L33

Mstrodl: Yeah sorry, I realised that, I thought I left a comment saying that when I approved the review

PR #4532, opened 2021-05-15
Rumble on collisions

Mstrodl:
## Agreement
```

```

### DIFF
--- a/src/input/input_manager.cpp
+++ b/src/input/input_manager.cpp
@@ -118,6 +118,14 @@ InputManager::InputManager() : m_mode(BOOTSTRAP),
         Log::error("InputManager", "Failed to init SDL game controller: %s",
             SDL_GetError());
     }
+
+#if SDL_VERSION_ATLEAST(1,3,0)
+    if (SDL_InitSubSystem(SDL_INIT_HAPTIC) != 0)
+    {
+        Log::error("InputManager", "Failed to init SDL haptics: %s",
+            SDL_GetError());
+    }
+#endif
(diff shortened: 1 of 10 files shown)


Mstrodl:
Code is certainly not ideal at the moment, we should probably reevaluate how to make this work in a way that isn't so hacky...

Benau:
btw Mary not sure if you saw in irc but it builds fail with SERVER_ONLY=ON....

Benau:
I  the SERVER_ONLY build for you! Also made some header changes and fix possible duplicated rumble for networked games 

just 1 question
```c++
#if SDL_VERSION_ATLEAST(1,3,0)
    if (m_haptic)
    {
        SDL_HapticRumblePlay(m_haptic, (strength_low + strength_high) / 2, duration_ms);
    }
    else
#endif
    {
        uint16_t scaled_low = strength_low * pow(2, 16);
        uint16_t scaled_high = strength_high * pow(2, 16);
        SDL_GameControllerRumble(m_game_controller, scaled_low, scaled_high, duration_ms);
    }
```

did you do that for compatibility reason with old SDL2 < 1.3.0? because only reason m_haptic is null is the joystick doesn't support rumble, and calling SDL_GameControllerRumble should not do anything 

and by the way we have a return; earlier in sdlcontroller constructor so I think m_haptic should be initialised earlier too

Mstrodl:
@Benau SDL Haptics I think aren't defined at all in those older SDL versions, so the field won't be there

PR #4535, opened 2021-05-22
Automatically update list when search string is changed in addons_screen

heuchi:
This adds the feature requested in the forum: https://forum.freegamedev.net/viewtopic.php?f=17&t=17734

The add-on list is updated whenever the search string is changed.
Since this is my first pull request here, I hope I've not made any mistakes concerning coding style etc.

## Agreement
```

```


Benau:
Yes nice PR 

-------
maybe  can  linux build after merging....

also not sure  wanna  duda to rpi4 to save power and avoid lag if pc is under full usage

PR #4537, opened 2021-06-02
Show kart color on icon list

heuchi:
This adds an option to show kart colors in the kart icon list during normal and time trial races and on result screens.

In the last few weeks I repeatedly heard online players saying, they would appreciate it a lot if kart colors were shown during races. Most importantly that allows to differentiate between players using the same kart class. "Pidgin" for example is used a lot in online races and so it's often hard to tell, which pidgin icon in the list belongs to which player. Knowing who's actually leading the race for example can make a huge difference when you have to decide if or if not to release a basketball.

Comments and suggestions are very welcome.

Here are two example images, so you can see how it looks.
Notice the two "Beastie" karts with two different colors:
KartColorDemo_Race01

Option:
KartColorDemo_Option01

## Agreement
```

```


Benau:
STK prefers a little option as possible 

Maybe you can draw kart color in "circle" like team color if kart color not original and make it always enable

heuchi:
Ok, I'll do a circle version.

Benau:
oh by the way alayan is  so many you can give the ranking points difference between your patch / non-patched stk so I can determine myself if it's correct

pseudo values will do fine

heuchi:
About the ranking thing:

Short version: The current implementation does not do what it claims to. However, the intended way (which I wrote to you) will lead to even more unexpected (and also questionable) results. I therefore thought it might be best to completely remove the "extra reduction", but that would not change a lot, and so maybe it's best to just leave it as it is.

Long version: Here's a PDF, that tries to dig a little deeper and show the three different options.
[STK-ER-Problem.pdf](https://github.com/supertuxkart/stk-code/files/6589280/STK-ER-Problem.pdf)

heuchi:
I've updated the PR to use circles on karts with custom color set.
It looks like this (the 2nd icon is from an AI kart which has no custom color set, so it doesn't get a circle).
KartColorsCircle02
However, on discord NickerAban had the idea to use colored icon-shaped frames instead of circles. He provided two demo shapes for Tux and Pepper which I added for a test:
KartColorsFrames01
If there is no colored frame for a kart, it uses the circle fallback.

What do you think?

heuchi:
Oh, btw: The code for the icon-shaped frames is here (didn't want to mix up things):
https://github.com/heuchi/stk-code/tree/show-colored-icons-on-kart-list

Benau:
Color frame is not so sweet

NickerAban:
> El marco de color no es tan dulce 

I was thinking about that, to look like a tux icons shape

NickerAban:
I dont know if that's okay
How about when a player has a kart color
for example: tux has blue kart color, left side has tux shape but hidden icons in blue

[https://cdn.discordapp.com/attachments/376132466152308738/849640702392401930/tux_color_kart_icon.png](url)

NickerAban:
hm..

NickerAban:
I was thinking, when a player has a kart color
for example: tuxedo is kart color blue, left side is tuxedo shape but hidden icons are blue

NickerAban:
> I've updated the PR to use circles on karts with custom color set.
> It looks like this (the 2nd icon is from an AI kart which has no custom color set, so it doesn't get a circle).
> KartColorsCircle02
> However, on discord NickerAban had the idea to use colored icon-shaped frames instead of circles. He provided two demo shapes for Tux and Pepper which I added for a test:
> KartColorsFrames01
> If there is no colored frame for a kart, it uses the circle fallback.
> 
> What do you think?

like this

Benau:
can you just use the icons-frame.png instead of a new png?

NickerAban:
> can you just use the icons-frame.png instead of a new png?

Hmm..  I think so

NickerAban:
but heuchi is doing that stuff for kart icons color

NickerAban:
Well, I'm also showing how is the shape for colored kart icons
I'm even editing to show
hm...
I don't know if you understand

heuchi:
I've removed the extra png. Looks like this now:
KartColorsCircle04

Benau:
btw @heuchi  is there any reason to use static bool for showing_kart_colors? is it very bad to not reset to original state like when playing back battle mode

heuchi:
I did it this way to be able to draw bigger circles around a player's own icon only if needed. Because in case we draw the local player's circle first (or before any other player who has set a custom color), that circle will be drawn too small (not easy to quickly find oneself). With the static flag this can only happen for the first time this method is called (and nobody will ever really notice that). The idea was to keep the original size of the circle in case there are no other circles.
I didn't know, an instance of this class was reused in a different game situation (different players and/or color settings?). If that's the case, then the flag could go somewhere inside RaceGUIBase and be reset when needed.

Or, if you'd be OK with always having the bigger circle around the player's own icon, we could completely abandon _showing_kart_colors_ and just always draw it bigger.
(Some addon-karts use pretty big icons anyway, so this might not be so bad after all.)

Issue #4550, opened 2021-07-14
display rank order numbers on match result screen


lokpro:
## Description
Should display order numbers on the match result screen, helpful for many-kart(like 20) matches.
![match result screen](https://149366088.v2.pressablecdn.com/wp-content/uploads/2019/12/supertuxkart-v1.1.jpg)

Benau:
maybe only do that if > 10 players

lokpro:
I don't know if 10 is easy to count or not :D
As now, we always play 20 karts, I and my children count our ranks every time we finished a race.

Benau:
fixed by kuba pr ![](https://cdn.lihkg.com/assets/faces/pig/free.gif)

maybe ![](https://cdn.lihkg.com/assets/faces/mouse/smile.gif) can ![](https://cdn.lihkg.com/assets/faces/mouse/potion.gif) linux build ![](https://cdn.lihkg.com/assets/faces/mouse/miracle_special.gif)

deveee:
http://31.media.tumblr.com/cdee4e1b4a85c30141ebf368149b04a8/tumblr_n5kaovBJUF1s2wio8o1_500.gif

PR #4551, opened 2021-07-14
Show kart rank in race result screen

mrkubax10:
## Agreement
```

```
Adds feature from #4550

### DIFF
--- a/src/states_screens/race_result_gui.cpp
+++ b/src/states_screens/race_result_gui.cpp
@@ -1325,11 +1325,14 @@ void RaceResultGUI::unload()
             : video::SColor(255, 255, 255, 255);
 
         unsigned int current_x = x;
+        if (RaceManager::get()->getNumPlayers() >= 10)
+        {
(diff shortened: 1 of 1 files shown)


src/states_screens/race_result_gui.cpp
```diff
@@ -1325,11 +1325,14 @@ void RaceResultGUI::unload()
             : video::SColor(255, 255, 255, 255);
 
         unsigned int current_x = x;
+        if (RaceManager::get()->getNumPlayers() >= 10)
+        {
+            int pos_rank_width = m_font->getDimension(core::stringw(n + 1).c_str()).Width;
```
lokpro, line 1330: Hi, I think _pos_rank_width_ should always be 2 digits to keep the names aligned.
```
1  john
2  john
10 john
11 john
```

(I only imagine it from the code. I only know a little bit c++. I programs c#/js...)

mrkubax10: Done

PR #4555, opened 2021-07-18
Add goal to race result

CodingJellyfish:
Hey I'm back!
Simpler and more reliable version of #4213


Benau:
We moved irc to libera.chat by the way 

Freenode is

Benau:
I manually edit the pr to fix the "tabs" by the way

PR #4561, opened 2021-07-21
Fix tracks not showing in grandprix info

mrkubax10:
## Agreement
```

```


Benau:
m_editable was added in https://github.com/supertuxkart/stk-code/pull/1286....

(Avoid the use of the new GP editor as a cheating aid)

not sure auria /  can figure out what / why it was added.....

Benau:
the reason is "preventing people from using the grand prix editor as a way to play tracks that still haven't been unlocked"...

so I think it's a correct behaviour to hide some locked track in grand prix, but for local multiplayer it should show all

PR #4562, opened 2021-07-21
rank by greatest time if points equality in Follow the leader

courbe:
## Agreement
```

```
In Follow the leader mode the greatest cumulated time is a better critera for deciding the rank if several karts have the same total of points. It was set to the less cumulated time as expected for the Grand Prix and Time Trial modes.

### DIFF
--- a/src/race/race_manager.cpp
+++ b/src/race/race_manager.cpp
@@ -785,8 +785,14 @@ namespace computeGPRanksData
         int m_score;
         int m_position;
         float m_race_time;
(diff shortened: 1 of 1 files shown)


Benau:
static variable should be used because it's the same for all data

Benau:
Or call race manager get minor mode == ftl inside sorting, no var is needed at all

courbe:
The call is now done from inside sorting (sorry for having missed the static get function in the hpp file).

PR #4583, opened 2021-08-21
load speedfore.png and speedback.png from skin

risostk:
This makes the speedometer images loadable from the Skin folder. A skin can have its own speedometer.
Not sure if they really need to be defined as _Materials_.

## Agreement
```

```


Benau:
material was historial reason, can you check if force_legacy_device works for your new speedmeter code?

Benau:
(see if transparency is intact)

risostk:
Everything works normally after setting force_legacy_device="true", at least on my computer. Transparency does work.

PR #4584, opened 2021-08-22
Tooltip background for Coal skin

risostk:
Coal skin uses a white font, which has poor contrast on the white background in the common folder. So I made this dark background. You may notice the file size is larger (7.7KB) than the one in the common folder (1.6KB). I tried optipng, but did not help much. Really appreciate if anyone can help.

## Agreement
```

```


Benau:
maybe added some gradient around the border

try gimp contrast change original, and then fill the tooltip rectangle with dark color

risostk:
I did almost as you said. I started from the original image. I didn't change the border, only fill the middle rectangle with a dark color (#373737). I think it looks good, see below:
Screenshot from 2021-08-22 19-10-48
The png is exported by GIMP, Not sure why it is larger than the original. I would expect a similar size.

risostk:
BTW, I tried
`optipng -quiet -o3 tooltip.png`
But it only reduces from 7.7KB to 7.4KB.

qwertychouskie:
`pngquant` is what you are looking for (and maybe run it through `optipng` after if to see if it's even smaller).

risostk:
> `pngquant` is what you are looking for (and maybe run it through `optipng` after if to see if it's even smaller).

Thanks! pngquant gives smaller size than optipng.
`pngquant tooltip.png` 7664KB -> 7010KB
`pngquant --quality=60-80 tooltip.png` 7664KB -> 6512KB --optipng--> 6509KB

Maybe 6.5KB is good enough?

Benau:
https://stackoverflow.com/questions/32355068/how-i-can-replace-colors-with-imagemagick

risostk:
> https://stackoverflow.com/questions/32355068/how-i-can-replace-colors-with-imagemagick
> 
> 

This is hacky! The following command is used to get a file less than 1KB.
`convert tooltip_original.png -fuzz 30% -fill '#373737' -opaque '#dedede' result.png`

Issue #4531, opened 2021-05-13
Inconsistency of icons for stats when choosing a kart


Dinock:
## Description
When you are choosing a kart the icon for nitro efficiency differs in style from the other icons. The icon for the nitro is an image of a rendered 3d-model but the other three icons are drawn (see screenshot).

![STK-icons](https://user-images.githubusercontent.com/54438803/118122298-f4799b80-b3f2-11eb-8d05-8c7ad24ae764.png)

The icon for nitro efficiency should also be drawn to reach consistency in style for all icons.

## Steps to reproduce
Start a new game and have a closer look at the kart selection menu.

## Configuration

STK release version: Version 1.2

System: Ubuntu 21.04

qwertychouskie:
The standard icon set has a lot of these types of inconsistencies, due to how long it's been around with various icons being added/copied as needed throughout the years without regards to whether the icon makes sense in the current location.  For example, this icon came from the Help screen, where a 3d model made slightly more sense there.

(FYI, the Cartoon skin shouldn't have these inconsistencies, as its icon set was created from scratch much more recently, therefore avoiding the whole pieced-together issue.)

Benau:
Yes. I would recommend switching to "cartoon" skin if you would like a better interface, I have no intention to update the original icons anymore

Maybe after a while they will be removed anyway ![](https://cdn.lihkg.com/assets/faces/pig/sowhat.gif)

Dinock:
Thanks for your answers and a suggestion:
Maybe in the next stable version of STK the Cartoon Skin could be set to the default skin. This would make the menus and interface look more consistent and polished right out of the box.

CodingJellyfish:
![image](https://github.com/user-attachments/assets/e00c666c-d99d-4282-b1aa-d50f4da7ec8b)

PR #4617, opened 2021-09-14
RFC: Use Cartoon skin by default?

qwertychouskie:
Many users seem to prefer it as it has a much more consistent icon set (reference https://github.com/supertuxkart/stk-code/issues/4531, https://forum.freegamedev.net/viewtopic.php?f=90&t=18285&p=104294#p104297).

This labels the older skins as "Classic" and sets the default skin to Cartoon.  Additionally, it changes the entry name in user_config so that existing users will have their skin setting reset on upgrade of STK (they can easily change it back if they prefer the Classic skins).

Putting this here to gather wider feedback.

## Agreement
```

```

### DIFF
--- a/data/skins/coal/stkskin.xml
+++ b/data/skins/coal/stkskin.xml
@@ -65,7 +65,7 @@ when the border that intersect at this corner are enabled.
 
 -->
 
-<skin name="Coal" author="Alayan">
+<skin name="Classic - Coal" author="Alayan">
 
 <!--  Stateless -->
 <element type="background" image="background.jpg" />
--- a/data/skins/forest/stkskin.xml
+++ b/data/skins/forest/stkskin.xml
@@ -65,7 +65,7 @@ when the border that intersect at this corner are enabled.
 
 -->
 
-<skin name="Forest" author="Benau">
+<skin name="Classic - Forest" author="Benau">
 
 <!--  Stateless -->
 <element type="background" common="y" image="background.jpg" />
--- a/data/skins/ocean/stkskin.xml
+++ b/data/skins/ocean/stkskin.xml
@@ -64,7 +64,7 @@ when the border that intersect at this corner are enabled.
 
 -->
 
-<skin name="Ocean" author="Dakal & Marianne Gagnon (Auria)">
+<skin name="Classic - Ocean" author="Dakal & Marianne Gagnon (Auria)">
 
 <!--  Stateless -->
 <element type="background" common="y" image="background.jpg" />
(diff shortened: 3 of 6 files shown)


risostk:
How about putting all the older skins to a same class "Classic", and add another spinner next to it to choose different colors? There is enough space for a skin color spinner. In the end, the skin spinner will switch between 'Classic' and 'Cartoon'. If Cartoon is selected, the skin color spinner is hide, and if Classic is selected, the color spinner will appear for selecting a color (Coal, Forest, ...).

Benau:
doesn't like too many buttons in UI...

Benau:
and qwerty https://forum.freegamedev.net/viewtopic.php?f=18&t=18273#p104116 .........................

qwertychouskie:
@Benau I can't seem to find the new image you posted, could you post the direct link?

Benau:
qwerty may just use it after 1.3....

Alayan-stk-2:
I personally don't agree that Cartoon offers a noticeably better experience. The UI/UX has its faults but changing the font ,and the icon set to a different style (that isn't more clear, just stylistically different and a tad more consistent with itself) isn't a solution. The menu background picture of the "classic" themes can easily be changed if it was just that.

A menu UX redesign would have to tackle deeper challenges.

General goals could be (some don't depend on others):
- Laying out all the important options in a way that make them easy to find.
- Minimize the number of clicks to do important actions.
- Take care to not introduce clutter.
- Make selection clear (a small sound when changing selection and small discreet animations for the selected UI item go a long way)
- Have appropriate menu music. MKWii's trick of getting a more dynamic music as the player got closer to the last screen before starting a race was clever. STK's menu music isn't bad but it's not awesome either and it gets tiring. A few different tunes depending on where in the menu the player is could be very welcome.
- Don't have the same background picture everywhere (change depending on the current menu or even on the current selection)
- Some animation in the background, or in "boxes" showcasing gameplay, to make it more lively.

Good UX is timeless and above trends about whether icons should be flat or with depth, square or rounded...

Benau:
maybe riso can show his skin to alayan and we use that one instead...

Benau:
not sure if alayan has strong opinon to merge it anyway (thinks no one uses old skin nowadays)

Alayan-stk-2:
Most people don't change defaults so obviously many people use it. Any default change would be aimed at these people, not at enthusiasts who already tweak the settings to their liking. I'm not fond of changing the default.

Benau:
Then there will be no such thing as 2.0 hype

PR #4624, opened 2021-09-24
Add releases bit to appstream xml.

BryanQuigley:
Noticed this was added to the flatpak here: https://github.com/flathub/net.supertuxkart.SuperTuxKart/blob/master/fix-AppData-file.patch

(Aside: Happy to update the pictures too, they are from 0.9.2 release.. Happy to take screenshots myself or link to others)

## Agreement
```

```


Alayan-stk-2:
I assume this has some use but perhaps you could clarify which.

Is it really a good idea to include the RC here? If the 1.3 RC1 deserve to be there, why not the RCs for 1.2 and others?

Regarding screenshots, newer ones could be appropriate indeed. I notice the current ones are all devoid of any user interface, is that typical? Correct me if I'm wrong, but these are used as previews in e.g. package managers. Showing some race interface in one of those could be useful.

Benau:
please do that changes in data/po/update_desktop_file_appdata.py instead as this xml will be overwritten everytime po is updated

BryanQuigley:
> I assume this has some use but perhaps you could clarify which.
It's the version number used by flatpak - https://flathub.org/apps/details/net.supertuxkart.SuperTuxKart.  Technically you just need to have it incrementing. It's totally possible to just have one release and just increment the number.

There isn't much point in doing more older releases afiau. 

>Regarding screenshots, newer ones could be appropriate indeed. I notice the current ones are all devoid of any user interface, is that typical?
Not at all. I'm happy to make some new ones, but where should I upload them?

> data/po/update_desktop_file_appdata.py
Will redo after I understand that.. Thanks!

BryanQuigley:
Fixed to use the script.  Had to fix a small missing \n though.

PR #4630, opened 2021-09-29
Fix #4625

### LINKED ISSUE #4625: The "early cake limiter" also affects non-racing mode
It doesn't make much sense to replace cakes by bowling ball in modes such as CTF, 3SB, FFA. The code Fouks submitted was only intended for races.

mrkubax10:
## Agreement
```

```
Fixes #4625


Benau:
Kuba it cannot be merged for now because it causes difference in network...

PR #4631, opened 2021-09-30
Remove false compiler warnings in mini_glm.hpp

ahitrin:
In the `compressQuaternion` function, the value of an `extra_2_bit` variable is strictly limited between 0 and 3 by the logic of an algorithm. Unfortunately, the compiler can't infer this limitation from the code and assumes that in some cases it may have another value. This leads to a lot of false warnings: "‘tmp_3’ may be used uninitialized" during the release build (since usage of this variable is not guarded by the `assert(false)` anymore).

In this PR, we replace usage of `std::max_value` and `std::distance` with manual iteration over quarternion coordinates during the search for index of the longest item. This may look ugly, but now the compiler could understand that the compressed vector is always initialized. By removing of the false warning, we improve visibility of other warnings that could point us to potential errors in the code.

Additionally, the change contains few renames and code shuffling. Feel free to point on any commit that looks inappropriate for you. It will be removed/reverted.

## Agreement
```

```

### DIFF
--- a/src/utils/mini_glm.hpp
+++ b/src/utils/mini_glm.hpp
@@ -436,16 +436,16 @@ namespace MiniGLM
                 q.w() / length
             }};
         int extra_2_bit = 0;
-        float max_val = abs(normalized[0]);
-        if (max_val < abs(normalized[1])) {
+        float max_val = std::abs(normalized[0]);
+        if (max_val < std::abs(normalized[1])) {
             extra_2_bit = 1;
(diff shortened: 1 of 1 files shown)


Benau:
actually using ```std::array<float, 3> tmp_3 = {};``` will fix the warning too by the way

Benau:
Fixed in fa1ddbd

ahitrin:
Thank you!

PR #4642, opened 2021-10-06
Use tilde for pre-release versioning in appdata

kalev:
Version 1.3-rc is considered to be higher than 1.3 as per appdata
version comparison logic, which leads to 'appstream-util validate-relax'
complaining:

<release> versions are not in order [1.3 before 1.3-rc1]

Fix this by using tilde instead of a dash to make sure 1.3~rc sorts
lower than 1.3.

## Agreement
```

```

### DIFF
--- a/data/supertuxkart.appdata.xml
+++ b/data/supertuxkart.appdata.xml
@@ -345,7 +345,7 @@
     <release version="1.3" date="2021-09-28">
       <url>https://blog.supertuxkart.net/2021/09/supertuxkart-13-release.html</url>
     </release>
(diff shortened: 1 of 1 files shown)


Benau:
Maybe it should not be mentioned anyway...

It was added for flathub package and noow they remove it too

https://github.com/flathub/net.supertuxkart.SuperTuxKart/blob/master/fix-app-data-versioning.patch

Benau:
Done in https://github.com/supertuxkart/stk-code/commit/1c983f7090cbdd6ad44537c7fad79761bd8df337

kalev:
Nice, thanks!

PR #4666, opened 2021-11-04
Update Angelscript from 2.32.0 to 2.35.1

BryanQuigley:
Didn't require any changes from tarball except dos2unix.

Checked and M1 Mac and DragonFly made it upstream.

Not sure how below agreement pertains to this.. but it's clearly not my code. 

## Agreement
```

```

### DIFF
--- a/lib/angelscript/include/angelscript.h
+++ b/lib/angelscript/include/angelscript.h
@@ -1,6 +1,6 @@
 /*
    AngelCode Scripting Library
-   Copyright (c) 2003-2017 Andreas Jonsson
+   Copyright (c) 2003-2021 Andreas Jonsson
 
    This software is provided 'as-is', without any express or implied
    warranty. In no event will the authors be held liable for any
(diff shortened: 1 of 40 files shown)


Benau:
We made some custom changes in cmakelist to allow building aarch64 in windows.........

And the github actions failling atm, this prevent from being merged...........

BryanQuigley:
Oops, I forgot a new file, that appears to be why it's failing. Will look into  aarch64 in windows

Benau:
```
  Cannot find source file:
172

173
    ../../source/as_callfunc_arm64.cpp
```

Benau:
maybe you can try something like if compiler name == clang define AS_MAX_PORTABILITY for aarch64...

BryanQuigley:
I just tested - and I can move this to just testing in my own fork so I don't keep making noise.

(I also setup a Linux aarch build.. but that works fine :/)

Benau:
if we have  ADD_DEFINITIONS(-DAS_MAX_PORTABILITY)

then

     if(CMAKE_ASM_COMPILER_WORKS)
         set(ANGELSCRIPT_SOURCE ${ANGELSCRIPT_SOURCE} ../../source/as_callfunc_arm64.cpp ../../source/as_callfunc_arm64_gcc.S)
         set_property(SOURCE ../../source/as_callfunc_arm64_gcc.S APPEND PROPERTY COMPILE_FLAGS)
     else()
         message(FATAL ERROR "ARM target requires a working assembler")
     endif()

can be avoided?

Benau:
Let me test in android later..

PR #4678, opened 2021-11-23
Fix GP dialog default option

cheymingyu:
## Agreement
```

```


Benau:
You should use `setFocusCancel` when creating that dialog in gp screen

cheymingyu:
oh i get it... I'll try it again

Issue #3849, opened 2019-04-16
Wrong difficulty displayed in end race screen if comparing ghosts with difficulty


Wuzzy2:
## Description
If you compare ghosts with a different difficulty, the displayed difficulty in the end card (race results) is misleading.

## Steps to reproduce

* Start a replay and compare 2 ghosts that have a different difficulty (e.g. Expert vs SuperTux)
* Watch the replay to the end, until the end card appears
* Look at the end card

EXPECTED:
The end card shows the correct difficulties for all ghosts.

ACTUAL:
The end card only shows one difficulty.

### Simple fix

If the ghost difficulties differ, display something different for the difficulty:

> Difficulties: [lowest difficulty] to [highest difficulty]

(as a range)

### Better fix

In the end card, show the difficulty for each ghost.

## Video

https://www.youtube.com/watch?v=sjjWWtihdac

This is a comparison between Expert and SuperTux, but the end card *only* shows:

> Difficulty: SuperTux

But it should show:

> Difficulties: Expert to SuperTux

# Config

Version: 0.10-rc1
Source: Git

PR #4679, opened 2021-11-25
Add the display of ghosts difficulties in end race screen for replay …

ldoyenard:
…files with standard names (#3849)

It is still a work in progress for the local generated files, but this part works and is an improvement.
Would it be possible to make the game generate replay files with a standard name like -> name_racetype_map_difficulty_[time].replay ?
It would simplify a lot of work on replays.

## Agreement
```

```


Benau:
const ReplayData& getReplayData

Benau:
can you figure it out?

ldoyenard:
This part is figured out, yes

PR #4680, opened 2021-11-25
Fix GP abort dialog default option

cheymingyu:
Hi 
I fixed it the way you suggest. 
But I think it might affect to other confirm dialog.
Is this okay?

## Agreement
```

```

### DIFF
--- a/src/states_screens/dialogs/message_dialog.cpp
+++ b/src/states_screens/dialogs/message_dialog.cpp
@@ -143,6 +143,10 @@ void MessageDialog::loadedFromFile()
         IconButtonWidget* cancel = getWidget<IconButtonWidget>("cancel");
         cancel->setText(_("Don't show again"));
     }
+    else if (m_type == MessageDialog::MESSAGE_DIALOG_CONFIRM)
+    {
+        this->setFocusCancel();
+    }
 }
 
 // ----------------------------------------------------------------------------


Benau:
```
    MessageDialog *dialog =
    new MessageDialog(_("SuperTuxKart may connect to a server "
        "to download add-ons and notify you of updates. "
        "Please read our privacy policy at https://supertuxkart.net/Privacy. "
        "Would you like this feature to be enabled? (To change this setting "
        "at a later time, go to options, select tab "
        "'General', and edit \"Connect to the "
        "Internet\")."),
        MessageDialog::MESSAGE_DIALOG_YESNO,
        new ConfirmServer(), true, true, 0.85f, 0.85f);

    // Changes the default focus to be 'cancel' ('ok' as default is not
    // GDPR compliant, see #3378).
    dialog->setFocusCancel();
    GUIEngine::DialogQueue::get()->pushDialog(dialog, false);
}   // askForInternetPermission
```

cheymingyu:
`else if (action == "right")        // Abort
            {
                MessageDialog* dialog =
                    new MessageDialog(_("Do you really want to abort the Grand Prix?"),
                        MessageDialog::MESSAGE_DIALOG_CONFIRM,
                        new IConfirmDialogListener(), true, true, 0.85f, 0.85f);
                dialog->setFocusCancel();
                GUIEngine::DialogQueue::get()->pushDialog(dialog, false);
            }`

I found that code on main.cpp, and I tried that on race_result_gui.cpp
above code is what i tried.
But it doesn't work well. 
It makes abort error or doesn't display confirm dialog on game.
There is no other examples except that code of main.cpp, so I couldn't figure out how to use 'setFocusCancel()'...
Would you let me know how to use that function?

Benau:
seems that you need to add focus on cancel as a constructor arguement to messagedialog class

Benau:
answser is here https://github.com/supertuxkart/stk-code/commit/b4411b11fe2feed40d92287a69ee315238646d4b

cheymingyu:
sorry, you've done easily... I thought fixing constructor might affects other parts :(
Thanks for fixing it!

Issue #4419, opened 2020-10-26
Add option to not show flags in multiplayer lobby


baryluk:
I wish it was possible for me to set option in my client, to tell other clients to not show my flag derived from geolocation. It would be also nice if server can figure out location, but not send IP or flags to the clients at all, so this can't even be bypassed as such.

I think it is a bit intrusive feature, and I think latency itself is enough in most cases.

PR #4681, opened 2021-11-26
Display_flags

Varro33:
## Agreement
```

```
The player can display flags with a checkbox in options_screen_general. Just translated in french.


Benau:
This option only affect locally... but not (I wish it was possible for me to set option in my client, to tell other clients to not show my flag derived from geolocation) from #4419

And tbh we prefer as little option as possible in stk-code, so sorry, nothing will be done...

Issue #3618, opened 2018-11-24
Minor logic issues with Icy Soccer Field


Wuzzy2:
## Description
The map “Icy Soccer Field” has some minor logical issues:

1. The name does not make sense. This map is not “really” a soccer field, because you play with a puck. Maybe rename the map to something else? I suggest “Frozen Lake”
2. In the minimap, the puck has a soccer ball icon
3. In the victory screen, both teams are represented with a soccer ball icon

## Steps to reproduce
1. Play on Icy Soccer Field
2. Drive forwards and kick the soccer puck … wait, what?

## Configuration
STK release version: eb0000f8e048007c807f16591639ee1af43d9301

dumaosen:
Maybe we can make the soccerball icon customizable in track.xml. It can be useful for those tracks with different ball asset.

PR #4691, opened 2021-12-08
fix : Minor logic issues with Icy Soccer Field (#3618)

rlawogns:
fix 2nd issue of #3618


Benau:
Sorry, the only possibility to identify icy soccer field is via track.xml (see https://github.com/supertuxkart/stk-code/issues/3618#issuecomment-546554491), and i dont think it can be done purely in stk-code 

You can join our telegram channel https://t.me/STKInternational if you need guidance before  a PR next time

rlawogns:
track.xml is included in the .gitignore.
How can I fix the track.xml?

Benau:
this cannot be done in a single PR just using stk-code alone

firetrance3:
Data is collected from a trail in which I am the only one that legitimately
can physically see,  let's just say, I personally,  invented technological
data that every thin flat screen television as well as Android and Apple
products is and was equipped with as a global standard that legitimately
has to be installed on every single double as well as all of those products
mentioned above,  for legitimate security purposes,  that will most
definitely prevent from future terrorism global attacks from occurring, I
did all that at the legitimate age of 4 years old  while carrying a
legitimate marriam Websters Dictionary at my side.  First book I ever
read,  just by picking it up and opening the pages while self teaching my
own self how to not only speak,  but to read as well,  with little help
from my parents,  only because they were to busy hiding around the corner
observing me teach my own self,  while firmly believing that I must have
been a legitimate genius baby either that or I was from outer space like
superman and was switched out with the real baby that my mother had birthed
as the contemplated. However, I have dumbed my very own self down to be at
an even lower level just to avoid daily severe physical abuse from the
students I went to grade school, and half my  middle school term with. My
actual grade point average at the age of 5 was so swell, that it made
legitimate global record history beating Eisenhowers highschool grade point
average, back then it was just Called school,  not split up into multiple
year terms like we have these days.  Sad part is, I get so nervous around
people as well as being watched at all in a general sense that I stress out
way more than anyone could ever imagine.  I sold almost all of my actual
knowledge of anything as well as everything electronically invented after
the year of 1992, to a few people, yes one of them just so happens to be
the actual u.s. government,  and yes for a whole legitimate $20.00 usd.
There is an actual very legitimate legal Contract designed by yours truly,
that states that I can only share so much information as well as speak it
and or in any form, to anyone due to actual safety reasons for the safe
keeping of this planet. So with that being stated,  there is absolutely no
way out of I either,  but with it being said Is the main reason I also had
to dumb myself down,  while mostly it was because of the non stop no reason
other than what I wore clothes wise and shoe wise to school that I got
physically abused by and for from actual school bullies.

PR #4713, opened 2022-01-10
Items animation

Semphriss:
This adds an animation when items respawn.

https://user-images.githubusercontent.com/66701383/148840908-b5c655dc-01f9-46df-893b-fe08b98a13cc.mp4

**Note:** With this animation alone, the only visual cue that happened when the item became pickable again was the resuming of rotation.

I wanted to add more visual cues to make it as intuitive as possible. I originally wanted to make the item much darker while respawning, then reverting back to normal brightness when the item was available again; however, I couldn't get this to work.

Instead, it works by making the item slightly smaller when respawning, then by taking its full size only when the respawn animation finished. I'm a bit unsatisfied with this option, but it seems to work.

If someone could help me with changing items' colors at runtime, please let me know.

## Agreement
```

```

### DIFF
--- a/src/items/item.cpp
+++ b/src/items/item.cpp
@@ -116,11 +116,7 @@ void ItemState::initItem(ItemType type, const Vec3& xyz, const Vec3& normal)
 void ItemState::update(int ticks)
 {
     if (m_deactive_ticks > 0) m_deactive_ticks -= ticks;
-    if (m_ticks_till_return>0)
-    {
-        m_ticks_till_return -= ticks;
-    }   // if collected
-
+    m_ticks_till_return -= ticks;
 }   // update
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


qwertychouskie:
Is it possible to just play the animation when the item is ready for pickup?  Admittedly I've never looked too hard at the code that controls all this, but it seems a bit weird that the item appears, then waits a second, then becomes actually available.

Semphriss:
The problem with removing the respawn time is that it will prevent players from expecting the return of an item. This affects all items, including bananas, and being caught by a banana that you didn't see because it just started respawning is hardly enjoyable :smile: Same goes for other track features like the gift box, if you want to keep your current item and don't want to risk having it replaced, it's better to avoid them.

qwertychouskie:
Maybe show some swilling particles or something like that before having it pop back into existence.  That would give you the warning necessary.

Semphriss:
Using particles don't show with much clarity which item is about to respawn, hence why I wanted to make the sequence using the item's mesh with a visual distinction, e. g. darker or semi-transparent.

Semphriss:
https://user-images.githubusercontent.com/66701383/150666761-8b980642-ed8a-46f1-82cf-97fe947eed10.mp4
https://user-images.githubusercontent.com/66701383/150666762-510aa38e-46b5-4ec5-8435-f4fda6e5ac94.mp4

This new method does not need to save anything to the network, appropriately signals the respawning, and plays the animation at the moment the player can collect the powerup.

I also packed all the textures in a .zip; the images should be placed in the `textures` folder. I also left the original .xcf file, if needed.

[assets.zip](https://github.com/supertuxkart/stk-code/files/7919885/assets.zip)

qwertychouskie:
Looking good!

Semphriss:
[assets.zip](https://github.com/supertuxkart/stk-code/files/7930611/assets.zip)

New assets, 256x256 instead of 512x512

PR #4731, opened 2022-02-11
m_gravity was set on the line above, so the condition will always be false

miselico:
Minor simplification while I am looking into how the cake logic might be improved when playing soccer.

## Agreement
```

```


Benau:
If you change the cake firing logic too siginificantly it may not be approved to avoid to much difference between client and server

PR #4732, opened 2022-02-12
illumos port.

devnexen:
overcoming sun constant defined in this platform.
forcing 64 bits build as angelscript assembly is incorrect for 32 bits.

## Agreement
```

```


Benau:
Does anyone actually use sunos nowadays? I'm not a huge fan of sun -> Sun conversion.....

Also why >> instead of || ?

devnexen:
ah no this is typo sometimes I forget is american keyboard...
otherwise there is illumos solaris itself is kind of dead indeed.

Benau:
maybe you can use `sun_` to make it less ugly

PR #4738, opened 2022-03-06
fix parachute powerup

heuchi:
only deshield karts ahead of the kart firing the parachute

## Agreement
```

```


Benau:
will be merged when close to next release to avoid servers problem....

PR #4743, opened 2022-03-12
widening a bit smart pointers usage considering xml data.

devnexen:
## Agreement
```

```


Benau:
told you these changes are not important.....

And whats the point of using unique_ptr when you need `profile->storeFriends(getXMLData().get());` later......

Benau:
And  thinks that using shared_ptr on request xmlnode is  because they don't need to shared with multiple threads.

Benau:
Maybe you can just keep the 'standalone' changes

PR #4748, opened 2022-03-18
adding the possibility to choose the audio device via a new sub-menu …

devnexen:
…entry.

## Agreement
```

```


devnexen:
no translation for now to see if the idea attracts :)

Benau:
actually staying default seems to be the best idea for all so it can be switched globally by users

devnexen:
Yeah I expected this kind of answer, was thinking in a case with a specialized sound card (eg music) you do not want to be picked on for few cases.

PR #4756, opened 2022-03-27
Fix URLs on Haiku

n0toose:
My general strategy of getting ports to work on Haiku is to just take advantage
of whatever makes things work on FreeBSD by adding an additional condition and
then fixing up things later once it compiles successfully. This generally
works, as Haiku contains FreeBSD headers for compatibility and code that
takes other operating systems into account other than Windows, Linux and OS X
and has a higher degree of POSIX compatibility tends to work on FreeBSD and
Haiku. This strategy is not perfect, and this commit proves it.

Haiku is the successor of BeOS, which, in turn, shows similarities to macOS
every now and then, as the company almost acquired Be, Inc., but later chose to
acquired NeXT, Inc. instead.

`xdg-open` is not available on Haiku the same way it is on FreeBSD and Linux.
With that in mind, this should fix URLs, which do not not work properly on
Haiku.

## Agreement
```

```

### DIFF
--- a/src/online/link_helper.cpp
+++ b/src/online/link_helper.cpp
@@ -88,13 +88,13 @@ namespace Online
         ShellExecuteA(NULL, "open", url.c_str(), NULL, NULL, SW_SHOWNORMAL);
 #elif defined(IOS_STK)
         irr::CIrrDeviceiOS::openURLiOS(url.c_str());
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) || defined(__HAIKU__)
         std::string command = std::string("open ").append(url);
         if (system(command.c_str()))
         {
             Log::error("OpenURL", "Command returned non-zero exit status");
         }
-#elif !defined(__ANDROID__) && (defined(__linux__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__HAIKU__))
+#elif !defined(__ANDROID__) && (defined(__linux__) || defined(__FreeBSD__) || defined(__NetBSD__))
         std::string command = std::string("xdg-open ").append(url);
 
         const char* lib_path = getenv("LD_LIBRARY_PATH");


n0toose:
The port of PPSSPP for Haiku also includes very similar code. I am not able to test this thoroughly right now, paging @devnexen in case he has the time for it.

n0toose:
In retrospect, I realize that I also added an extra parenthesis that made all of this never *not* work to begin with.

devnexen:
well ... xdg-open is available via xdg_utils ? I won t argue this change if you prefer to change but I do not think you re "fixing"

n0toose:
Well, it is fixing it in the sense that there was also a typo on there, but it's true that I just tried compiling this on a virtual machine after fixing the typo and got slightly concerned since I did not have the package for it, even if I was aware of being able to do so. Do you think that it's best to keep code as "native" as possible or as "close to other operating systems so that we don't end up having issues" as possible?

I really should update the commit message with more accurate information.

devnexen:
so I took the time to try out your branch, there is a typo, more precisely a missing closing parenthesis, in src/online/linker_helper.cpp line 97

> Do you think that it's best to keep code as "native" as possible or as "close to other operating systems so that we don't end up having issues" as possible?

If by being "native" you mean having less dependencies to install, well you already need to install openssl, sdl2, cmake, openal ... one package more it s ok I think. but that s not something I m willing to debate, if you prefer to use open (which is even on linux as xdg-open alias) go ahead, that s a tiny detail.

n0toose:
Yeah, fair enough. I am kind of a minimalist in that regard, thought I would also work on that since the thing is that this didn't work in the first place. I'll try to remember to submit a patch to HaikuPorts once the newest release has been shipped.

n0toose:
> so I took the time to try out your branch, there is a typo, more precisely a missing closing parenthesis, in src/online/linker_helper.cpp line 97

*sigh* I thought this was actually a problem. Apologies, will push a commit that fixes it in a sec, thanks for bearing with me.

devnexen:
Do you squash your changes ? I still see the issue

PR #4759, opened 2022-04-03
Upstreaming Haiku changes, disabling some graphics defaults

devnexen:
more inline with the OpenGL capacities.

original author @threedeyes

## Agreement
```

```


Benau:
Does it use only opengl 2.0?

Benau:
I meant if haiku only has opengl 2.0 those settings can't be used by stk anyway. (Ticking advanced lighting in haiku has no effect)

devnexen:
> Does it use only opengl 2.0?

yes :) I guess more to do with having "sane" defaults.

Benau:
in latest commit I deactivate unusable settings in old renderer so it should cause less confusion 

and I don't think options like 'animated characters' or 'particle effects' will have significant impact on performance

PR #4762, opened 2022-04-08
Allow commandline arguments to be passed in on Android

isjerryxiao:
Currently on Android there is no way of using commandline arguments. With this patch at least Android users can do something like `am start --esa argv "--demo-mode=1,--demo-tracks=zengarden\,xr591,--demo-laps=1,--demo-karts=10" org.supertuxkart.stk/.SuperTuxKartActivity`.

## Agreement
```

```


isjerryxiao:
🤔 I guess it's unnecessary then.

PR #4771, opened 2022-04-29
Game exit after a race segfault fix.

devnexen:
Checking if the process kind is set beforehand.

## Agreement
```

```

### DIFF
--- a/src/audio/sfx_manager.cpp
+++ b/src/audio/sfx_manager.cpp
@@ -291,6 +291,7 @@ void SFXManager::queueCommand(SFXCommand *command)
         
     m_sfx_commands.lock();
     if (StateManager::get()->getGameState() != GUIEngine::MENU &&
+        RaceManager::get() &&
         m_sfx_commands.getData().size() > 20*RaceManager::get()->getNumberOfKarts()+20 &&
         RaceManager::get()->getMinorMode() != RaceManager::MINOR_MODE_CUTSCENE)
     {


Benau:
can you try 
```patch
diff --git a/src/main.cpp b/src/main.cpp
index 42343b125..6c210bbf6 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2686,9 +2686,23 @@ static void cleanSuperTuxKart()
         music_manager->stopMusic();
     if (SFXManager::get())
         SFXManager::get()->stopThread();
+
     irr_driver->updateConfigIfRelevant();
     AchievementsManager::destroy();
     Referee::cleanup();
+
+    if (SFXManager::get() &&
+        !SFXManager::get()->waitForReadyToDeleted(2.0f))
+    {
+        Log::info("Thread", "SFXManager not stopping, exiting anyway.");
+    }
+    SFXManager::destroy();
+
+    // Music manager can not be deleted before the SFX thread is stopped
+    // (since SFX commands can contain music information, which are
+    // deleted by the music manager).
+    delete music_manager;
+
     RaceManager::destroy();
     if(grand_prix_manager)      delete grand_prix_manager;
     if(highscore_manager)       delete highscore_manager;
@@ -2746,18 +2760,6 @@ static void cleanSuperTuxKart()
         }
     }
 
-    if (SFXManager::get() &&
-        !SFXManager::get()->waitForReadyToDeleted(2.0f))
-    {
-        Log::info("Thread", "SFXManager not stopping, exiting anyway.");
-    }
-    SFXManager::destroy();
-
-    // Music manager can not be deleted before the SFX thread is stopped
-    // (since SFX commands can contain music information, which are
-    // deleted by the music manager).
-    delete music_manager;
-
     // The add-ons manager might still be called from a currenty running request
     // in the request manager, so it can not be deleted earlier.
 #ifndef SERVER_ONLY
```
and see if this fix your problem?

devnexen:
seems ok.

PR #4774, opened 2022-05-17
Add new things to Appdata

JakobDev:
## Agreement
```

```
I added a few new things to the Appdata file. See [here](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) for the full specification. We can also add a minimum screen size, if there any official.


Benau:
Please make changes in https://github.com/supertuxkart/stk-code/blob/master/data/po/update_desktop_file_appdata.py

Benau:
gift
090d2e80380d8024ba9252bcde8ecaed9fe6523b

PR #4783, opened 2022-07-02
vulkan driver add optional validation layers debug capabilities proposal

devnexen:
## Agreement
```

```


Benau:
With the current stk-code I'm still getting validation message from stdout

devnexen:
yes it s not to fix an issue but to refine the situation (there is room for improvement even).

Benau:
Sorry I don't think it will have any benefit for current sitution... 

Especially I don't even use a cmake option but just make ifdef ENABLE_VALIDATION to if 1 so no recompiling the whole project is needed

PR #4794, opened 2022-09-03
Revert "Fix build when SDL2 is built with DirectFB support"

Newbytee:
The upstream bug that required this will be fixed in a future SDL release, so marking as a draft until that happens.

See https://github.com/libsdl-org/SDL/pull/6084

This reverts commit f5d4475efc1a54400a8e75710317f5de1b1a68b5.

## Agreement
```

```


qwertychouskie:
Linked PR wasn't merged.  This PR should probably be closed for now, one day when STK eventually moves to SDL3, this can be revisited.

PR #4799, opened 2022-09-20
Initialize vulkan handles properly

AMDmi3:
## Agreement
```

```
I'm not really into vulkan, but this code does not build on FreeBSD/i386:
```
/work/usr/ports/games/supertuxkart/work/SuperTuxKart-1.4-rc1-src/lib/graphics_engine/src/ge_vulkan_texture.cpp:440:17: error: cannot initialize a variable of type 'VkImageView' (aka 'unsigned long long') with an rvalue of type 'nullptr_t'
    VkImageView view_ptr = NULL;
                ^          ~~~~
1 error generated.
```
so these types are not pointers and may not be initialized with `NULL`. Judging by vulkan code these may be pointers or ints depending on platform, and `VK_NULL_HANDLE` seems to be the right way to initialize them.


Benau:

AMDmi3:
Sure

PR #4804, opened 2022-10-08
irrlicht custom allocator warning build fix proposal.

devnexen:


Benau:
Now clang warns about unknown pragma...
maybe it can be done without pragma, otherwise just ignore it.....

devnexen:
was wrong only gcc complains originally

Benau:
Now it becomes uglier..... you sure there is no other solution without pragma?

Benau:
Or maybe gcc needs patching

devnexen:
> Or maybe gcc needs patching

let's say today compiler become slightly oversensitive over things :)

anyway it s just a small change this file won t be changed anytime soon (unless you use also the fast allocator but I do not think you would).

qwertychouskie:
IMHO this should be merged, right now building STK is very build-warning-spammy, making it harder to notice legit issues.

Benau:
I would suggest just use clang instead because other than this warning there are so much more false positive warning produced by gcc

PR #4807, opened 2022-10-21
make.sh - specify interpreter directly

licaon-kter:
Run directly it fails with: `./make.sh: 203: Syntax error: "(" unexpected`

A scan on https://www.shellcheck.net/ shows too many issues, this needs to be redesigned a lot to be clean of bashisms

## Agreement
```

```

### DIFF
--- a/android/make.sh
+++ b/android/make.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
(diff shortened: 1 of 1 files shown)


Benau:
if @deveee () think it's ok benau () will merge later...

Benau:
has tested it and git20201030.apk was built from this commit

PR #4812, opened 2022-11-01
Increase default max FPS to suit 144 Hz and 165 Hz displays

Calinou:
The frame limiter's timer lacks precision, which seems to limit it to millisecond timing (rather than microsecond timing). This means we can't limit to 145 FPS (for 144 Hz displays), as this requires a sleep duration of roughly 6.9 milliseconds (and not 7 milliseconds).

We can however set the sleep duration to 6 milliseconds, which is good for both 144 Hz and 165 Hz displays. Both refresh rates are becoming increasingly common in displays used for gaming.

For context, see https://www.phoronix.com/news/SuperTuxKart-1.4 :slightly_smiling_face:
That said, for benchmarking, you'll still want to increase `max_fps` to `1000` in the configuration file. Values above `1000` don't seem to act any different compared to `1000`.

## Agreement
```

```

### DIFF
--- a/src/config/user_config.hpp
+++ b/src/config/user_config.hpp
@@ -657,8 +657,8 @@ namespace UserConfigParams
             PARAM_DEFAULT(  BoolUserConfigParam(false, "show_speedrun_timer",
                             &m_video_group, "Display the speedrun timer") );
     PARAM_PREFIX IntUserConfigParam         m_max_fps
-            PARAM_DEFAULT(  IntUserConfigParam(120, "max_fps",
-                       &m_video_group, "Maximum fps, should be at least 60") );
+            PARAM_DEFAULT(  IntUserConfigParam(167, "max_fps",
+                       &m_video_group, "Maximum frames per second rendered, should be at least 60") );
     PARAM_PREFIX BoolUserConfigParam        m_force_legacy_device
         PARAM_DEFAULT(BoolUserConfigParam(false, "force_legacy_device",
         &m_video_group, "Force OpenGL 2 context, even if OpenGL 3 is available."));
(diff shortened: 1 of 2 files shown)


Benau:
waiting a little while because  is busy fixing archlinux sk

qwertychouskie:
Back when this was implemented, most displays were 60Hz, and a few were maybe 75Hz.  The 120FPS limit was supposed to be something you never hit, it was just a failsafe of sorts.  Nowadays 360Hz monitors are already on the market, with even higher refresh rates in development (and probably released by the time the next version of STK drops).  I'd argue the number should be set to at least 500, maybe even 1000.  Or perhaps it should just be removed altogether.

Calinou:
>  I'd argue the number should be set to at least 500, maybe even 1000. Or perhaps it should just be removed altogether.

On the other hand, on systems with V-Sync disabled, rendering at very high framerates for no good reason can increase power draw significantly and [increase input lag](https://www.youtube.com/watch?v=W66pTe8YM2s) (due to render pipeline stalling). In a world with rising energy costs, we should be careful about this.

Does SuperTuxKart have a way to query monitors' refresh rates? We can automatically cap the framerate to the highest monitor refresh rate on the system (for handling multi-monitor setups with different refresh rates).

Benau:
actually just turn on vsync and the driver will do its task when you swap buffer (no need manually swap)

 thinks then you need a higher value here for 165hz monitor because current with 120fps it cannot reach 'real' 120fps for 120hz monitor without setting it higher (shows mostly 111, probably due to the manual sleep)

Benau:
or maybe make max fps configurable in ui (spinner values), with a 'zero' value indicates to follow monitor refresh rate

there is already max fps setting for ios

not sure calinou wanna try himself or  can do that later

PR #4845, opened 2023-01-26
Add missing includes to fix the build with gcc 13

heirecka:
Like other versions before, gcc 13 moved some includes around and as a result <stdexcept> and <cstdio> are no longer transitively included. Explicitly include them for std::runtime_error and snprintf.

## Agreement
```

```


Benau:

heirecka:
Yes, generally I'm ok with that, although it may depend on specifics. But BSD or MIT for example would be fine for me, too.

Benau:
Yeah probably they will use mit or zlib library license (same as irrlicht)

PR #4849, opened 2023-02-23
gcc13 fixes

limburgher:
## Agreement
```

```


Benau:

limburgher:
I am.

Issue #4853, opened 2023-03-20
Link error when compiling with Vulkan support


qwertychouskie:
When compiling with Vulkan support, I get this error:
```
[100%] Linking CXX executable bin/supertuxkart
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::setNanMinMaxClamp(bool)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::parse(TBuiltInResource const*, int, EProfile, bool, bool, EShMessages, glslang::TShader::Includer&)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TPoolAllocator::allocate(unsigned long)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TPoolAllocator::push()'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::setEntryPoint(char const*)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::preprocess(TBuiltInResource const*, int, EProfile, bool, bool, EShMessages, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*, glslang::TShader::Includer&)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::setShiftUboBinding(unsigned int)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TShader::setShiftSsboBinding(unsigned int)'
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/libshaderc.so: undefined reference to `glslang::TIntermediate::findLinkerObjects() const'
[...]

PR #4854, opened 2023-03-21
Revert "Remove bundled shaderc"

qwertychouskie:
Re-include `shaderc` for now, as leaving it out causes linking failures, segfaults, and endless headaches: https://github.com/supertuxkart/stk-code/issues/4853

Current stable/LTS versions of Ubuntu/Debian don't include `shaderc` in the repos, and the version included in Debian Sid/Ubuntu 23.04 Beta causes immediate segfaults when using the Vulkan renderer even after manually fixing linking errors.

Closes https://github.com/supertuxkart/stk-code/issues/4853


Benau:
no sorry i have no intention to merge this because vulkan is always experimental, as long as it works for myself i will bother it later when it becomes final

Benau:
and debian when package stk it will always use latest shaderc anyway so the broken shaderc needs to be dealt with (by official shaderc later)

PR #4865, opened 2023-05-15
Fix vk_mvk_moltenvk.h on macOS 10.12 and earlier

evanmiller:
IOSurfaceRef.h does not exist on macOS 10.12 and earlier. However the `IOSurfaceRef` type will be indirectly included by the Metal include, so we just need to ensure IOSurfaceRef is defined when `__OBJC__` is not. This can be done with a typedef.

See: https://trac.macports.org/ticket/67263

## Agreement
```

```

### DIFF
--- a/lib/graphics_engine/include/vk_mvk_moltenvk.h
+++ b/lib/graphics_engine/include/vk_mvk_moltenvk.h
@@ -26,13 +26,12 @@
 extern "C" {
 #endif	//  __cplusplus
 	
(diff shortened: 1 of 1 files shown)


evanmiller:
Actually, I'm told this patch doesn't work correctly on 10.9 so it may need some tweaking.

Benau:
It should be done in https://github.com/KhronosGroup/MoltenVK anyway

Benau:
With latest moltenvk it's no longer required

PR #4918, opened 2023-10-28
Improved soccer timing

Athozus:
Hi,

I like soccer in STK but I noticed few things. Here are my changes, 3 commits that you can merge or not :

1. In countdown mode, wait for the last goal scored to finish the game. It adds `m_goal_after_count_down_ending` boolean.
2. (Useful for previous commit), it shows the 00:00.000 time in red instead of orange as it was previously.
3. In both modes, but more useful for countdown, it pauses the timing when a goal is scored, and unpauses it when game is relaunched. It needs changes in `WorldStatus` so compilation might take a bit more time, but 2 or 3 minutes at all. It adds `m_paused_clock_time`, `pauseClockTime()` and `unpauseClockTime()`.

Hope you'll like it.

## Agreement
```

```

### DIFF
--- a/src/modes/soccer_world.cpp
+++ b/src/modes/soccer_world.cpp
@@ -333,6 +333,7 @@ void SoccerWorld::reset(bool restart)
     }
 
     m_count_down_reached_zero = false;
+    m_goal_after_count_down_ending = false;
     m_red_scorers.clear();
     m_blue_scorers.clear();
     m_ball_hitter = -1;
@@ -563,6 +564,9 @@ void SoccerWorld::onCheckGoalTriggered(bool first_goal)
                 sd.m_time = getTime();
             m_blue_scorers.push_back(sd);
         }
+        if (RaceManager::get()->hasTimeTarget() and m_count_down_reached_zero) {
+            m_goal_after_count_down_ending = true;
+        }
         if (NetworkConfig::get()->isNetworking() &&
             NetworkConfig::get()->isServer())
         {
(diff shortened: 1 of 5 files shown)


Benau:
Save the score first and let the game end naturally is better.

Athozus:
> Save the score first and let the game end naturally is better.

You mean for removing variable `m_goal_after_count_down_ending` ?

Benau:
Actually we already save the score first, so if you only want the goal animation to finish before showing result screen, we can't pause the timer to achieve that because of network synchronization.

Athozus:
> Actually we already save the score first, so if you only want the goal animation to finish before showing result screen, we can't pause the timer to achieve that because of network synchronization.

Nope, it doesn't do that. It just let the players finish the move : when countdown reaches 00:00, the game finishes when the next and last goal is scored.

Benau:
as long as the ball crossed the goal line before count down reaching zero, the last score is saved and displayed in the result screen, that's enough, and we don't care about 'let player finish the move' or whatever

Athozus:
No, but, it's just that it's Countdown + game finish (like in irl, even I don't watch soccer a lot). Anyway, if you don't like/want it, just don't merge it.

Alayan-stk-2:
> 1. In countdown mode, wait for the last goal scored to finish the game. It adds `m_goal_after_count_down_ending` boolean.

The point of countdown mode is that it's finished at 0:00. I don't like this idea.

> 3\. In both modes, but more useful for countdown, it pauses the timing when a goal is scored, and unpauses it when game is relaunched. It needs changes in `WorldStatus` so compilation might take a bit more time, but 2 or 3 minutes at all. It adds `m_paused_clock_time`, `pauseClockTime()` and `unpauseClockTime()`.

I like this idea however.

Benau:
As long as there's network involved in a world, time can never be paused, so sorry nothing can be done about that for now

PR #4968, opened 2023-12-13
Fix #4947

### LINKED ISSUE #4947: The typing bars are very confusing with the coal theme
Using the coal theme, the typing bars that were added in some screens such as the track selection screen, are very confusing:

- Unlike for other themes, they don't have proper borders, no matter if focused or not
- The blinking | you get when selecting them is black-on-dark-grey background so almost invisible if you don't know to look for it.
- A sample text saying it's a search bar that's shown inside in a color contrasting less with the background than the default text color could also be useful for all themes. As the user would start to type in, it would disappear.

CodingJellyfish:
## Agreement
```

```

### DIFF
--- a/data/skins/coal/stkskin.xml
+++ b/data/skins/coal/stkskin.xml
@@ -338,9 +338,9 @@ to specify all types of ttf.
 <color type="tips_background" state="neutral" a="120" r="0" g="0" b="0" />
 
 <!-- Text field color -->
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Thank you for this PR. I tested the change, and it fixes the background and the border having the same color, but I'm not quite satisfied with the new result.

I am going for a different set of colors, and making the cursor color depend on the skin-theme.

PR #4973, opened 2023-12-19
Fix #3471

### LINKED ISSUE #3471: Font scaling issues
The rank indicator font is currently upscaled using setScale, and its blurriness is painfully obvious.

Regular text fonts also become blurry when modified with setScale, but afaik this is not really needed anywhere.

CodingJellyfish:
## Agreement
```

```
IDK why don't fix it like this. It's called high res digit font anyway.

### DIFF
--- a/src/font/digit_face.hpp
+++ b/src/font/digit_face.hpp
@@ -35,7 +35,7 @@ private:
     // ------------------------------------------------------------------------
     virtual float getScalingFactorOne() const OVERRIDE         { return 0.7f; }
     // ------------------------------------------------------------------------
-    virtual unsigned int getScalingFactorTwo() const OVERRIDE    { return 40; }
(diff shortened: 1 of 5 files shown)


CodingJellyfish:
Before:
屏幕截图 2023-12-19 195316
After:
屏幕截图 2023-12-19 200715

Benau:
Instead of setting 0.4 everywhere you should make a getScalingFactorThree or something like that so that the glyph page itself is scaled up internally. Just like the 'native scale' in iOS

Benau:
This probably too difficult for you so let me do it later

Benau:
something like https://github.com/Benau/stk-code/commit/1e80122c6176b15897175b464839cf0cf501044e but not quite working for the speedometer and the score in overworld, figure out the rest yourself

PR #4983, opened 2024-01-04
Fix #3894

### LINKED ISSUE #3894: Zipper effect is not visible to outsiders
## Description
If you play in an online race and the other player uses a zipper, the effects for this are missing.

There is no graphical effect, nor a sound effect. You can only see it indirectly by the sudden speed boost and (if spectating) a zipper icon disappearing.

## Steps to reproduce
- Spectate a race online
- Wait until someone uses a zipper

## Configuration
STK release version: 1.0.0

Nomagno:
## Agreement
```

```
Make the zipper sound positional. Make network racers zipper sound be played.
This probably requires the corresponding stk-assets sfx.xml edit to be pushed
to work properly.


Benau:
Logically it shouldn't be played for remote player because it's not the sound coming from the engine but from the "GUI / game" itself

PR #4987, opened 2024-01-07
Update bowling.cpp

Snoker101:
When firing backwards, multiply bowling speed by 4 if kart is moving backwards so the player doesn't catch up with the ball, and multiply by 2 otherwise  (useful in soccer).

## Agreement
```

```


Alayan-stk-2:
No functional change before 1.5 anyway

PR #4988, opened 2024-01-07
Update cake.cpp

Snoker101:
When firing backwards, cake speed must be high enough to hit the target if the kart moving backwards.

## Agreement
```

```

### DIFF
--- a/src/items/cake.cpp
+++ b/src/items/cake.cpp
@@ -99,7 +99,9 @@ void Cake::onFireFlyable()
     m_speed *= m_owner->getSpeed() / 23.0f;
 
     //when going backwards, decrease speed of cake by less
-    if (m_owner->getSpeed() < 0) m_speed /= 3.6f;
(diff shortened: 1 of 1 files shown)


Benau:
There is if (m_speed < 1.0f) m_speed = 1.0f; anyway

Benau:
Difficult to convince me to merge your PR just because you changed some number and says it works better unless you have someone else to vouch for you

Same for your bowling PR

Snoker101:
Here, check the difference yourself:
https://youtu.be/ACGmu2lD1hc

Benau:
According to your rival 

```Backwards-going karts (like slowly forward-moving karts) can't really use cakes or bowling balls. I never considered that a problem. I think it's intended. You need some speed to use the items.```

Alayan-stk-2:
No functional change before 1.5 anyway

PR #5038, opened 2024-03-29
Super mega LOD fix

CodingJellyfish:
## Agreement
```

```
Quick cull invisible LOD nodes & reduce dynamic_cast

### DIFF
--- a/src/graphics/draw_calls.cpp
+++ b/src/graphics/draw_calls.cpp
@@ -145,40 +145,32 @@ void DrawCalls::parseSceneManager(core::array<scene::ISceneNode*> &List,
 {
     for (unsigned i = 0; i < List.size(); ++i)
     {
+        if (!List[i]->isVisible())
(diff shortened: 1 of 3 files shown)


Alayan-stk-2:
Testing shows (for the non-Vulkan part):
* The patch is a net performance gain in select tracks and settings, where time spent in draw call generation is reduced.
* Unfortunately, the gains are not uniform. There is an increased variability in frame-time, with a bigger improvement in already fast frames than in slower frames, so the real smoothness gains are lower than average FPS would indicate.
* In most tracks, there is no appreciable difference with or without the patch.
* The patch helps, as expected, in CPU-limited situations, not in GPU-limited situations.
* Nonetheless, in all situations it appears as fast or faster than the master branch.

The following data is taken from tests made in Restful River Ring:

RRR_frames
FrameTime2_nopatch
FrameTime2

PR #5047, opened 2024-04-08
improve MinGW build

ognevny:
based on patch from https://github.com/msys2/MINGW-packages/pull/20587 but without breaking changes

## Agreement
```

```


CodingJellyfish:
Sorry but it didn't pass the check.

ognevny:
> Sorry but it didn't pass the check.

I see

Issue #5035, opened 2024-03-19
The bundled bullet code fails to compile with strict-aliasing violations


eli-schwartz:
Relevant: #5

Essentially https://github.com/bulletphysics/bullet3/issues/4590

tl;dr

Gentoo is trying to get all distro packages to be LTO-clean, and as part of that we are trying to compile with `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing` to check for cases where the compiler can try to optimize based on assuming UB cannot happen, and miscompile code that has UB in it. strict-aliasing issues are always bad but LTO can make them even worse.

We already track bullet as an issue -- and we have a workaround, in that we can tell the compiler "well just don't optimize this code, it's too risky". The problem here is that since bullet is part of the supertuxcart source build, we can't just build supertuxcart with optimizations, but link to an unoptimized bullet (until bullet is fixed).

You should probably track the bullet ticket, but it might also be good to somehow consider unbundling bullet if possible.

I have no idea if you still need a patched copy -- it would be great if that need is obsolete. The other reason mentioned in #5 was about using the correct version of bullet:
> Updating bullet to a newer version is a major undertaking, since (in the past) the physics behaviour can change drastically from one version to the next (which is why bullet recommends not to use a system library).

I think it would be quite reasonable honestly to just... require an exact version (or range of versions) of bullet that are *known to work*, and allow people to configure against a system copy of that and otherwise force them to use the bundled copy.

I think "it changes drastically from version to version" is overall a weak argument against the use of system libraries.

...

Reported downstream at https://bugs.gentoo.org/458894

qwertychouskie:
Un-bundling Bullet is not going to happen.  Even the smallest differences between versions can create massive de-synchronizations in networked multiplayer.

Although upgrading Bullet is *technically* possible between 1.5 (the last planned release in the 1.x series) and 2.0, it is extraordinarily unlikely to happen, given how much it could affect the feel of the game, and how much validation (playtesting, network testing across a variety of platforms and architectures, etc) would be required.  Such a major undertaking would only happen if a newer Bullet provided some feature or improvement considered important to how the game plays, not just to enable a small compile optimization.

Patches to make the bundled Bullet LTO-clean could be merged if provided, but **only** if they can be proven to not affect physics in any way.

Alayan-stk-2:
[...]

PR #5048, opened 2024-04-08
add support for system Bullet

ognevny:
## Agreement
```

```


qwertychouskie:
Using system Bullet is not an option, see https://github.com/supertuxkart/stk-code/issues/5035#issuecomment-2005581607

eli-schwartz:
This pull request doesn't contain any version matching to ensure that the correct version of bullet is used. In theory the objections from the linked ticket could could be answered via requiring specific versions or else mandating the bundled copy

While I would absolutely appreciate support for a system copy too, I think it's extremely important to collaborate with upstream's concerns that "just any version" will produce incorrect results.

qwertychouskie:
> This pull request doesn't contain any version matching to ensure that the correct version of bullet is used. In theory the objections from the linked ticket could could be answered via requiring specific versions or else mandating the bundled copy
> 
> While I would absolutely appreciate support for a system copy too, I think it's extremely important to collaborate with upstream's concerns that "just any version" will produce incorrect results.

Even "the exact same version" wouldn't work, as STK's Bullet has had modifications made over the years, including changes to functionality.

https://github.com/supertuxkart/stk-code/commits/master/lib/bullet

PR #5055, opened 2024-04-17
Fix #5054

### LINKED ISSUE #5054: Culling of invisible elements and LoD bugged in splitscreen
The recent LoD merge is almost assuredly the culprit. I didn't find any display error without splitscreen, but testing now the error is unmissable.

As soon as the top screen section kart gets too far away from the bottom screen section kart, it turns invisible. Driving with the other kart to get closer and it turns back visible.

LoD is also bugged in the top screen. It seems to be displaying high quality with only the bugged kart nearby, low quality when the non-bugged kart comes closer and high quality when it gets close.

CodingJellyfish:
## Agreement
```

```


Alayan-stk-2:
Tested, it solves the issue. Thanks!

PR #5071, opened 2024-05-05
Fix #5067

### LINKED ISSUE #5067: Noticable seperation between different shadow passes in cascade shadow mapping
Would be fixed very soon
image
屏幕截图 2024-05-04 010214

CodingJellyfish:
## Agreement
```

```
Sunlight pass takes 15% more time, which is trivial.
Shadows gaussian blur isn't needed anymore, which means blurshadowsV.comp and blurshadowsH.comp should be deleted.


Alayan-stk-2:
In my local testing, the transition between the good and worse looking shadow is now less obvious.

CodingJellyfish:
Tested on Android, and new PCF algorithm works well on GLES.

PR #5100, opened 2024-05-25
Avoid giving bomb to team mate

Athozus:
This pull request avoid a thing I noticed in football mode, that you can give your bomb to your teammate. <ins>But maybe this is intended, so not sure</ins> (however I did not find anything relevant).

I just copy-pasted the code from the swatter.

## Agreement
```

```

### DIFF
--- a/src/items/attachment.cpp
+++ b/src/items/attachment.cpp
@@ -400,7 +400,8 @@ void Attachment::hitBanana(ItemState *item_state)
  */
 void Attachment::handleCollisionWithKart(AbstractKart *other)
 {
-    Attachment *attachment_other=other->getAttachment();
+    Attachment *attachment_other = other->getAttachment();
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Hello, thank you for your PR.

For gameplay changes, it is generally best do discuss them on the forum or the IRC before submitting a PR.

Because of compatibility, this won't be merged for 1.5, but I'm unsure if it should  be merged at all. Passing the bomb can be bad, if it makes the teammate's kart explode. But in theory it could also be useful, if your teammate has a gum or a swatter that could get rid of the bomb after it's passed. And does it make much difference for the team, whether your kart explodes or your teammate's kart explode? Colliding with your teammate is still something you are in control of.

Athozus:
Closing it until I'm not the only who noticed that. Plus the reasons are totally significant.

PR #5119, opened 2024-06-26
Two New skin 

willliam9:
## Agreement
```

```


willliam9:
This modification is not intentional 
[WTF - Fix of stk_host.cpp](https://github.com/supertuxkart/stk-code/pull/5119/commits/f66be00b98ed68a738c0ee0ee49690650228cc1c)

CodingJellyfish:
These new skins are good, but the background of mars doesn't come from official tracks, so further modifications are needed.

Alayan-stk-2:
I'm not saying the skins are bad, but I don't think they bring much to the existing set. 5 color variants (+ a dark variant) per base skin is quite enough in my opinion, getting redder red instead of pink-red or getting another shade of green isn't bringing much.

In the future, it would be nice to have the possibility to install skins from the addons, but we have to put some limits on what gets shipped with the game. [More choice is not always better.](https://thedecisionlab.com/biases/choice-overload-bias)

PR #5124, opened 2024-07-03
Separate database-related code from ServerLobby

kimden:
Here, I split the database-related code previously located in ServerLobby implementation into logic part, which stayed in ServerLobby, and database part (forming queries, all types of sqlite3_* functions), which moved to a new DatabaseConnector class. Apart from that, I reused some code previously repeated while making queries of same type and while binding values (by adding two auxiliary structures), and added some doxygen comments.

I felt like making a separate class for that purpose would do better than not making it, as there a few more variables to keep track of than just database handle, and passing database handle to each of many functions would probably be too long in terms of code.

The amount of changes is pretty big, so I suppose that despite I did some testing, it requires even more of it + I could miss something, feel free to ask me anything about it.

## Agreement
```

```

### DIFF
--- a/src/network/database_connector.cpp
+++ b/src/network/database_connector.cpp
@@ -1,6 +1,6 @@
 //
 //  SuperTuxKart - a fun racing game with go-kart
-//  Copyright (C) 2013-2015 SuperTuxKart-Team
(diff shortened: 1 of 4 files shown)


CodingJellyfish:
I trust you.

PR #5104, opened 2024-05-28
Update AppData file

AsciiWolf:
Add launchable tag that is now required by many AppStream parsers.

## Agreement
```
By creating a pull request in stk-code, you hereby agree to dual-license your contribution as
GNU General Public License version 3 or any later version and
Mozilla Public License version 2 or any later version.

This includes your previous contribution(s) under the same name of contributor.

Keep the above statement in the pull request comment for agreement.

```

/cc @Benau

Also consider renaming the AppData file (and the app id) to a rDNS (net.supertuxkart.SuperTuxKart) format. The non-rDNS one is deprecated.

PR #5137, opened 2024-08-04
Rename AppStream metadata file to a rDNS format

AsciiWolf:
Also:
- Set the component type to desktop-application instead of a deprecated desktop one.
- Update data/po/update_desktop_file_appdata.py + add the previously added launchable type there.

## Agreement
```

```

Follow-up to the [previous PR](https://github.com/supertuxkart/stk-code/pull/5104).

/cc @CodingJellyfish

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -910,7 +910,7 @@ install(FILES data/supertuxkart_128.png DESTINATION share/icons/hicolor/128x128/
 install(FILES data/supertuxkart_256.png DESTINATION share/icons/hicolor/256x256/apps RENAME supertuxkart.png)
 install(FILES data/supertuxkart_512.png DESTINATION share/icons/hicolor/512x512/apps RENAME supertuxkart.png)
 install(FILES data/supertuxkart_1024.png DESTINATION share/icons/hicolor/1024x1024/apps RENAME supertuxkart.png)
-install(FILES data/supertuxkart.appdata.xml DESTINATION share/metainfo)
+install(FILES data/net.supertuxkart.SuperTuxKart.metainfo.xml DESTINATION share/metainfo)
 
 if(MINGW)
     install(DIRECTORY ${CMAKE_BINARY_DIR}/bin/ DESTINATION ${STK_INSTALL_BINARY_DIR}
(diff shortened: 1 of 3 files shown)


AsciiWolf:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

> The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.gwenview or com.hugski.colorhug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.
> 
> To increase the uniqueness and to distinguish between different pieces of a software suite, it is suggested to append the type name to the component-id in these cases.

(Do not worry about the previous sentence about lowercase letters. Even GNOME uses a mix of lowercase and uppercase for app IDs and there are no problems with it.)

AsciiWolf:
Any news?

qwertychouskie:
Looks good to me, @CodingJellyfish feel free to merge.

AsciiWolf:
@CodingJellyfish Ping. :-)

AsciiWolf:
@CodingJellyfish Ping. If there is any problem with this PR, please let me know. Thanks!

By the way, don't forget to re-generate [the pot file](https://github.com/supertuxkart/stk-code/blob/master/data/po/supertuxkart.pot) / po files after this PR gets merged.

deveee:
Don't you think that pinging so many people so often isn't abusive behavior?

AsciiWolf:
> Don't you think that pinging so many people so often isn't abusive behavior?

Maybe a little. On the other hand, do you think that completely ignoring this PR for almost two months (and even ignoring the review from another contributor) is a proper behavior?

If you want to keep ignoring this PR, please at least fix the bug in your update_desktop_file_appdata.py script (fixed as part of this PR) that is currently missing the [launchable tag](https://github.com/supertuxkart/stk-code/pull/5137/files#diff-db0d766d0261f41d1a70a1c8f30fc349eeb6f8d908cd9a3e939626da07333d67R73) and **will damage the AppStream metadata file** the next time it is launched.

qwertychouskie:
> Don't you think that pinging so many people so often isn't abusive behavior?

Pinging once when opening the PR and once after 1.5 months is hardly what I would consider excessive.  (And the one other ping was me.)

PR #5140, opened 2024-08-08
Replace std::random_shuffle with std::shuffle

nyllet:
random_shuffle was removed in c++17

## Agreement
```

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -500,7 +500,7 @@ endif()
 
 # Set some compiler options
 if(UNIX OR MINGW)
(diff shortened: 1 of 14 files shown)


Alayan-stk-2:
Thanks for the PR.

As-is, however, it fails to pass some our CI checks.

nyllet:
hi
> As-is, however, it fails to pass some our CI checks.
yes, i think it's because i changed from gnu++0x to c++11. i'll revert that and hopefully it builds on mingw as well

kimden:
If you continue working on that, please test afterwards whether the online gameplay is unaffected — there was a case before when something was accidentally changed in random generators, and 'updated' servers/clients had different items from the 'old' ones. (If I'm not mistaken, something was postponed to the next major release even because of that.)

PR #5143, opened 2024-08-08
Use CURLOPT_XFERINFOFUNCTION instead of CURLOPT_PROGRESSFUNCTION

nyllet:
## Agreement
```

```

### DIFF
--- a/src/online/http_request.hpp
+++ b/src/online/http_request.hpp
@@ -31,11 +31,13 @@
 #include <assert.h>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


nyllet:
the old check:

#if defined(CURLOPT_XFERINFODATA)

was not working since CURLOPT_XFERINFODATA defaults to NULL

CodingJellyfish:
defined(CURLOPT_XFERINFODATA) is 1 since CURLOPT_XFERINFODATA is defined

PR #5145, opened 2024-08-09
Replace std::random_shuffle with std::shuffle

nyllet:
std::random_shuffle was removed in c++17

## Agreement
```

```


CodingJellyfish:
Should it use mt19937? From what I know it has no benefit other than a much longer cycle.

PR #5146, opened 2024-08-09
Upgrade c++ standard from gnu++0x to c++17

nyllet:
Make sure it builds for mingw. The WIN32 macro appears not to be set when building for mingw. However, __MINGW32__ is set for both 32-bit and 64-bit

## Agreement
```

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -500,7 +500,7 @@ endif()
 
 # Set some compiler options
 if(UNIX OR MINGW)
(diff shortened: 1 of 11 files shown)


nyllet:
#5145 must be merged before this one, otherwise the macos builds will fail

deveee:
I'm not keen to make such changes because I don't see real benefit and it usually breaks compilation. For example I build linux package using old debian for better compatibility.

CodingJellyfish:
Is there any progress on this? If not I guess I should close this.

PR #5190, opened 2024-10-21
Add 144 FPS cap option

LukeWarnut:
Adds 144 as an option for 'Maximum FPS' in the graphics settings. This is useful for people that have 144 Hz monitors that do not wish to use vsync.

## Agreement
```

```


CodingJellyfish:
I think it's better to detect the physical screen refresh rate instead of making an additional option.

PR #5217, opened 2024-12-01
Add Arch install instructions and fix typos (among other things) in INSTALL.md

FifthTundraG:
## Agreement
```

```

Arch instructions were tested in a VM with the only packages installed being `base`, `linux`, `linux-firmware`, `sudo`, `networkmanager`, and `lxqt`. No errors during compile and it ran as expected.

I also reordered the package manager commands listed to be more reflective of the number of users that distro has (there are undoubtedly more users of Debian-based distros then "Mageia 6", which I had never even heard of until reading this list 🙃). If there's any strong objections towards this change then feel free to order it back to the way it was.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,6 +26,15 @@ To build SuperTuxKart from source, you'll need to install the following packages
 * jpeg (libjpeg-turbo-devel)
 * SDL2 (libsdl2-devel)
 
+Debian-based Distributions command:
+
+```bash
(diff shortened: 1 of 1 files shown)


searinminecraft:
This looks good to me. I was frustrated that there were no arch build instructions which made me figure out the packages needed myself. Now if this gets merged, i wont have to worry about it again :3

CodingJellyfish:
Thank you for your contribution.

PR #5235, opened 2024-12-29
Fix #5234

### LINKED ISSUE #5234: Client GP result screen shows point additions using local config instead of server
## Description
An online GP server doesn't have to conform to scoring system used in local GPs. It can send any points.

The client has to use those points received from the server in three cases: (1) in the race results screen, showing +delta; (2) in the animation of current GP results screen, showing +delta, then slowly converting it into points on the screen; (3) after the animation, just showing the current GP standings.

Currently, the client uses points from the server in (1) and (3), and uses local GP scoring in (2), which leads to confusion, 

Illustrations:
(1)
Screenshot from 2024-12-29 16-24-12

(2)
Screenshot from 2024-12-29 16-24-15

(3)
Screenshot from 2024-12-29 16-24-19

## Steps to reproduce

1. Start a GP with any server that has a GP scoring different from the default one (e.g. by modifying stk_config.xml)
2. Finish a game (preferably with more than one player, otherwise there's no animation, but the issue is visible there too as the server points for a single player can be nonzero)
3. Observe the result screens.

## Configuration

STK release version: latest git 4ea1a28ca26d34ae6a

kimden:
As always, a bit more testing is always appreciated if possible

## Agreement
```

```

### DIFF
--- a/src/states_screens/race_result_gui.cpp
+++ b/src/states_screens/race_result_gui.cpp
@@ -1286,20 +1286,15 @@ void RaceResultGUI::renderGlobal(float dt)
             {
                 WorldWithRank *wwr = dynamic_cast<WorldWithRank*>(World::getWorld());
                 assert(wwr);
-                int most_points;
-                if (RaceManager::get()->isFollowMode())
-                    most_points = wwr->getScoreForPosition(2);
-                else
-                    most_points = wwr->getScoreForPosition(1);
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
Tested, and it resolves the bug online. I didn't notice any regression checking in local play.

Issue #5109, opened 2024-06-10
[GUI] Consider removing blurred elipse from 5 main menu entries


pedropaulosuzuki:
### Description

In the main menu, the selected item already changes color, so it is a bit redundant to have the ellipse down there as well. And it makes the UI look a bit more dated.

Refering to the blurred white ellipse under "Single Player" (the selected box) on the screenshot below:

![Screenshot of Super Tux Kart 1.4 main menu](https://github.com/supertuxkart/stk-code/assets/32345702/607be8ac-dbb0-4c7a-afb3-85c1041eb15d)

Maybe it could be changed to look less dated, maybe like a glow around the box or something, but removing is fine, in theory.

Not talking about removing the glow in other areas other than the main menu, or even the ones in the bottom right corner, since they do not have the same behavior of changing colors when selected. Only talking about the 5 big icons in the center.

### Configuration
STK release version:  1.4 Flatpak

CodingJellyfish:
Indeed, the icon buttons having a dedicated focus image should have a different logic from those don't.

PR #5245, opened 2025-01-12
Removed glow under main menu options

MollyMaclachlan:
## Agreement
```

```

This follows the proposal in #5109 to remove the glow under the five main menu options, and derives its approach from the recommendation given when #5222 was closed.

To see if an icon button has a focus image specified, I check if the size of `m_properties[PROP_FOCUS_ICON]` exceeds 0, as is done when assinging `m_highlight_texture` in `icon_button_widget.cpp`.

### DIFF
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -1424,10 +1424,11 @@ void Skin::drawRibbonChild(const core::recti &rect, Widget* widget,
         }
 
         const bool mark_focused =
-            focused || (parent_focused && parentRibbonWidget != NULL &&
+            (focused || (parent_focused && parentRibbonWidget != NULL &&
                           parentRibbonWidget->m_mouse_focus == widget) ||
                        (mark_selected && !always_show_selection &&
-                          parent_focused);
+                          parent_focused)) &&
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
From a glance at your code, you seem to set use_square to true by default, and then rely on the if/else if structure to ensure that the code meant for use_square=true is not used by things that should use the glow and not the square highlight.

This may work as is, but the name of the variable and the content of the variable not matching when the variable is used is susceptible to cause trouble down the line if surrounding code is changed.

It's cleaner to ensure the variables get values that match their meaning when setting them up at the beginning, so that they can be used later without the danger of running into gotchas.

MollyMaclachlan:
Thanks for the feedback. I've updated the approach to use the `widget->m_properties[PROP_FOCUS_ICON].size()` check to avoid having `mark_focused` set to true at all for widgets with a specified focus image; specifically, in my new code `mark_focused` is only set to true where a focus image is not specified. This seems more straightforward to me.

CodingJellyfish:
The code looks good to me.

CodingJellyfish:
Thank you so much for your contribution.

PR #5249, opened 2025-01-16
Make SSAO correct

CodingJellyfish:
How could the current SSAO avaliable to users?
## Agreement
```

```

### DIFF
--- a/data/shaders/IBL.frag
+++ b/data/shaders/IBL.frag
@@ -1,6 +1,7 @@
 uniform sampler2D ntex;
 uniform sampler2D dtex;
 uniform sampler2D albedo;
+uniform sampler2D ssao;
 
 #ifdef GL_ES
 layout (location = 0) out vec4 Diff;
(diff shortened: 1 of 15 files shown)


CodingJellyfish:
Before:
image

After: 
image

Alayan-stk-2:
The SSAO visualization looks a lot better with the changed code! I'll have to test in-game to see the effect of the change on the general look of a track.

Obviously a more correct implementation is preferred, but are you expecting a change in performance behavior? I guess we'll have to test, it's useful to know about even if it won't change the decision to merge.

CodingJellyfish:
> The SSAO visualization looks a lot better with the changed code! I'll have to test in-game to see the effect of the change on the general look of a track.
> 
> Obviously a more correct implementation is preferred, but are you expecting a change in performance behavior? I guess we'll have to test, it's useful to know about even if it won't change the decision to merge.

The performance is overall doubled.

CodingJellyfish:
Now it looks like 
image

PR #5274, opened 2025-01-25
Remove 3 spaces for consistency with other strings

scootergrisen:
## Agreement
```

```
For consistency with other strings.

Both of these exists I'm just removing the one with the extra space:

    Locked: solve active challenges to gain access to more!" without space already exists.
    Locked : solve active challenges to gain access to more!" without space already exists.

### DIFF
--- a/src/states_screens/arenas_screen.cpp
+++ b/src/states_screens/arenas_screen.cpp
@@ -328,7 +328,7 @@ void ArenasScreen::buildTrackList()
         Track *curr = tracks.get(i);
         if (PlayerManager::getCurrentPlayer()->isLocked(curr->getIdent()))
         {
-            w->addItem( _("Locked : solve active challenges to gain access to more!"),
+            w->addItem( _("Locked: solve active challenges to gain access to more!"),
                         "locked", curr->getScreenshotFile(), LOCKED_BADGE );
         }
         else if (PlayerManager::getCurrentPlayer()->isFavoriteTrack(curr->getIdent()))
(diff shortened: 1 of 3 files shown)


CodingJellyfish:
Sorry it will break translation string.

scootergrisen:
That was the whole point.
I saw them in the translation and want to improve it.

Both of these exists I'm just removing the one with the extra space:

    Locked: solve active challenges to gain access to more!" without space already exists.
    Locked : solve active challenges to gain access to more!" without space already exists.

scootergrisen:
How about now?
Not sure what you mean by agreement part.

Alayan-stk-2:
> How about now? Not sure what you mean by agreement part.

Read the text that's in a new PR's text field before deleting it...

> Both of these exists I'm just removing the one with the extra space:

Both exist for the locked one, but not for the anti-aliasing off. Of course, this last one also ought to be corrected, but doing it now would mean having to go over all the translations that had the previous string and fixing them. Not great soon before a release.

scootergrisen:
Translators might be used to strings being changed so i dont consider it a big deal changing already translated strings.
And I don't think you have to do a search/replace in the translations to help out the poor translatons so 30 translators don't have to translate the string again.
It's what translators do.
If they know the translations and know how to use tools like Transifex it's easy. They can look at the differences between the old and new source string and copy the old translation and remove the space. It don't take much time.

Alayan-stk-2:
You assume, incorrectly, that translators for all relevant languages will check on the status of translations before the next release. The changes are easy to do, but you need someone to do them. Breaking strings after a string freeze is problematic. The only way this doesn't break translation for multiple languages is me going to manually update it.

scootergrisen:
Ok just forgot about it then.

PR #5275, opened 2025-01-25
Remove newline

scootergrisen:
So the text does not appear as "...copy[newline]this error[newline]to..." in the dialog.

### DIFF
--- a/src/utils/crash_reporting.cpp
+++ b/src/utils/crash_reporting.cpp
@@ -101,7 +101,7 @@
                 getCallStack(callstack);
 
             std::string msg =   "SuperTuxKart crashed!\n"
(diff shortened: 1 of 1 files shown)


qwertychouskie:
Good catch, LGTM

CodingJellyfish:
There will be problem of dual licensing if you delete agreement from the description...

Alayan-stk-2:
I think the fact that our forum isn't open to new posts and that the message leads user to a place where they can't actually report the issue is a much bigger concern than this insignificant newline.

Alayan-stk-2:
I fixed the reference to freegamedev and removed the new line in the process. I just hope the dialog text is going to properly break the line instead of overflowing...

PR #5291, opened 2025-01-31
Fix ImageMagick 7 deprecation warning

S0nter:
Improve android build scripts by fixing ImageMagick deprecation warnings

## Agreement
```

```


deveee:
I was thinking about it too. But maybe wait a bit more because it's still fairly new. I still have imagemagick 6.9 on Ubuntu 24.10 that doesn't have magick command.

or maybe something like this (I didn't test)
```
if [ -f "/usr/bin/magick" ]; then
    im_cmd="magick"
else
    im_cmd="convert"
fi
```
Probably there is better way to check if that command exists.

S0nter:
> or maybe something like this (I didn't test)

It seems to work.
`convert` supports input filename anywhere in the arguments. `magick` only supports specifying the input filename before manipulations like `-scale`, so this approach seems like the way to go.

> Probably there is better way to check if that command exists.

```bash
# We can use `command` builtin to test the existence of an executable:
if command -v magick > /dev/null; then
    echo Magick exists
else
    echo Magick not found
fi
```

Alayan-stk-2:
Thank you for the PR!

PR #5299, opened 2025-02-02
Put the macOS icon in a rounded rectangle

maksalees:
The current icon for macOS does not follow design guidelines introduced in Big Sur and therefore looks out of place. I placed the icon in a rounded rectangle with the game background to make it look like the rest of macOS icons.
# Before
image
# After
image
## Agreement
```

```


Alayan-stk-2:
Some questions:
- Doesn't your change affect all OSes, not just MacOS?
- Is it really necessary to use almost a MB on icons? That's 30 times what it used to be.

The current application icon of STK could be better of course, even beyond it fitting or not with other icon styles.

qwertychouskie:
> Doesn't your change affect all OSes, not just MacOS?

`.icns` is a macOS-specific format.

maksalees:
> Some questions:
> 
> * Doesn't your change affect all OSes, not just MacOS? EDIT: When I try to visualize the old .icns file, I get shown the various supertuxkart_xx.png files, when I download the new one and try to check its content I just see the 16x16 picture. It might be related to me not running MacOS, but I'm not sure how I can check the file.
> * Is it really necessary to use almost a MB on icons?
> 
> The current application icon of STK could be better of course, even beyond it fitting or not with other icon styles.

When I view the old .icns file, I only see one 128x128 image. The new file contains several images of different resolutions, the largest being 1024x1024. Having high resolution icons is necessary for them to display well on high resolution screens.

PR #5300, opened 2025-02-03
Implement PCSS & Revert graphics level 8

CodingJellyfish:
## Agreement
```

```
4096 shadow map is still available in config.xml, but I don't recommend to use it.


CodingJellyfish:
image

PR #5302, opened 2025-02-06
Revert recent camera change & fix #5270

### LINKED ISSUE #5270: Remove camera panning at the start of races or when switching between backward and forward
STK has camera smoothing, which is really useful when driving, especially when turning.

But there are a few situations where it seems to do more harm than good:
- At the start of the race, the default camera position is incorrect. Instead of immediately going to the correct camera position, it takes a little time to adjust, and it looks quite weird in the process. This can also be tested in the overworld going in and out of the kart selection screen.
- Likewise when restarting a race, the camera remembers what was its orientation before the reset, and turns back to the normal orientation. For example, if you turn 90° right from the default position and then restart the race, you'll see the camera start pointing towards the right and moving back towards the normal direction.
- When looking backward and going back to the forward camera, if the kart has turned, the forward camera will take some time to adjust its orientation instead of being instantly correct. In normal races, players rarely look back for more than a split second, but in soccer using the ball-camera, the kart can turn significantly before going back to the normal view. This is especially annoying when moving towards the ball already.

A common feature of all these situations is that we just entered the current camera view, so there is a clearly different image. There is a hard transition no matter where the camera starts - and a hard transition plus a weird panning from an incorrect position is worse than just a hard transition.

CodingJellyfish:
## Agreement
```

```


zikzaksuper:
idk if it was these camera changes but im pretty sure i wasnt getting motion sick when the cart slides down a steep slope while spinning

PR #5307, opened 2025-02-12
Make three strike battle any other strike battle and add an option to steal tires

zikzaksuper:
## Description

adds a spinner widget (or rather, repurposes an existing one) and a checkbox

## Issues

i made a mistake by not creating a branch until i already coded all of that in a couple of days ago.
even though i have checked all the modifications to see if i had accidentally removed something please check again just in case.

## Agreement
```

```


src/modes/three_strikes_battle.cpp
```diff
@@ -542,22 +553,21 @@ void ThreeStrikesBattle::getKartsDisplayInfo(
...
-            color = video::SColor(255, 255, 0, 0);
-            break;
-        case 0:
-            color = video::SColor(255, 128, 128, 128);
-            break;
+        float lerp_time;
+        lerp_time = float(m_kart_info[id].m_lives-1)/float(UserConfigParams::m_ffa_time_limit);
+        int lerp_lower = floor(lerp_time*gradientLength);
+    
+        color = video::SColor(255, lerp(redGradient[lerp_lower+1],redGradient[lerp_lower+2],lerp_time*gradientLength-lerp_lower), lerp(greenGradient[lerp_lower+1],greenGradient[lerp_lower+2],lerp_time*gradientLength-lerp_lower), lerp(blueGradient[lerp_lower+1],blueGradient[lerp_lower+2],lerp_time*gradientLength-lerp_lower));
```
kimden, line 570: For readability, please don't use lines that are that long, and use spaces around operators, after commas, etc. (not just here).

Regarding this particular line, it might be better to introduce an auxiliary function to find the color to not retype the same code three times for three gradient arrays, especially because it's used in another place.

src/states_screens/track_info_screen.cpp
```diff
@@ -251,14 +253,23 @@ void TrackInfoScreen::init()
...
         m_target_value_spinner->setValue(UserConfigParams::m_ffa_time_limit);
+        if (UserConfigParams::m_use_ffa_mode)
+            m_target_value_label->setText(_("Maximum time (min.)"), false);
+            
+		else
+		    m_target_value_label->setText(_("Starting tires"), false);
+        
+        m_tire_stealing->setVisible(!UserConfigParams::m_use_ffa_mode);
+        getWidget<LabelWidget>("tire-stealing-text")->setVisible(!UserConfigParams::m_use_ffa_mode);
+		getWidget<LabelWidget>("tire-stealing-text")->setText(_("Steal tires"), false);
```
kimden, line 268: Each tab → 4 spaces

src/modes/three_strikes_battle.hpp
```diff
@@ -150,6 +150,16 @@ class ThreeStrikesBattle : public WorldWithRank
     bool spareTireKartsSpawned() const;
     // ------------------------------------------------------------------------
     void spawnSpareTireKarts();
+    // ------------------------------------------------------------------------
+    const int redGradient[12] = {30, 64, 255, 255, 255, 128, 0, 0, 0, 0, 0, 0};
+    // ------------------------------------------------------------------------
+    const int greenGradient[12] = {0, 0, 0, 128, 255, 255, 255, 128, 255, 128, 0, 0};
```
kimden, line 156: These arrays are used in the implementation only, it might be better to move them to .cpp into e.g. anonymous namespace

src/modes/three_strikes_battle.cpp
```diff
@@ -55,16 +56,38 @@ ThreeStrikesBattle::ThreeStrikesBattle() : WorldWithRank()
...
+    // ------------------------------------------------------------------------
+    /* Must be two lower than the actual length, the first and last items in the 
+     array are just buffers to prevent floating-point disasters */
+    const int gradientLength = 10;
+
+    video::SColor calculateColor(int lerp_index, float lerp_time)
+    {
+        int redVal = lerp(redGradient[lerp_index + 1], redGradient[lerp_index + 2], lerp_time);
+        int greenVal = lerp(greenGradient[lerp_index + 1], greenGradient[lerp_index + 2], lerp_time);
+        int blueVal = lerp(blueGradient[lerp_index + 1], blueGradient[lerp_index + 2], lerp_time);
```
kimden, line 86: This could be a separate function on an array/vector to invoke three times, but I suppose at this point it's kinda ok already.

kimden (commented):
Code style is alright now I suppose.

kimden:
I tested the changes a bit.
- In the original 3 strikes mode, I think it was impossible to get more than 3 lives. Testing the mode without stealing, you can get a spare kart and get more than the initial number.
- If the initial number of lives is 1, you get two wheels for the kart :)
- Colors feel very dark, and even for 3 lives for some reason 0 is dark green-yellow, 1 is very dark red, 2 is yellow-green, 3 is cyan. My opinion is subjective but this doesn't seem to be an intuitive color scheme.
- If you go above the initial amount due to stealing or taking spare karts, your color is unpredictable:
image
- I'm not sure whether community and/or developers approve the general idea. I find it okayish.

src/modes/three_strikes_battle.cpp
```diff
@@ -55,16 +56,39 @@ ThreeStrikesBattle::ThreeStrikesBattle() : WorldWithRank()
...
+    The first and last items in the array must be the same.*/
+    const int gradientLength = 9;
+
+    video::SColor calculateColor(int lerp_index, float lerp_time)
+    {
+        int redVal = lerp(redGradient[lerp_index], redGradient[lerp_index + 1], lerp_time);
+        int greenVal = lerp(greenGradient[lerp_index], greenGradient[lerp_index + 1], lerp_time);
+        int blueVal = lerp(blueGradient[lerp_index], blueGradient[lerp_index + 1], lerp_time);
+        return video::SColor(255, redVal, greenVal, blueVal);
+    }
```
kimden, line 89: This function is called twice in two different ifs, those ifs seem identical and related only to the color logic, and not to the game mode itself. Pass the current and the maximum number of lives to this function, and let it do all the transformations with `lerp_***` inside. Don't forget to put spaces around operators `-*/`.

lerp is still repeated three times. Please move it into a separate function or lambda, e.g. return `video::SColor(255, f(redGradient), f(greenGradient), f(blueGradient))` with
```
auto f = [=](const int* a) -> int {
    return lerp(a[lerp_index], a[lerp_index + 1], new_lerp_time);
};
```

kimden (commented):
Other than these, the code seems fine. Colors are better, I'm not sure I like the changing colors over max value, but I'll let developers decide.

src/modes/three_strikes_battle.cpp
```diff
@@ -542,22 +577,17 @@ void ThreeStrikesBattle::getKartsDisplayInfo(
...
-        case 2:
-            color = video::SColor(255, 255, 229, 0);
-            break;
-        case 1:
-            color = video::SColor(255, 255, 0, 0);
-            break;
-        case 0:
-            color = video::SColor(255, 128, 128, 128);
-            break;
+        color = video::SColor(128,128,128,0);
```
kimden, line 586: Is this an intentionally semi-transparent yellow, or it should have been `video::SColor(255, 128, 128, 128);` ?

src/karts/controller/spare_tire_ai.cpp
```diff
@@ -142,12 +143,12 @@ void SpareTireAI::crashed(const AbstractKart *k)
...
 
-    // Tell players that they can have at most 3 lives
+    // Add a life
     RaceGUIBase* r = World::getWorld()->getRaceGUI();
-    if (m_tsb_world->getKartLife(k->getWorldKartId()) == 3)
+    if (m_tsb_world->getKartLife(k->getWorldKartId()) == UserConfigParams::m_ffa_time_limit && !UserConfigParams::m_tire_steal)
     {
         if (r)
-            r->addMessage(_("You can have at most 3 lives!"), k, 2.0f);
+            r->addMessage(_("You can have at most %d lives!"), k, 2.0f);
```
kimden, line 151: I'm not an expert but I suppose this would need a pluralized form. Wait until someone else's feedback please.

kimden (approved):

src/modes/three_strikes_battle.cpp
zikzaksuper: i think i wrote that piece of code before noticing the alpha is the first parameter

Alayan-stk-2:
There are some issues:
- The PR mixes two different concepts. It's not rare for new contributors to do that (I did it too in some of my first PRs) but it's not good practice especially as one change may be less problematic than the other.
- With 3 lives or less, the visual spare tires on the karts work as a representation, but if someone pick more lives, there is no such representation I think. While I can see the point of wanting to play with say 5 starting lives, supporting significantly higher values by default seems very niche.

Then there are some code details. I haven't reviewed in detail, but the text needs plural translation handling and some of kimden's comments look like they may apply still.

zikzaksuper:
yes, i have noticed that the tires show the lives but i really dont know how to fix it (not code-wise, i dont have an idea on how to do it at all). but from a game design perspective it seems strange to not implement something that adds a gameplay feature because it would negatively impact a minor visual one. not a game designer though.

for the translations, i have no clue how its done still, but if i ever try checking this again i will look into how to do that..

also, it might just be me but i very much enjoy playing with more lives, specifically 10-20 (depends on how much time i have), but i also really like battle mode in general so im not sure about others

PR #5369, opened 2025-03-11
Upgrade FPS

Peter34596:
I consider that upgrading fps from 25 to 30 is more fluent stereotype devices. Likewise, thank you, greetings form Argentina


CodingJellyfish:
Looks okay though I don't know if there's smth not specifying its fps...

CodingJellyfish:
Sorry this pr will break physics and is overall useless.

Peter34596:
It was a school work i change a random varablie, thx for all

PR #5374, opened 2025-03-21
Update cutscene_gui.cpp

iamnytik:
This update will move the subtitles from the bottom of the screen a few pixels above.This is intended so that the subtitles stay above the Skip buttons in cutscenes

## Agreement
```

```

### DIFF
--- a/src/states_screens/cutscene_gui.cpp
+++ b/src/states_screens/cutscene_gui.cpp
@@ -54,7 +54,7 @@ void CutsceneGUI::renderGlobal(float dt)
 
     if (m_subtitle.size() > 0)
     {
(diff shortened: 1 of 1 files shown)


CodingJellyfish:
Hi.
My suggestion is use 4 instead of 10. The skip button itself cannot be higher than 3 fonts.

iamnytik:
Hey would it be better if it was 6? because if the text is really long it might come very close to the skip button
image

This how it looks if it's 6.It might give a little spacing from the button
image

CodingJellyfish:
Please test very large font size to prevent text go to the center of the screen

iamnytik:
Very large font:
image
Large font:
image
Very small font:
image

PR #5399, opened 2025-04-17
added random track button outside the track chooser

Hanuko33:
## Agreement
```

```
##
Adds the "random track" button outside the track chooser, because once you have >100 tracks, then finding the "random track" button is very annoying. I can't count the amount of times I had to scrool through my whole "gallery" of maps, just to find the "random track" button.

The code must be cleaned up (the button functionality is very similar to the one in the track chooser and I didn't know what's the best way to make it good).
There must be work on UX, (the button "fits/feels" weirdly in its current place)

### DIFF
--- a/data/gui/screens/tracks_and_gp.stkgui
+++ b/data/gui/screens/tracks_and_gp.stkgui
@@ -25,7 +25,7 @@
             <header width="30%" I18N="In the track and grand prix selection screen" text="All Tracks"
                     align="center" text_align="center" />
             <textbox width="30%" id="search"/>
-            <icon-button id="random_track" width="fit" height="100%" icon="gui/icons/track_random.png" I18N="In the track and grand prix selection screen" text=""/>
+            <icon-button id="random_track" width="10%" height="100%" icon="gui/icons/track_random.png" I18N="In the track and grand prix selection screen" text=""/>
             <div width="fit" height="100%" layout="horizontal-row" align="center">
                 <checkbox id="favorite"/>
                 <tiny-header width="fit" height="100%" I18N="In the track and grand prix selection screen" text="Edit favorite tracks"/>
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
The idea is interesting, but you'll understand I won't merge what you yourself calls a hacky solution.

Hanuko33:
If I improve the solution, will it be considered for merge?

Hanuko33:
Should it be the same for karts?

Alayan-stk-2:
I tested your code.

The code changes are rather simple and the button works.

The main issue as I see it is that the button is really tiny so it looks weird. But I understand why it's there, as there is no obvious space to put it without disrupting other elements.

RandomTrackButton

It could perhaps be placed under the random GP box which has some space to be moved up but to allow it to use the further height you'd need to reorganize the box structure of the screen to have the list of GPs and the control buttons inside a div on the left and the random GP and random kart buttons in a div on the right.

> Adds the "random track" button outside the track chooser, because once you have >100 tracks, then finding the "random track" button is very annoying. I can't count the amount of times I had to scrool through my whole "gallery" of maps, just to find the "random track" button.

I don't know if you are aware, but if you just scroll left instead of right, you will immediately find the random track button.

The issue is still relevant for people that have 100+ tracks, but with the game now willing to display more than 20 tracks at once (40 are displayed on my screenshot above) it's also less relevant than it used to be, as scrolling through in the wrong direction will be faster.

> Should it be the same for karts?

The concerns of tiny button and of location in already very busy screen are also here. I'm not sure how necessary it is considering that now the game is willing to display more than 20 karts at once (40 is easily displayed at 1600x900), and that you can scroll left if you know from the start you want the random kart option.

Alayan-stk-2:
I merged your code. Thank you for the PR and congrats. :)

For the adjustments, I added my own GUI layout changes on top with commit 9d9c74f to do what I was thinking about. I thought it would be simpler to do it myself than to try to explain in details what I meant, as I had to change the underlying structure a lot. Screenshot below.

Random_track_button2

Hanuko33:
:)

PR #5405, opened 2025-04-30
Remove PrefersNonDefaultGPU from desktop file

Jan200101:
## Agreement
```

```

PrefersNonDefaultGPU is currently very broken, see https://github.com/ValveSoftware/steam-for-linux/issues/9940

### DIFF
--- a/data/supertuxkart.desktop
+++ b/data/supertuxkart.desktop
@@ -99,4 +99,3 @@ Keywords[zh_CN]=企鹅、游戏、赛车
 Keywords[zh_HK]=tux;game;race;
 Keywords[zh_TW]=tux;game;race;
 Keywords=tux;game;race;
-PrefersNonDefaultGPU=true


qwertychouskie:
I agree that it's somewhat broken by design, but removing it would break STK on more systems.  Laptops with a dual-GPU setup are much more common than desktops with a dual-GPU setup.  Until a better mechanism exists, this should remain.

Jan200101:
I disagree, most CPUs these days have a GPU integrated, combined with a full dedicated GPU you run into this issue.

Jan200101:
it might also be worth noting that the Fedora maintainer of the stk package has decided to patch the key out to address a bug report

https://src.fedoraproject.org/rpms/supertuxkart/c/8e143f6cd414e7f9559565f920a9b30b72cfc452?branch=rawhide
https://bugzilla.redhat.com/show_bug.cgi?id=2346005

Alayan-stk-2:
The entire design of this setting is appalling. It's a pick your poison situation...

Jan200101:
the right solution to this would be to fix the issue itself instead of trying to remove the key, however my patches have thus far not gone anywhere
https://gitlab.freedesktop.org/hadess/switcheroo-control/-/merge_requests/69

on the off-chance you know anyone that can help with this it might be worth sharing this with them
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/5PUGDWKQJQ7LZNAUXAINVITTNWWGZKN7/

Alayan-stk-2:
> the right solution to this would be to fix the issue itself instead of trying to remove the key, however my patches have thus far not gone anywhere

That would be the best outcome indeed



> on the off-chance you know anyone that can help with this it might be worth sharing this with them

I don't have such contacts unfortunately

Alayan-stk-2:
@Jan200101 It looks like your PR to switcheroo-control got merged. That's some good news.

This should allow to detect the presence of a more powerful discrete GPU if available ; but so far I assume that PrefersNonDefaultGPU still has the same issues and that there is not yet a new setting we could use instead?

Jan200101:
> This should allow to detect the presence of a more powerful discrete GPU if available ; but so far I assume that PrefersNonDefaultGPU still has the same issues and that there is not yet a new setting we could use instead?

`PrefersNonDefaultGPU` is in the process of being fixed
KDE already fixed it ([merge](https://invent.kde.org/frameworks/kio/-/merge_requests/1556))
Gnome is bike-shedding it ([merge](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3193/diffs))
Cinnamon also fixed it ([merge](https://github.com/linuxmint/xapp/pull/178))
PopOS still has an outdated switcheroo-control fork ([issue](https://github.com/pop-os/switcheroo-control/issues/1) but would otherwise pull in support via gnome-shell

I doubt there will be a new setting within the XDG Desktop spec since it impossible for an app to statically define what it is needs but there is some discussion about a new solution by RedHat and System76 https://www.phoronix.com/news/Better-Linux-Multi-GPU-2025
(though considering switcheroo-control was a RedHat sponsored project I am not sure how much I trust them with maintaining it)

Alayan-stk-2:
Ok, thank you for the information, and for your work to get this problem fixed.

If PrefersNonDefaultGPU is already fixed on some desktop environments and in the process of being fixed on others, I'm going to close this PR.

PR #5412, opened 2025-05-07
Revert "Removed glow under main menu options"

qwertychouskie:
This complete breaks the Cartoon theme unfortunately, as the highlight effect is very subtle vs the classic theme.  Open for a better solution, but this should certainly be fixed before 1.5 is released as it's a pretty big usability issue.

### DIFF
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -1424,11 +1424,10 @@ void Skin::drawRibbonChild(const core::recti &rect, Widget* widget,
         }
 
         const bool mark_focused =
-            (focused || (parent_focused && parentRibbonWidget != NULL &&
+            focused || (parent_focused && parentRibbonWidget != NULL &&
                           parentRibbonWidget->m_mouse_focus == widget) ||
                        (mark_selected && !always_show_selection &&
-                          parent_focused)) &&
-                        widget->m_properties[PROP_FOCUS_ICON].size() == 0;
+                          parent_focused);
 
         /* draw "selection bubble" if relevant */
         if (always_show_selection && mark_selected)


Alayan-stk-2:
Since these menu icons already use a different icon (the "_focus" versions), I'd prefer updated focus icon with a clearer difference compared to the base version. You perhaps can't change the color within too much, but the outline could receive an highlight color?

qwertychouskie:
> Since these menu icons already use a different icon (the "_focus" versions), I'd prefer updated focus icon with a clearer difference compared to the base version. You perhaps can't change the color within too much, but the outline could receive an highlight color?

In a perfect world, sure, but I have a lot on my olate right now, so I don't have the time to figure out a design that makes the selection clear, update all the icons, etc.  Also, TBH I don't consider having the glow in the Classic theme to be a bug, like I see the argument that it's redundant, but I'd argue that it's better for consistency to always have the glow.

Alayan-stk-2:
@qwertychouskie I think you missed my question above.

qwertychouskie:
> @qwertychouskie I think you missed my question above.

Sorry for the slow response, been a lot on my plate lately.  https://github.com/supertuxkart/stk-code/issues/4066#issuecomment-2079967700 has SVG icons for both the classic and cartoon skins, so it should probably have what you are looking for.

Alayan-stk-2:
Closing, I think https://github.com/supertuxkart/stk-code/commit/5d299e3f5ff44a1ef39c82fe5018c241549fb229 addresses your concerns well enough

SvenAndreasBelting:
I changed the cartoon focus icons, I think it's better now.

PR #5444, opened 2025-05-30
Add RTL Marker after name, not before it. Fixes #5443.

### LINKED ISSUE #5443: StringUtils::startsWith works incorrectly with RTL characters
### Description

One of reported use cases: sending messages that should look like `kimden: السلام عليكم` ("hello" in Arabic) no longer works because of #5121 and because `startsWith("‏kimden: السلام عليكم", "kimden: ")` apparently returns false using strncmp.

While it may be better to resolve 5121 differently now that apparently only one player is displayed in the message from a single peer, StringUtils works incorrectly anyway, hence the issue.

### Steps to reproduce

1. Send a message on an online server containing right to left writing system characters.

### Configuration

latest master 4dbe097195054c5e785e77e19f6d9ff8b735699f
git
Ubuntu 23.10

### Error message

​

### Log File (stdout.log)

​

### Input Configuration (input.xml)

​

nutzboi:
## Agreement
```

```

### DIFF
--- a/src/network/protocols/client_lobby.cpp
+++ b/src/network/protocols/client_lobby.cpp
@@ -1488,6 +1488,7 @@ void ClientLobby::sendChat(irr::core::stringw text, KartTeam team)
         // Make message look better by aligning to left or right side depends
         // on name and text
         // (LTR: RTL text always right; RTL : LTR text always left)
+        core::stringw name_suffix (": ");
 #ifndef SERVER_ONLY
         if (!name.empty() && !text.empty())
         {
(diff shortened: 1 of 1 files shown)


nutzboi:
On second thought, this would fail if the name itself is RTL.

Alayan-stk-2:
The entire point of that block of code is to align-right or align-left based on the text typed (someone typing in english in an online chat  with the game set in arabic would have the english text be shown properly on the left). Your change breaks this functionality, and is functionally no different from removing the entire block of code.

PR #5459, opened 2025-06-15
Fix use of uninitialized variable in PowerupManager

nyllet:
Also avoid a potenial null pointer dereference
Remove two unused includes
Minor cleanups

## Agreement
```

```


src/items/powerup_manager.cpp
```diff
@@ -219,7 +216,7 @@ void PowerupManager::WeightsData::reset()
 void PowerupManager::WeightsData::readData(int num_karts, const XMLNode *node)
 {
     m_num_karts = num_karts;
-    for (unsigned int i = 0; i < node->getNumNodes(); i++)
+    for (unsigned int i = 0; i < node->getNumNodes(); ++i)
```
Alayan-stk-2, line 219: Leave i++ in loops. The vast, vast majority of loops in the STK code base use post-increment notation, so this change creates inconsistencies.

In for loops, both pre-increment and post-increment will generate exactly the same code, so this is purely a stylistic matter.

src/items/powerup_manager.cpp
```diff
@@ -232,12 +229,19 @@ void PowerupManager::WeightsData::readData(int num_karts, const XMLNode *node)
 
         // Keep a reference for shorter access to the list
         std::vector<int> &l = m_weights_for_section.back();
-        for(unsigned int i=0; i<l_string.size(); i++)
+        for(unsigned int j=0; j < l_string.size(); ++j)
```
Alayan-stk-2, line 232: Same remark about ++j vs j++

Switching to j instead of i is definitely good for code clarity.

Alayan-stk-2 (changes requested):
Thank you for the PR!

Please adjust the mentioned stylistic elements and it can be merged.

PR #5463, opened 2025-06-24
Replace random_shuffle with std::shuffle

nyllet:
## Agreement
```

```

### DIFF
--- a/src/utils/random_generator.cpp
+++ b/src/utils/random_generator.cpp
@@ -18,9 +18,6 @@
 
 #include "utils/random_generator.hpp"
 
-#include <stdlib.h>
-#include <ctime>
-
 //std::vector<RandomGenerator*> RandomGenerator::m_all_random_generators;
 
 RandomGenerator::RandomGenerator()
@@ -28,8 +25,20 @@ RandomGenerator::RandomGenerator()
     m_a = 1103515245;
     m_c = 12345;
     //m_all_random_generators.push_back(this);
-    m_random_value = 3141591;
+    m_random_value = default_seed;
 }   // RandomGenerator
+
+std::mt19937& RandomGenerator::getGenerator() {
+   static thread_local std::mt19937 generator = [] {
+      if (m_random_value == default_seed)
+         return std::mt19937{std::random_device{}()};
+      else
+         return std::mt19937{m_random_value};
+   }();
+   return generator;
+}
+
+unsigned int RandomGenerator::m_random_value = RandomGenerator::default_seed;
 
 // ----------------------------------------------------------------------------
 #if 0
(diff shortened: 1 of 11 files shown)


nyllet:
std::random_shuffle has been deprecated since c++14. this is my attempt at replacing it

src/utils/random_generator.cpp
```diff
@@ -18,19 +18,28 @@
...
+   static thread_local std::mt19937 generator = [] {
+      if (m_random_value == default_seed)
+         return std::mt19937{std::random_device{}()};
+      else
+         return std::mt19937{m_random_value};
+   }();
+   return generator;
+}
+
+unsigned int RandomGenerator::m_random_value = RandomGenerator::default_seed;
```
Alayan-stk-2, line 41: What's the purpose of this line here?

deveee:
I don't know this code so maybe it doesn't matter, just note that any changes in random generator may affect network multiplayer.

src/utils/random_generator.cpp
nyllet: maybe i misunderstood the question. do you mean what the default_seed is used for? only to check if a seed was provided from the command line or not. if no seed is provided, the random generator will be initialized like this: std::mt19937{std::random_device{}()}, so the default_seed will never actually be used as a seed

nyllet:
regarding the comment from @deveee regarding network multiplayer, i don't think it is affected yet. there are hints in this comment in random_generator.hpp: "// This generator is (currently) not good enough, i.e. it often gives
    // long sequences of same numbers. And the seeding is not done (the
    // mid term goal is to synchronise all random number generators on
    // client and server to make less communication necessary).
    // For now: just use standard random numbers:"
that someone thinks random_generator should be improved and used in network play in the future, but i don't think that is the case yet. i searched for random_generator in the files under src/network and the only hit was that random_generator.hpp is included from server_lobby.cpp, but it is not used. i also think that the important part if we want to have the same random number sequence in all clients and server in network game is the synchronization, not how the numbers are generated. i have not touched code related to synchronization as far as i understand at least

Alayan-stk-2:
How the numbers are generated matters for networked play, because the client may not get the authoritative data from the server for 50-200ms depending on ping.

However, RNG for powerups picked from gift boxes uses a special system, so as to make client-side prediction effective. The relevant code can be seen from here on: https://github.com/supertuxkart/stk-code/blob/6260dc8e330d9bf91387b7bec9dee5c95c3041dd/src/items/powerup.cpp#L552

I also had concerns about potential compatibility issues, but from what I can tell most uses from this random generator are for things such as giving a random track when picking the random track button. The one exception is picking nodes to spawn tire karts in 3-strikes battle, but online is FFA instead of 3 strikes anyways.

PR #5466, opened 2025-07-02
Fix GLES sampler precision issue for advanced pipeline

Icenowy:
Currently we do not add precision description to sampler2D's, and the default precision defined by ESSL spec is lowp.

Experimentation on an Imagination AXM-8-256 GPU shows that the advanced pipeline needs highp sampler2D, otherwise some distant objects will glitch (noticable at the start point of black_forest map).

Manually add precision modifiers for sampler2D to fix the glitch.

Tested on Eswin EIC7700 with X11+EGL+GLES with Imagination DDK 23.2 driver.

P.S. setting mediump instead of highp on sampler2D's seems to be not enough for resolving the glitch; maybe hardwares w/o highp support need to be forbidden from advanced pipeline or some shaders need to be rewritten to be mediump-safe.

## Agreement
```

```

### DIFF
--- a/src/graphics/shader_files_manager.cpp
+++ b/src/graphics/shader_files_manager.cpp
@@ -202,12 +202,14 @@ ShaderFilesManager::SharedShader ShaderFilesManager::loadShader
         code << "precision highp float;\n";
         code << "precision highp sampler2DArrayShadow;\n";
         code << "precision highp sampler2DArray;\n";
+        code << "precision highp sampler2D;\n";
     }
     else
     {
         code << "precision mediump float;\n";
         code << "precision mediump sampler2DArrayShadow;\n";
         code << "precision mediump sampler2DArray;\n";
+        code << "precision mediump sampler2D;\n";
     }
 #endif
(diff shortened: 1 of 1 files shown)


Icenowy:
Re the P.S.: Looks like highp is core (instead of optional) to ESSL3.0, and as the whole advanced shader-based pipeline requires ESSL3.0, maybe the fallback mediump codepath isn't needed?

Alayan-stk-2:
> Re the P.S.: Looks like highp is core (instead of optional) to ESSL3.0, and as the whole advanced shader-based pipeline requires ESSL3.0, maybe the fallback mediump codepath isn't needed?

@deveee Do you know anything about that?

deveee:
For GLES2 it uses only very simple irrlicht renderer (can be tested with force legacy device in config.xml). For shader-based pipeline it needs GLES3 for both enabled and disabled advanced lighting. And indeed it seems that highp is core feature and I probably added glGetShaderPrecisionFormat() just to be safe. 

Btw there is also
```
    if (CVS->isGLSL())
        code << "#version 300 es\n";
```
and isGLSL() should be always true.

Maybe better change after release.

PR #5487, opened 2025-08-12
Add a speedmeter on multitouch

CreamBR-STK:
Finnally!!! A way to know your speed on Android (and iOS).

Anyway, this add the same speedmeter from PC to Android in left of the control buttons (skid, nitro, etc).

## Agreement
```

```

### DIFF
--- a/src/states_screens/race_gui.cpp
+++ b/src/states_screens/race_gui.cpp
@@ -374,20 +374,16 @@ void RaceGUI::renderPlayerView(const Camera *camera, float dt)
 
     if (!isSpectatorCam) drawPlungerInFace(camera, dt);
 
-    // Scale race GUI along screen size
-    scaling *= sqrtf(float(viewport.getWidth()) / 800.0f);
-    scaling *= sqrtf(float(viewport.getHeight()) / 450.0f); 
-
-    // Scale X and Y separately in splitscreen for proper relative positioning
     if (viewport.getWidth() != (int)irr_driver->getActualScreenSize().Width ||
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
1 - Your PR touches scaling code which it has no reason to. You mentioned on IRC using a old source file as a base, that's not a good idea before making a PR.
2 - Your PR adds a lot of code duplication. The MultiTouchSpeedRank function copy-pastes a lot from the original function used for PC.
3 - Because of the code duplication, it also means that updating for the 2.0 branch would cause additional issues.

There may be other issues, but that's what I can see at a glance.

Gustanator:
I will remake this code, but i want to use this code in 1.5 (too hard to compile the Android Version, i'm using the multitouch "hack" on PC)

Gustanator:
I update the code, removing the MultiTouchSpeedRank function, and using the SpeedRank instead (with two new if-cases)

Alayan-stk-2:
Why does the entire .hpp file show up as modified? Especially as at a glance the content doesn't really seem to have changed.

Perhaps some kind of encoding issue, maybe try to start again from the original .hpp file present in master and apply any needed change to it.

Gustanator:
LMAO, I use the original .hpp

Anyway, I sent now a mail to you

Gustanator:
I put the original .hpp file now

I've made modifications in this file (in the 1st version), but I remove this mods

Gustanator:
I messed my code :P

I will open another PR.

PR #5490, opened 2025-08-15
Add a speedmeter on multitouch

GustaLOLE:
Speedmeter in left of the power and see behind buttons

Now, all is over

## Agreement
```

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 SuperTuxKart is a free kart racing game. It focuses on fun and not on realistic kart physics. Instructions can be found on the in-game help page.
 
-The SuperTuxKart homepage can be found at <https://supertuxkart.net/>. There is also our [FAQ](https://supertuxkart.net/FAQ) and information on how get in touch with the [community](https://supertuxkart.net/Community).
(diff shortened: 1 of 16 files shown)


Alayan-stk-2:
@deveee Since this is a change meant for the android/mobile builds, can you perhaps have a look if the code works as intended and how it feels?

This updated code is a lot cleaner than the first PR, but I think there might be concern with overloading the GUI with data, as this adds an additional visual element by default. I know it wouldn't be very discoverable if it required editing a config file, but that may be a safer choice.

Gustanator:
Now, I create a setting to activate/disable the speedmeter (active on PC, disabled on mobile) to have a option to activate this on mobile (or keep disabled for cleaner ui).

(Note: 4 commits because i´m using my father's PC, and he don´t have git, only MSVC :P)

Gustanator:
And AGAIN some files is being deleted and rewritten with my code aaaaaaaaaaaaaa

Gustanator:
I will open a new PR with the final code

This PR is a mess of code insertion :P

PR #5493, opened 2025-08-24
Fix wrong powerup sound when using after a rewind/forced sync

Nomagno:
Known bug for years, happens most often when playing on a server with a custom powerup.xml or when one's internet connection is bad/has a lot of packet loss. This PR simply removes a check in `Powerup::set()` that made sure not to re-create the sound source when rewinding. This check was unnecessary and produced an audio bug where basically the client would decide on an item sound, and never change it regardless of whether the server corrected this item or not. The `Powerup::rewindTo()` function already makes sure not to invoke `Powerup::set()` for cases where the sound source would not need to be recreated (item of the same type), so as far as I know there are no concerns about unneeded memory accesses.

## Agreement
```

```

### DIFF
--- a/src/items/powerup.cpp
+++ b/src/items/powerup.cpp
@@ -149,10 +149,6 @@ void Powerup::set(PowerupManager::PowerupType type, int n)
 
     m_number=n;
 
-    // Don't re-create sound sound during rewinding
-    if (RewindManager::get()->isRewinding())
-        return;
-
     if (m_sound_use != NULL)
     {
         m_sound_use->deleteSFX();


Alayan-stk-2:
If I understand correctly: in case the client mispredicts the powerup (rare event or config mismatch), the client currently plays the sound of the wrong powerup, and this check prevents rewind from fixing it?

I'll have to do some testing to double-check, but if there is already a check to prevent the correct sound from being restarted or played multiple times, then this should indeed work.

Nomagno:
That's correct.

This touches the set function, which just sets up the sound source itself to be played later when the powerup is actually used. I believe playing the sound is handled in a completely different part of the code and I have encountered no such issue in my testing, but yeah probably good to re-verify.

PR #5501, opened 2025-09-09
Added batch downloading

Hanuko33:
## Agreement
```

```
This allow the user to automatically download all of the listed karts/tracks/arenas. This is useful with #5500 as the combination of this pr and that one allows the user to (for example) download all of the featured tracks, with just a single mouse press.

So far it seems to work, excluding the FIXME (mem-leak) and the icon-button doesn't working well with the mouse for some reason (high chance I messed up the .stkgui)

### DIFF
--- a/data/gui/screens/addons_screen.stkgui
+++ b/data/gui/screens/addons_screen.stkgui
@@ -2,6 +2,8 @@
 <stkgui>
     <div x="1%" y="0" width="98%" layout="horizontal-row" height="9%">
         <icon-button id="back" height="100%" icon_align="left" icon="gui/icons/back.png"/>
+        <spacer proportion="0.1" height="1"/>
+        <icon-button id="download_all" height="100%" icon="gui/icons/package-update.png" icon_align="left"/>
         <spacer proportion="1" height="1"/>
         <icon-button id="reload" y="5%" height="90%" icon_align="right" icon="gui/icons/restart.png"/>
     </div>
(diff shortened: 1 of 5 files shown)


Hanuko33:
I found a horrible bug. Switching to a draft.
Dismiss frees the memory and then cancelled is being overwritten.
I didn't notice it during testing, because I was lucky (or unlucky) with the memory.

Hanuko33:
Fixed

Alayan-stk-2:
Being able to download multiple addons without having to manually open a new dialog for every single one of them is definitely a useful QoL feature.

However, even with the featured filter available, downloading all the listed addons is very crude, so I'm not sold on the current version of this PR.

Also, this will probably lead to more people who have tons of addons they never use that just take up space. Of course this is the responsibility of the user more than of the game, but I see a line between making it easier and faster to get multiple addons and making it trivial to just download a few GBs worth of addons from the addons server.

Finally, there is also the question of uninstalling. Having to go back to the extensions menu and searching for the extension you didn't like after testing it is rather unpractical as is. But if someone can install 200 addons with a few clicks, and then has a change of heart and needs to spend half-an-hour clicking to remove them, that's also an issue.

Ideally, the user would be able to manually check/uncheck addons that will get downloaded, and to use the same selection tools to be able to mass uninstall addons.

Hanuko33:
I'll get on it once I have time and energy.
Thanks for reviewing this.

PR #5526, opened 2025-10-20
Remove releases from AppStream metadata

AsciiWolf:
If there is no _releases_ section in the upstream metainfo file, the release version and release date are auto generated by downstream packaging tools. However, if there is a releases section but is outdated, the outdated data is used instead - and that is a problem. In other words, all downstream STK 1.5 packages will now appear as 1.4 because you forgot to update the releases section. And this is sadly not the first time this happened. 1.4 had the same problem if I remember correctly. So... Let's remove the whole section to prevent this from happening again.

/cc @Alayan-stk-2

## Agreement
```

```

### DIFF
--- a/data/net.supertuxkart.SuperTuxKart.metainfo.xml
+++ b/data/net.supertuxkart.SuperTuxKart.metainfo.xml
@@ -433,17 +433,6 @@
     <content_attribute id="violence-cartoon">mild</content_attribute>
     <content_attribute id="social-chat">intense</content_attribute>
   </content_rating>
-  <releases>
-    <release version="1.4" date="2022-10-31">
-      <url>https://blog.supertuxkart.net/2022/11/supertuxkart-14-release.html</url>
-    </release>
-    <release version="1.3" date="2021-09-28">
-      <url>https://blog.supertuxkart.net/2021/09/supertuxkart-13-release.html</url>
-    </release>
-    <release version="1.2" date="2020-08-27"/>
-    <release version="1.1" date="2020-01-05"/>
-    <release version="1.0" date="2019-04-21"/>
-  </releases>
   <languages>
     <lang percentage="98">ar</lang>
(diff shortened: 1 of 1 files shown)


AsciiWolf:
If possible, please consider re-releasing 1.5 after merging this change.

Alternatively, you can add the missing 1.5 release data instead:

```
    <release version="1.5" date="2025-10-20">
      <url>https://blog.supertuxkart.net/2025/10/supertuxkart-15-release.html</url>
    </release>
```

But this won't prevent it from happening again in the future.

AsciiWolf:
Thanks. I forgot to fix it in _data/po/update_desktop_file_appdata.py_ too though and you were quicker with the merge than me with the fix. :-) I will fix it in upcoming PR.

Alayan-stk-2:
Removing is the correct  choice since there is already way too many things to update for each release. Anything that's not in a dedicated checklist risks being missed, and in general less things to update is nicer.

> If possible, please consider re-releasing 1.5 after merging this change.

I think this change matter only for places where the game is repackaged like for the flatpack or for various distro, not for the packages we release, correct?

AsciiWolf:
>I think this change matter only for places where the game is repackaged like for the flatpack or for various distro, not for the packages we release, correct?

Correct. It matters for all **downstream** packaging - this includes Flatpaks, RPMs, DEBs etc. Basically all packaging formats that use AppStream metadata and fetch the release data from there.

AsciiWolf:
#5527

Alayan-stk-2:
The SRC package (which contains code + assets) has been updated with the hole drop fix as well as those meta-data changes, so if downstream use it they'll be fine. The Source code .zip and .tar.gz can't be easily updated though. I'd have to remove the release on GitHub entirely and redo it, it seems. With all the files to upload back, that'd be quite annoying.

PR #5530, opened 2025-10-21
text added on help section for splitscreen

bricealbiol:
## Agreement
```

```


Alayan-stk-2:
I didn't comment earlier, but doing this change would break translations and it's not sure all of them would be updated. For something so minor I prefer to not touch it for 1.5.1. For later we'll have to see, the text may require deeper changes.

PR #5542, opened 2025-10-26
Fix embedded irrlicht build failures on Debian big endian archs

Vincent-C:
## Agreement
```

```

Debian's supertuxkart package has carried these big-endian-specific changes to the embedded irrlicht copy for a while now, and I'd like to forward them upstream for inclusion ([patch included in Debian package here](https://salsa.debian.org/games-team/supertuxkart/-/blob/master/debian/patches/irrlicht/arch-support.diff?ref_type=heads)). This was included to fix build failures on Debian mips/powerpc/s390x/sparc, and should not affect building on any other architecture adversely. Thanks!

Patch header has some additional context:

```
Building on non-linux architectures currently fails with unpatched
irrlicht because irrlicht tries to create Joystick support using
linux-specific headers. However there's infrastructure to disable
Joystick support, we just need to activate that on non-linux
architectures.

Additionally if built on a sparc machine irrlicht assumes wrongly it's
a solaris system. We fix this wrong assumption as our sparc builds are
all on linux.
```

### DIFF
--- a/lib/irrlicht/include/IrrCompileConfig.h
+++ b/lib/irrlicht/include/IrrCompileConfig.h
@@ -116,6 +116,10 @@
 #if !defined(_IRR_WINDOWS_API_) && !defined(_IRR_OSX_PLATFORM_) && !defined(_IRR_ANDROID_PLATFORM_) && !defined(_IRR_HAIKU_PLATFORM_)
 #ifndef _IRR_SOLARIS_PLATFORM_
 #define _IRR_LINUX_PLATFORM_
+#include <endian.h>
+ #if __BYTE_ORDER == __BIG_ENDIAN
+  #define __BIG_ENDIAN__
+ #endif
 #endif
 #define _IRR_POSIX_API_
 #endif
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
I'm quite surprised by the claim that

> Building on non-linux architectures currently fails with unpatched
> irrlicht because irrlicht tries to create Joystick support using
> linux-specific headers.

Beyond the fact that of course STK currently has working non-linux releases, if you look at the code on line 133, you have
`#undef _IRR_COMPILE_WITH_JOYSTICK_EVENTS_`

And this hasn't changed in 10+ years. https://github.com/supertuxkart/stk-code/blame/master/lib/irrlicht/include/IrrCompileConfig.h#L133

When Irrlicht code was embedded in our libs, some parts of it that were not used remained there because it was simpler that way, however setting that define to true does nothing (the game doesn't use Irrlicht for gamepad handling) except apparently creating compilation issues.

I don't want to merge this patch as-is, because it carries the idea that setting this define has some kind of use.

If anything, it may make more sense to completely remove all the code that's guarded by the _IRR_COMPILE_WITH_JOYSTICK_EVENTS_ defines.

Vincent-C:
Thanks for the feedback! I was more worried about the big endian part of the patch, but I just tried building on a s390x porterbox without the patch and it appears to build fine, so I plan on just dropping the patch altogether in Debian's supertuxkart package.

PR #5543, opened 2025-10-26
Added camera shake at high speeds

BigBuckBunny12:
## Agreement
```

```
Added a simple camera shake when the kart reaches high speeds. The intensity of the shaking effect is proportional to the speed and capped at a maximum. I included a toggle to disable the feature in the camera settings menu.

### DIFF
--- a/data/gui/dialogs/custom_camera_settings.stkgui
+++ b/data/gui/dialogs/custom_camera_settings.stkgui
@@ -71,6 +71,10 @@
                     <checkbox id="use_soccer_camera"/>
                     <spacer width="1%" height="100%" />
                     <label height="100%" text_align="left" I18N="In the ui/camera screen" text="Follow ball in soccer mode"/>
+		    <spacer width="3%" height="100%" />
+		    <checkbox id="use_camera_shake"/>
+                    <spacer width="1%" height="100%" />
+                    <label height="100%" text_align="left" I18N="In the ui/camera screen" text="Enable camera shake"/>
                 </div>
             </div>
 
--- a/src/config/user_config.hpp
+++ b/src/config/user_config.hpp
@@ -1071,6 +1071,11 @@ namespace UserConfigParams
             PARAM_DEFAULT(  BoolUserConfigParam(false, "reverse-look-use-soccer-cam",
                             "Use ball camera in soccer mode, instead of reverse") );
 
+    PARAM_PREFIX BoolUserConfigParam         m_camera_use_shake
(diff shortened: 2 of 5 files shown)


Alayan-stk-2:
While not as bad as I feared it may be, I would generally characterize the camera shake as unpleasant. Personally, I wouldn't want this. If you want you could make a video demo (using an appropriate amount of zippers, nitro and drifting) and show it on the forum and IRC to get some opinions.

Besides, there is also the motion blur option if the goal is to communicate a feeling of high speed. Its current trigger requires a zipper usage (see HandleZipper in local_player_controller.cpp), but I think it would be more worthwhile to revise its activation conditions than to introduce camera shake.

BigBuckBunny12:
Thank you for the feedback!

My initial idea was to have the camera shake occur only at maximum speed (i.e., when the speedometer is completely full), but I couldn’t find a clear way to detect that condition in the codebase.
Would you happen to know where I could check for this, so I can run some more tests?

I could also try applying the shake only when a zipper is used at top speed, making it a short, instantaneous effect rather than a continuous one.

I’ll make sure to share the different versions with the community before updating the PR.

Alayan-stk-2:
> My initial idea was to have the camera shake occur only at maximum speed (i.e., when the speedometer is completely full), but I couldn’t find a clear way to detect that condition in the codebase. Would you happen to know where I could check for this, so I can run some more tests?

There is no well-defined maximum speed. The maximum speed you can see in the speedometer is just a display limit, not something that caps how fast the kart goes (in 0.9.3 the speedometer was full at a speed of 30, in 1.x at a speed of 40, and for STK Evolution it will be at a speed of 55 - but the differences in actual maximum achievable speed are much smaller, there simply were many situations where the real speed would exceed what the speedometer can display).

The karts have a standard maximum speed, but various sort of boosts (nitro, drifting, zippers, slipstream, plunger...) provide for temporarily increased max speed. It's not really possible in practice to accumulate all the various kind of boost types at once, but having 2 or 3 different boost types active at once is not uncommon (for one, good players will nearly always have some level of drift boost active). The same "very high speed" could be achieved in various combinations of boost types.

BigBuckBunny12:
After further testing and considering the feedback received, I’ve decided to discard this PR and rethink the use of screen shake to enhance immersion. Instead of triggering it at high speeds, I’ll apply the effect when an explosion occurs near the player, which should feel more natural and less intrusive.

I will open a new PR with the revised implementation after gathering enough feedback.

Issue #5338, opened 2025-02-27
Make renaming things like controls configuraiton/username/grandprix a consistent experience


scootergrisen:
When renaming things like controls configuration, username, grandprix it would be nice if the experience is the same.

Have it like you experience in so much other software, text input prefilled with old name, maybe focused and cursor at the end.

Right now it's a mixture, username is prefilled but not focused, grandprix is empty and focused.

When renaming username the text input is high up on the screen while the buttons are at the bottom of the screen.

bricealbiol:
Could you please let me know if there is anything else left to do, or if this issue can now be considered resolved and closed? Thanks!

PR #5549, opened 2025-10-30
Replace RegisterScreen with GeneralTextFieldDialog for player renaming

bricealbiol:
Replace the full RegisterScreen with GeneralTextFieldDialog when renaming a player to match the UI used for grand prix and device configuration naming.
Issue #5338 

## Agreement
```

```
image_2025-10-30_134208351


Alayan-stk-2:
That might work, but it doesn't really simplify the renaming process and it adds additional code complexity, so I'm not quite convinced. I much prefer when the code is simplified.

PR #5557, opened 2025-11-03
Prefil textbox for options and grand prix renaming

bricealbiol:
I made sure that when the text box opens to edit a name for the grand prix and configuration, the current name is already there.

Issue #5338 
## Agreement
```

```


Alayan-stk-2:
Thank you for the PR, simple but nice QoL change.

PR #5562, opened 2025-11-10
Set focus on username textbox when opening rename screen

bricealbiol:
Set focus on username textbox when opening rename screen.
It's a small add, but I thought it would be useful for users.
Issue #5338 

## Agreement
```

```


Alayan-stk-2:
Simple and small change perhaps, but also very easy for me to test, and a clear improvement in the UI behavior. I call that a good PR. :)

PR #5566, opened 2025-11-17
Add scrollbar to server lobby chat

BcfWor:
## Agreement
```

```


Alayan-stk-2:
There are some minor things that could be improved, but the important parts (displaying the scrollbar and having it allow to move through chat history by clicking it or using the mouse wheel, retaining text history between races, etc.) are working in my testing and the code looks ok.

Issue #5505, opened 2025-09-18
Lap counter stops working around lap 130


Alayan-stk-2:
The game in its normal UI only supports up to 20 laps race, and even for endurance races that some people may want to play, 100 laps is plenty. This issue is therefore of extremely low importance.

My guess from the reported symptoms is that the lap counter is encoded using 8 bits. If the value is signed rather than unsigned, that would put the max supported number as 128.

Using 8-bits instead of 16-bits slightly reduces how much data is transmitted in networked mode, however the lap counter may not need to be sent too often and it's just 8 bits so it shouldn't matter too much.

Khooaaa:
Hi, i'm a student and for my studies i need to contribute to a project.
I wanna work on this issue, so i was wondering how to reproduce this bug? 
I'm currently trying to but i can't do it.

Alayan-stk-2:
The issue was reported on the forum here: https://forum.supertuxkart.net/thread-25.html

lokpro:
> Hi, i'm a student and for my studies i need to contribute to a project. I wanna work on this issue, so i was wondering how to reproduce this bug? I'm currently trying to but i can't do it.

1.
https://github.com/supertuxkart/stk-code/blob/172f4dd038350e3c061642dab1f6a49ff0687775/data/gui/screens/track_info.stkgui#L35
Adjust the max_value to 150 without requiring compilation.  

2. Play this "buggy" track: https://online.supertuxkart.net/addons.php?type=tracks&name=holiday4way1  Move back and forth at the finish line to add one lap.

Alayan-stk-2:
Who knew that buggy addon tracks could be useful :)

Alayan-stk-2:
Fixed with https://github.com/supertuxkart/stk-code/commit/f0b2184cea6c19730fbbeaec59669646f083aba3

PR #5593, opened 2025-12-15
Lap counter bug

Khooaaa:
I fixed the problem that caused the lap counter to block around 130 laps.

issue #5505 

## Agreement
```

```

### DIFF
--- a/src/network/peer_vote.hpp
+++ b/src/network/peer_vote.hpp
@@ -31,7 +31,7 @@ class PeerVote
 public:
     core::stringw m_player_name;
     std::string   m_track_name;
-    uint8_t       m_num_laps;
+    uint16_t       m_num_laps;
     bool          m_reverse;
 
     // ------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Kadajett:
I am guessing this isn't backwards compatible as the byte offset for network requests will be changing. I would just bump the server version, or add a compatibility negotiation so old clients would be limited but not broken. Unsure what the common practice is for this app, so feel free to ignore me!

Alayan-stk-2:
> I am guessing this isn't backwards compatible as the byte offset for network requests will be changing

Correct.

I should arguably have left a comment here earlier, but the reason I have not merged this so far is that it breaks network compatibility. For the Evolution branch that won't be an issue but for the time being I'm avoiding that in the master branch.

Alayan-stk-2:
I wasn't really into the habit of changing the target branch of a PR when I last checked this PR, but it's actually quite convenient in cases like this.

Alayan-stk-2:
Sorry it stayed open for so long, thanks for your contribution!

Kadajett:
Happy to see stuff moving forward again! Just in time, cause I just got an RTO layoff!

Issue #5555, opened 2025-11-01
Add a character limit for the username (30)


willliam9:
### Description

A username that is too long may cause display problems 

When creating an account, a character limit for the username should be set (50)

Alayan-stk-2:
The online username is already subject to a limit, the local username is not. In general, we want to leave freedom for people to do as they wish on their local installation, but in its default configuration, the game shouldn't let users shoot themselves in the foot.

Gustanator:
I did a fix some months ago in my old fork project, but in some days, I'll PR a reviewed version of this here

Alayan-stk-2:
Gusta made the character limit 30 instead of 50 in his PR, and after some thinking I think it is a better upper limit. 30 is already high enough that very few people will hit the limit, it matches the online username limit, and it makes it easier to prevent overflows without having to systematically cut part of the username.

PR #5594, opened 2025-12-17
character limit for local username

Khooaaa:
Limit local username to 50 characters

issue #5555 

## Agreement
```

```

### DIFF
--- a/src/config/stk_config.cpp
+++ b/src/config/stk_config.cpp
@@ -265,6 +265,8 @@ void STKConfig::init_defaults()
     m_banana_item_return_ticks    = -100;
     m_bubblegum_item_return_ticks = -100;
 
+    m_max_username_length = 50;
+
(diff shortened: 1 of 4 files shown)


willliam9:
The value 50 should be in a configuration value

Alayan-stk-2 (changes requested):
Thank you for your Pull Requests!

The extra + symbols need to be removed. It seems like you copied from a patch file and forgot to check if the code actually compiles.

Wil's suggestion of making the limit configurable in something like stk config also makes some sense, although it's more of a nice to have.

Alayan-stk-2:
The config handling in this PR didn't fit the norms of the codebase, so there were still some significant problems left.

Closing since https://github.com/supertuxkart/stk-code/pull/5787 was merged instead. I gave co-author credit to @Khooaaa in the commit that was merged, though.

PR #5600, opened 2025-12-28
Fix inverted gravity on track edges

### LINKED ISSUE #5584: Inverted Gravity in Ravenbridge Mansion
OS: ArchLinux x84_64
Version: 1.5 in Flathub

Story Mode
Ravenbridge Mansion

> the_edge
> I took it after that.

I was keeping high speed on the edge before that I fell into it.

My car was inverted because of the edge.

And then I found that my gravity was also inverted.

AIs' cars weren't inverted.

> inverted0
>
> inverted1
> I took it when I was inverted.

Is it a feature?

You'd better fix it if it's not a feature.

Thanks!

Kadajett:
## Summary

Fixes #5584

When driving at high speed over track edges on Ravenbridge Mansion, the kart could become inverted with gravity also inverting. This occurred because edge faces with `hasGravity` textures have near-horizontal normals that were incorrectly being used to set gravity direction.

## Root Cause

1. Edge geometry on the track has textures with `hasGravity=true`
2. Edge faces have normals pointing sideways/diagonally instead of upward
3. When the kart's wheel raycasts hit these edge faces, gravity was set to align with the unusual normal
4. When the kart lost ground contact (fell into hole), flying mode tried to align the kart with this incorrect gravity direction
5. The torque impulse flipped the kart to match the wrong gravity vector

## Fix

Two-layer defense:

1. **Surface normal validation** (`kart.cpp`): Only apply gravity from `hasGravity` surfaces when the normal has sufficient vertical component (`|Y| > 0.3`). Edge faces with near-horizontal normals are rejected.

2. **Grounded gravity memory** (`btKart.cpp`): Store the last known gravity direction when wheels were on ground. Flying mode now uses this stored value instead of current gravity, preventing incorrect orientation when gravity was set by an edge surface just before becoming airborne.

## Test Plan

- [x] Build compiles successfully on BalanceSTK2 branch
- [x] Follows STK coding style guidelines

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -1982,7 +1982,13 @@ void Kart::update(int ticks)
         if (material && material->hasGravity())
         {
             Vec3 normal = m_terrain_info->getNormal();
-            gravity = normal * -g;
+            // Only apply surface gravity if normal has sufficient vertical
+            // component (|Y| > 0.3). Edge faces have near-horizontal normals
+            // that would incorrectly flip gravity direction.
+            if (fabsf(normal.getY()) > 0.3f)
+            {
+                gravity = normal * -g;
+            }
         }
 
         body->setGravity(gravity);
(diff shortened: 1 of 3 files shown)


Kadajett:
Closing - mistakenly targeted BalanceSTK2 instead of master. Reopening with correct base branch.

PR #5605, opened 2025-12-29
Fix inverted gravity on track edges

### LINKED ISSUE #5584: Inverted Gravity in Ravenbridge Mansion

Kadajett:
## Summary

Fixes #5584

When driving at high speed over track edges on Ravenbridge Mansion, the kart could become inverted with gravity also inverting. This occurred because edge faces with `hasGravity` textures have near-horizontal normals that were incorrectly being used to set gravity direction.

## Root Cause

1. Edge geometry on the track has textures with `hasGravity=true`
2. Edge faces have normals pointing sideways/diagonally instead of upward
3. When the kart's wheel raycasts hit these edge faces, gravity was set to align with the unusual normal
4. When the kart lost ground contact (fell into hole), flying mode tried to align the kart with this incorrect gravity direction
5. The torque impulse flipped the kart to match the wrong gravity vector

## Fix

Two-layer defense:

1. **Surface normal validation** (`kart.cpp`): Only apply gravity from `hasGravity` surfaces when the normal has sufficient vertical component (`|Y| > 0.3`). Edge faces with near-horizontal normals are rejected.

2. **Grounded gravity memory** (`btKart.cpp`): Store the last known gravity direction when wheels were on ground. Flying mode now uses this stored value instead of current gravity, preventing incorrect orientation when gravity was set by an edge surface just before becoming airborne.

## Test Plan

- [x] Build compiles successfully
- [x] Follows STK coding style guidelines

### DIFF
--- a/src/karts/kart.cpp
+++ b/src/karts/kart.cpp
@@ -1982,7 +1982,13 @@ void Kart::update(int ticks)
         if (material && material->hasGravity())
         {
             Vec3 normal = m_terrain_info->getNormal();
-            gravity = normal * -g;
+            // Only apply surface gravity if normal has sufficient vertical
+            // component (|Y| > 0.3). Edge faces have near-horizontal normals
+            // that would incorrectly flip gravity direction.
+            if (fabsf(normal.getY()) > 0.3f)
+            {
+                gravity = normal * -g;
+            }
         }
 
         body->setGravity(gravity);
--- a/src/physics/btKart.cpp
+++ b/src/physics/btKart.cpp
@@ -123,6 +123,10 @@ void btKart::reset()
     m_max_speed                  = -1.0f;
     m_min_speed                  = 0.0f;
     m_leaning_right              = true;
+    // Initialize to default downward gravity - will be updated when grounded
+    m_last_grounded_gravity      = btVector3(0, -(Track::getCurrentTrack()
+                                              ? Track::getCurrentTrack()->getGravity()
+                                              : 9.80665f), 0);
 
     // Set the brakes so that karts don't slide downhill
     setAllBrakes(20.0f);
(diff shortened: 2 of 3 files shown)


Kadajett:
@Alayan-stk-2 Obv feel free to close this one too. I haven't had time to write code here or test much this week, so this could very well be another situation where I didn't have the in-game experience to test, so my couple hours of in-game testing were false positives. 

I appreciate you taking the time to review new contributors code! Its basically a full time job, especially these days lol

Alayan-stk-2:
Closing for a few reasons:
- Using the last "on-ground" gravity for orientating a flying kart is a very bad idea. While it would certainly fix the specific issue present in Ravenbridge, it would also mess things up in all tracks where a jump is followed by a change of gravity (for example, going from a wall-drive section into a a normal section through a jump).
- The surface normal validation concept also appears unsound.
- The contributor agreement is missing
- I have strong suspicion that the PR summary was at least partially AI-generated, and I suppose the code too

Kadajett:
I honestly don't remember how much I used AI to assist on this one, but the PR description definitely was written with AI! 

I have had my own ups and downs trying to figure out where AI fits in after 15 years of hand coding, so I apologize if that broke any of the rules in this project!

I also agree that it was a fix that I figured would work mathematically, but I didn't have the full scope of the game at that time to make the call.

I do hope you guys find a solution for this one though!

Alayan-stk-2:
We plan on fixing it at the track-design level for STK's standard tracks. Magnetic texture should not be applied willy-nilly, the code can't just fix everything if the track tells the game karts should stick to some surface they shouldn't actually stick to.

Issue #5468, opened 2025-07-03
Tooltip for options disabled in the options accessed during a race pause


Alayan-stk-2:
It's possible for some users to check some options during a race pause, and to see them disabled without making the connection with the race status. A tooltip explaining that the option is not available during a race pause would make it much clearer that the users need to access options from the main menu to change those.

bricealbiol:
<img width="1065" height="654" alt="Image" src="https://github.com/user-attachments/assets/44aa6a64-cc00-492f-b115-3fc258bdc51a" />

When you talk about non-modifiable options, are you referring, for example, to the render resolution, which cannot be changed during the race?

Alayan-stk-2:
Yes, that's correct.

PR #5618, opened 2026-01-12
add tooltips for disabled options during a race

bricealbiol:
Issue #5468 
I added a tooltip for all parameters where the boolean “in_game” was true. I don't think I forgot any parameters, but it would still be worth checking. I hope you like it.

## Agreement
```

```


Alayan-stk-2:
Thank you for your code!

After testing, I see two main issues with this PR:
- The tooltips persist outside of race pause
- Repeating the same string over and over is generally to be avoided, as it makes it more annoying to update it if needed.

For simplicity's sake, I'm going to merge the PR with these known defects and will fix them myself.

Issue #5348, opened 2025-02-28
Actions without key bindings are considered in conflict


scootergrisen:
Go to Preferences > Controls
Double-click on a keyboard configuration like Keyboard 0.
Remove two key bindings so they say "[none]".
They turn red and you get the message:

    * A red item means a conflict in the current configuration

So actions without key binding should not be considered in conflict with other actions without key binding.

Alayan-stk-2:
The message is incorrect, but a game action having no binding at all is not good either and it would actually make sense to highlight it as wrong (even if there is only one unassigned control)

PR #5636, opened 2026-02-02
Actions without key binding

Khooaaa:
Mark unassigned actions in blue #5348 

## Agreement
```

```
image


Alayan-stk-2:
Thank you for the PR.

I switched markItemBlue to emphasisItem, because markItemBlue is already used when there is a conflict between two different configurations (this applies mostly to keyboard configurations), and there is even a text message telling the player that's the meaning of the blue color.

The emphasis color is currently yellow for the coal theme and a slightly darker blue for other themes. The emphasis color is for now only used here and in the achievement list. The yellow color of the coal theme works well, I'll update the emphasis color used by other themes to distinguish it.

PR #5637, opened 2026-02-02
Just use radius for activate change camera_tv

LLS-lopos:
## Agreement
```

```
I follow the principle of other types of cameras:
parameter "start" for the radius


LLS-lopos:
[icy_soccer_field.zip](https://github.com/user-attachments/files/25022433/icy_soccer_field.zip)
[xr-4r3n4.zip](https://github.com/user-attachments/files/25022435/xr-4r3n4.zip)
[oasis.zip](https://github.com/user-attachments/files/25022439/oasis.zip)
[soccer_field.zip](https://github.com/user-attachments/files/25022438/soccer_field.zip)
[hole_drop.zip](https://github.com/user-attachments/files/25022451/hole_drop.zip)
[lasdunassoccer.zip](https://github.com/user-attachments/files/25022457/lasdunassoccer.zip)

Alayan-stk-2:
I think I don't quite understand the reasoning behind the change here. The radius approach may make sense for end-race cameras, but the needs for the soccer cameras are quite different.

This won't fix the issue of picking a camera that's looking "the wrong way", is it? Or picking a camera that's too close to the ball?

By looking the wrong way, I mean for example that the ball is moving towards a goal and the game picks a camera that looks from the goal towards the outside instead of a camera looking from the middle of the field towards the center.

And if I understand correctly it also requires an update to the fields adding custom radius values for all of the soccer cameras, as you've linked new soccer field zips.

LLS-lopos:
It now checks which camera to use based on the viewing angle of the active camera.
If the camera angle is less than the defined value (if too close to the active camera), it switches to the active camera.
@Alayan-stk-2

LLS-lopos:
Here, the view of the ball is taken into account to ensure it is always at the same distance, and therefore the camera zooms in and out relative to the ball. #5788 @Alayan-stk-2

Issue #5619, opened 2026-01-12
[GUI] Text on the team selection screen for football mode is misaligned vertically


pedropaulosuzuki:
### Description

When you go to the 'Football' mode selection screen, the text on the top is not centered vertically, being located a bit higher than intended.

### Steps to reproduce

1. Launch SuperTuxKart
1. Select 'Singleplayer'
1. Select any character
1. Select 'Football' mode
1. Look at the text on the top of the screen. "Use left/right to choose your team and press fire" is not centered vertically.

<img width="1911" height="1076" alt="Image" src="https://github.com/user-attachments/assets/60a7b7f2-e11f-41f9-907c-23f78759d9b9" />

### Configuration

SuperTuxKart 1.5 (Also tested on latest git).

ewingwm:
I am looking into this issue and the issue seems to originate from the Irrlicht Engine that is used by STK.

pedropaulosuzuki:
Something similar happens on the character and map selection screens (to a lesser degree), but on the buttons/tabs at the bottom:

### Map selection

<img width="1920" height="103" alt="Image" src="https://github.com/user-attachments/assets/9a109fc1-ac5c-44db-9d66-e1263f2edc92" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e170b46e-5b7a-4b4b-b3cb-55d6574c70ec" />

### Character selection

<img width="1920" height="181" alt="Image" src="https://github.com/user-attachments/assets/6fb1a3c5-0ec9-436c-b5c4-28ea32204650" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/31718538-c24b-44ce-a77b-7bc7d8084650" />

pedropaulosuzuki:
Also in the graphics settings:

<img width="1523" height="962" alt="Image" src="https://github.com/user-attachments/assets/0f3a071d-5337-4b65-9dea-9f37324531d0" />

Text on the right of checkboxes always aligns to top, instead of middle

PR #5638, opened 2026-02-03
Correct text vertical alignment in Football mode and UI

bricealbiol:
Issue #5619 
Previously, using only `getHeightPerLine()` for centering calculations caused text to appear offset because it included line padding. By introducing `getHeightPerGlyph()`, we can now center text based on the visual bounds of the characters.

- **IGUIFont.h & CGUIFont.h**: Added `getHeightPerGlyph()` to the font interface to retrieve the actual height of characters without line spacing.
- **GlyphLayout.h**: Updated `getGlyphLayoutsDimension` to accept `height_per_glyph` as a parameter for more accurate dimension calculations.
- **CGUIButton.cpp**: Improved the centering logic in `draw()` to use the new glyph height, ensuring text is vertically centered within buttons.
- **CGUIStaticText.cpp**: Adjusted height and width calculations to account for the difference between line height and glyph height.

- **FontWithFace**: Added `m_glyph_max_above` and `m_glyph_max_below` to track precise font metrics (ascender/descender) via FreeType.
- **ScalableFont**: Implemented `getHeightPerGlyph()` for STK's scalable font system, including support for borders and shadows.
- **GUI Components**: Updated `CGUIEditBox`, `MessageQueue`, `NetworkingLobby`, and `RaceResultGUI` to use the new dimension logic.

### Technical Background: Bearing vs. Height
To fix the vertical alignment, I've distinguished between the line height and the visual glyph metrics. As explained in the [FreeType documentation](https://freetype.org/freetype2/docs/glyphs/glyphs-3.html):

- **Line Height (Height)**: The distance between two consecutive baselines. It often includes internal leading/padding, which causes "centering" logic to look offset.
- **Glyph Bearing/Bounding Box**: This represents the actual ink of the characters. 

By implementing `getHeightPerGlyph()` (based on the max ascender and descender metrics), we ensure that UI elements like buttons and the Football score are centered based on the **visible** part of the text rather than the theoretical line height.

## Agreement
```

```

### DIFF
--- a/lib/irrlicht/include/GlyphLayout.h
+++ b/lib/irrlicht/include/GlyphLayout.h
@@ -105,8 +105,8 @@ inline void eraseTopLargerThan(std::vector<GlyphLayout>& gls,
 }
 
 inline core::dimension2d<u32> getGlyphLayoutsDimension(
-    const std::vector<GlyphLayout>& gls, s32 height_per_line, f32 inverse_shaping,
-    f32 scale, s32 cluster = -1)
+    const std::vector<GlyphLayout>& gls, s32 height_per_line, s32 height_per_glyph, 
+    f32 inverse_shaping, f32 scale, s32 cluster = -1)
 {
     core::dimension2d<f32> dim(0.0f, 0.0f);
     core::dimension2d<f32> this_line(0.0f, (f32)height_per_line);
@@ -148,7 +148,7 @@ inline core::dimension2d<u32> getGlyphLayoutsDimension(
             break;
     }
 
-    dim.Height += this_line.Height;
+    dim.Height += height_per_glyph;
     if (dim.Width < this_line.Width)
         dim.Width = this_line.Width;
 
@@ -328,7 +328,7 @@ inline bool getDrawOffset(const core::rect<s32>& position, bool hcenter,
     if (hcenter || vcenter || clip)
     {
         text_dimension = gui::getGlyphLayoutsDimension(
-            gls, next_line_height, inverse_shaping, scale);
+            gls, next_line_height, next_line_height, inverse_shaping, scale, -1);
 
         if (hcenter)
         {
--- a/lib/irrlicht/include/IGUIFont.h
+++ b/lib/irrlicht/include/IGUIFont.h
@@ -112,6 +112,8 @@ public:
 
 	//! Returns the height per line (including padding between 2 lines)
 	virtual s32 getHeightPerLine() const = 0;
+	virtual s32 getHeightPerGlyph() const = 0;
+
 
 	//! Define which characters should not be drawn by the font.
 	/** For example " " would not draw any space which is usually blank in
--- a/lib/irrlicht/source/Irrlicht/CGUIButton.cpp
+++ b/lib/irrlicht/source/Irrlicht/CGUIButton.cpp
@@ -333,9 +333,15 @@ void CGUIButton::draw()
 		}
 
 		if (font)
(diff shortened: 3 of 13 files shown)


Gustanator:
@bricealbiol, I tested your PR, but I cannot see difference between your code and the actual code in the soccer setup screen. You are finishing this yet?

bricealbiol:
Hello @GustaLOLE,
Without my code in the team selection screen when choosing soccer mode, the text “Use left/right to choose your team and press fire” is not centered in the div as shown in the attached screen in the issue, and with this code it is.

Gustanator:
I see you don't edited the soccer setup file, why? And send us images with this screen, please.

Gustanator:
@bricealbiol Can you send this PR to my fork, pls?

bricealbiol:
image

The image above represents with the code, the text “Use left/right to choose your team and press fire” is centered.

The reason I didn't edit the soccer setup file directly is that the misalignment wasn't caused by incorrect coordinates in that specific file, but by how the engine calculated text dimensions.

bricealbiol:
How can i send this PR in your fork pls ?
I rarely use GitHub, sorry...

Gustanator:
Send a PR in (https://github.com/GustaLOLE/stk-code) as the same way of you send here

pedropaulosuzuki:
> @bricealbiol Can you send this PR to my fork, pls?

@GustaLOLE You can just pull the commits once this is merged upstream. Better than having contributors do everything twice.

Gustanator:
Ehm, Alayan will take time to put this maybe

pedropaulosuzuki:
> image
> 
> The image above represents with the code, the text “Use left/right to choose your team and press fire” is centered.
> 
> The reason I didn't edit the soccer setup file directly is that the misalignment wasn't caused by incorrect coordinates in that specific file, but by how the engine calculated text dimensions.

@bricealbiol Thanks for taking a look into this!

It seems the Continue button on your screenshot is also vertically misaligned, maybe this could also be fixed in the same patch?

Also, there is a small problem when aligning text based on the actual rendered height: anything containing characters that draw below the "text bottom line" (there's probably an official term for that) will make the text go up. Think any of these characters: "JQÇ" and "çgjpqy"). Also, if no capital letters are present (or taller characters, like "bdfhijklt") will bring the text down. The ideal way to align text vertically usually uses the height of the "capital X" character for that font at that size, in order to correctly align text without the issues presented in this paragraph.

Edit: by the code, it seems it samples the character 'A', so all good in that regard.

pedropaulosuzuki:
@bricealbiol it seems the "Continue" button was correctly aligned before the patch, so it seems to be a regression.

Before patch:
image

bricealbiol:
Yes, actually, without the modifications to the file lib/irrlicht/source/Irrlicht/CGUIButton.cpp, the text on the “continue” button is completely misaligned. I saw it in the middle, but it doesn't actually appear to be centered.

Alayan-stk-2:
I applied your PR to test it, and while it clearly fixes the soccer screen text, it seems to mess up alignment for all kind of buttons. It also seems to have some other side effects, you can observe that the spinners become narrower with the patch (presumably text height is used as a parameter somewhere ?).

Master:
VerticalTextAlignMaster


Patch:
VerticalTextAlignPatch

So unfortunately I cannot merge this PR as-is.

Alayan-stk-2:
Feel free to open a new PR if you can fix the issues I pointed above.

PR #5642, opened 2026-02-08
Fix kart_charteristics.xml parser generator marker tag detection regex

Nomagno:
A few simple typos and/or code requirements desync have kept this from working properly for a long while.

## Agreement
```

```


Nomagno:
Whoops, turns out the issue was with escape sequences in the python script regex, not just the marker tags. Fixed now

Alayan-stk-2:
So this commit mostly changes the order things are in to match what's generated by the script?

Nomagno:
Yes. Category names can't contain dashes so I had to change Nitro-Hack to NitroHack among other things, and from there I just regenerated everything to keep consistency. Regenerating everything also gets the xml_characteristics.cpp to a consistent format. If merged, commits related to new kart characteristics after this would be back to being small and would just need to follow the simple procedure detailed in https://forum.supertuxkart.net/thread-244.html
No manual editing required

Issue #5589, opened 2025-12-08
In splitscreen kart selection, the kart stats tooltip can overflow


mPapasavvas:
### Description

In splitscreen multiplayer mode, after you choose a kart, the labels on the icons for Mass, Maximum Speed, Acceleration and Nitro Efficiency when you hover the mouse over them, are out of frame, on both fullscreen and windowed. mp4 included

https://github.com/user-attachments/assets/0269a803-2009-4bb1-87fa-0986127e9a93

### Steps to reproduce

1. Launch game
2. Navigate to Splitscreen Multiplayer mode
3. Press Select button and choose a kart
4. Hover your mouse on the kart specific icons of the stats
5. The labels of the icons are out of frame

### Configuration

SuperTuxKart version 1.5
Installation using apt
OS Kali Linux
CPU AMD Ryzen 7 5800H
GPU GeForce RTX 3060 Mobile

### Error message

```
[Paste error message here]
```

### Log File (stdout.log)

```
[Paste contents of stdout.log here]
```

### Input Configuration (input.xml)

```
[Paste contents of input.xml here]
```

Alayan-stk-2:
I remember having noticed this issue previously, but apparently there was no existing ticket about this.

Thanks for the report, but please try to use more descriptive titles!

Alayan-stk-2:
See also https://github.com/supertuxkart/stk-code/commit/1e794cee07c58d1e8dec782395c5fbcdf0c71301

PR #5644, opened 2026-02-09
Fix tooltip for horizontal overflow

bricealbiol:
Fix the issue #5589

### Changes
- **Horizontal Clamping:** Added logic to detect if the tooltip exceeds the screen width (left or right).
- **Auto-adjustment:** If the tooltip goes out of bounds, its position is shifted to stay visible within the screen.
- **Safety Margin:** Added a 10px margin so the text doesn't touch the screen edges directly.

image

## Agreement
```

```


Alayan-stk-2:
Thank you for the PR!

The fixed margin of 10 isn't ideal because at higher screen sizes, tooltip decoration overflows even if the text doesn't, but that'll be very simple to adjust.

Issue #5070, opened 2024-05-04
Completely exclude GUI files from SERVER_ONLY builds


Alayan-stk-2:
The GUI code files are peppered with define guards excluding part of them from SERVER_ONLY builds. But SERVER_ONLY builds are meant to be used through the command line, so no GUI should be needed at all.

This just adds unnecessary complication to the code base.

Alayan-stk-2:
Some work done with https://github.com/supertuxkart/stk-code/commit/b41c0d53a6ee0b38756c3e1b016158aaaf6e09fb but a lot still to be done.

Issue #3751, opened 2019-02-13
Copy & Paste on Wayland is broken


rusty-snake:
## Description
If I try to paste somethings (e.g. passwords for OnlineMultiPlayer) it doesn't work with wayland but under X11 it's work. Primary clipboard never work.

## Steps to reproduce

Start STK under wayland and try to paste, then start STK under X11 and try to paste.
## Configuration

STK release version: lastest git
STK Source: self compile
System: Fedora29

deveee:
copy & paste is not implemented yet.

Benau:
![](https://cdn.lihkg.com/assets/faces/mouse/smile.gif) with SDL2 being used now can this be ![](https://cdn.lihkg.com/assets/faces/pig/cut.gif)?

Alayan-stk-2:
As confirmed in PR #5646, copy & paste on Wayland works since SDL2 was introduced in STK.

PR #5646, opened 2026-02-10
Properly handle Unicode clipboard data on Windows.

nutzboi:
## Bug
Currently, copying and pasting text with non-ASCII characters to and from STK on Windows is **completely broken**.
This is because
1. STK ignores the SDL device when on Windows and OSX, and
2. Unicode data in Windows clipboards are encoded in UTF-16 LE while STK attempts to read it as UTF-8.

Also, a slight memory leak seems to occur on Linux systems as memory allocated by [SDL_GetClipboardText](https://wiki.libsdl.org/SDL2/SDL_GetClipboardText#:~:text=which%20must%20be%20freed%20with%20SDL%5Ffree%28%29%2E) is currently not freed.

# Fix
This PR fixes it by always using the SDL2 functions for handling clipboards when available. This also eliminates the need for the overloading clipboard functions with wchar_t types. Memory allocated by the ``SDL_GetClipboardText`` function is also properly freed now that all pointers returned by ``COSOperator::getTextFromClipboard()`` are compatible with SDL_free();

## Notes 
The ``MacOSX/OSXClipboard.h`` header is no longer in use in the codebase after this PR.

## Agreement
```

```

### DIFF
--- a/lib/irrlicht/source/Irrlicht/CGUIEditBox.cpp
+++ b/lib/irrlicht/source/Irrlicht/CGUIEditBox.cpp
@@ -12,7 +12,10 @@
 #include "rect.h"
 #include "os.h"
 #include "Keycodes.h"
+
+#if !defined(SERVER_ONLY) && defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
 #include "SDL.h"
+#endif
 
 /*
 	todo:
@@ -255,6 +258,10 @@ bool CGUIEditBox::OnEvent(const SEvent& event)
 
 bool CGUIEditBox::processKey(const SEvent& event)
 {
+#ifdef SERVER_ONLY
+    return false;
+#else
+
 	if (!event.KeyInput.PressedDown)
 		return false;
 
@@ -713,6 +720,7 @@ bool CGUIEditBox::processKey(const SEvent& event)
 	}
 
 	return true;
+#endif
 }
 
 


nutzboi:
Also, https://github.com/supertuxkart/stk-code/issues/3751 should have already been fixed with SDL2 like Benau said. Can a linux Wayland user confirm?
**Update**: A Wayland gamer has confirmed to me that clipboards work within gamescope sessions.

nutzboi:
By the way, while rewriting this for the zillionth time I thought I'd see how other versions of irrlicht handle it.
I found that Luanti's implementation of the [COSOperator](https://github.com/luanti-org/luanti/blob/master/irr/src/COSOperator.cpp) class does the same thing as this code BUT makes the class destructor function handle the memory management of the getTextFromClipboard after its use and frees and recycles the same pointer defined as a class variable on new copy attempts, eliminating the need for calling SDL_free every time the function is used. 
I believe using that implementation should fix both the improper Unicode support on Windows and the memory leak on Linux while remaining correctly compatible with existing code ***if*** the ``Operator`` defined in CGUIEditBox is declared as an object instead of a pointer so that the constructor and destructor for it are properly called.

Alayan-stk-2:
CI Failed on the linux builds because the header in

`#include <SDL_stdinc.h>`

wasn't found in CI.

nutzboi:
> CI Failed on the linux builds because the header in
> 
> `#include <SDL_stdinc.h>`
> 
> wasn't found in CI.

it wasn't found in CI on the job where SERVER_ONLY is ON. I'm not sure why it attempts to compile and link the code under ``#ifdef _IRR_COMPILE_WITH_GUI_`` in ``lib\irrlicht\source\Irrlicht\CGUIEditBox.cpp`` when SERVER_ONLY is on. does the server version have a GUI? 
also, as far as i understand, the SDL_free function is pretty much an alias of C's free function, though i'm not sure if i should  use that instead...

Alayan-stk-2:
> it wasn't found in CI on the job where SERVER_ONLY is ON. I'm not sure why it attempts to compile and link the code under `#ifdef _IRR_COMPILE_WITH_GUI_` in `lib\irrlicht\source\Irrlicht\CGUIEditBox.cpp` when SERVER_ONLY is on. does the server version have a GUI? also, as far as i understand, the SDL_free function is pretty much an alias of C's free function, though i'm not sure if i should use that instead...

A lot of GUI files are currently compiled into the SERVER_ONLY version although they are not used.

The ultimate goal in this regard is to exclude the entirety of the GUI files (guiengine, state screens...) from SERVER_ONLY versions. See this ticket: https://github.com/supertuxkart/stk-code/issues/5070

But there is a lot to do, and when you guard a class you break all the places that depend on it so you need to add new guards there. It's also annoying when merging into the STK Evolution branch, so it probably makes most sense to fully resolve it once 1.5.1 is out.

For this PR, just add the minimal guards for SERVER_ONLY versions to keep compiling and working.

Besides better Unicode handling, this PR also represents a nice code simplification, so I wouldn't want to delay its merging unnecessarily.

PR #5648, opened 2026-02-17
Fix tooltip overflow: Added dynamic margins and vertical flipping

bricealbiol:
### Description
This PR addresses the tooltip overflow issues described in #5589.
I took the opportunity to do the vertical overflow as well.

#### Key Changes
* **Dynamic Margin:** Following the maintainer's feedback, I replaced the fixed margin with a dynamic one using `params.m_right_border`. This ensures that the tooltip decoration (borders) remains inside the screen even at high resolutions.
* **Vertical Flip & Clamping:**
    * Added logic to handle the bottom of the screen. If a tooltip would overflow vertically, it now "flips" upwards to stay visible.
    * Included a safety offset (`- 20`) in the vertical clamping to prevent the tooltip from overlapping with the mouse cursor, ensuring better readability and interaction.

#### How to test
To verify the vertical overflow fix (especially for elements at the very bottom of the screen), you can temporarily replace the code in `src/states_screens/kart_selection.cpp` (line 342) with the following snippet:

```cpp
for (int n=0; n<group_amount; n++)
{
    if (groups[n] == "standard") { // Fix capitalization (#4622)
        tabs->addTextChild( _("Standard") , groups[n]);
        Widget* standard_tab = tabs->findWidgetNamed("standard");
        if (standard_tab != NULL)
            standard_tab->setTooltip(_("I have to be above !"));
    }
    else { // Try to translate group names
        tabs->addTextChild( _(groups[n].c_str()) , groups[n]);
    }
} // for n<group_amount
```

This displays a tooltip at the bottom of the screen for testing the code.

## Agreement

### DIFF
--- a/src/guiengine/skin.cpp
+++ b/src/guiengine/skin.cpp
@@ -2446,7 +2446,9 @@ void Skin::drawTooltip(Widget* widget, bool atMouse)
     
     core::position2di pos(widget->m_x + 15, widget->m_y + widget->m_h);
     const core::dimension2d<u32> screen_size = irr_driver->getActualScreenSize();
-    const int margin = 10; // Space from screen edges so tooltip doesn't get cut off
+    const BoxRenderParams& params = SkinConfig::m_render_params["tooltip::neutral"];
+    const int margin = params.m_right_border; // Space from screen edges so tooltip doesn't get cut off
+    
 
     if (atMouse)
     {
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
The margin issue isn't really fixed. m_right_border and m_left_border are fixed values. Looking further into it:
- There is the hborder_out_portion parameter that explains why the cartoon theme was overflowing.
- On small screen sizes when the tooltips gets tiny, the borders get below that set in stkskin.xml. So it starts creating a bigger offset between the screen border and the start of the tooltip. But these values act as a maximum.

As for vertical flipping, it works well. However, using the horizontal border margins to evaluate the vertical margins isn't quite correct. For the classic theme, they happen to be the same, but that's not guaranteed, and for the cartoon theme if we start calculating the margin by accounting for the out_portion parameter we land on different values.

I'll merge the PR and add a commit afterwards to adjust this.

PR #5655, opened 2026-02-27
initializes some variables as 0 to fix gcc Wuninitialized warnings

virophagesp:
## Agreement
```

```

this may or may not prevent undefined behavior, probably not


virophagesp:
found bug with items having a glowing outline, I'll reopen after fixing it

virophagesp:
turns out my changes didn't cause it

virophagesp:
it's ready

Alayan-stk-2:
It probably doesn't matter very much either way in this specific case, but it can't hurt I suppose. It's merged now so you won't get the compiler warnings anymore.

PR #5660, opened 2026-03-05
In local sessions, pause the race when the window leaves focus.

Big-Blue-Gnu:
If you were ever in the middle of a game and you accidentally alt-tabbed out or an application suddenly popped up to steal the focus, chances are you have lost some progress by the time you returned in haste. This change adds a mechanism intended for non-networked sessions. The mechanism, which is triggered when the player is in a race, opens the pause menu and has the same effect as pressing <kbd>Esc</kbd> on the keyboard. I only tested the feature on Linux Mint, on which I report no regressions. Non-Linux versions may need testing.

## Agreement
```

```

### DIFF
--- a/lib/irrlicht/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/lib/irrlicht/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -987,10 +987,9 @@ bool CIrrDeviceSDL::run()
 				else if (SDL_event.window.event == SDL_WINDOWEVENT_FOCUS_LOST)
 				{
 					// During a local session, pause the race when the window loses focus.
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Thank you for your PR!

I think that on the balance of things, this is going to be more useful than not, helping like you said when the focus is lost accidentally.

However, the annoying thing with this is that it makes it impossible to just let the race running on purpose if you want to focus to another window. That's something that can sometimes be convenient, mostly when testing some elements of the game.

So I'd like a new option in the player profile's config.xml that would be checked to determine if the pause should be activated or not upon loss of window focus. A GUI setting is not necessary for this.

PR #5662, opened 2026-03-05
Simplify color initialization in SunNode constructor

Sulca2099:
## Agreement
```

```


Alayan-stk-2:
Using an initialization list to assign values in an already existing array is not correct C++. Not surprisingly, it doesn't compile for me.

Sulca2099:
Ok thanks, I really needed that feedback

PR #5665, opened 2026-03-12
Re-weight random track probability based on highscores

Al2Klimov:
Tracks with more highscore entries are less likely to be selected when using the random track button. Uses the Efraimidis-Spirakis weighted shuffle algorithm with weight = 1/(1+scores) per track.

Tested and I need much more re-tries to randomly get a track I played often.

This is awesome when one installs a bunch of new tracks from the addons!

## Agreement
```

```

### DIFF
--- a/src/race/highscore_manager.hpp
+++ b/src/race/highscore_manager.hpp
@@ -78,6 +78,19 @@ public:
     // ------------------------------------------------------------------------
     int highscoresSize()                { return m_all_scores.size(); }
     // ------------------------------------------------------------------------
+    /** Returns the total number of highscore entries for a given track,
+     *  across all configurations (difficulty, karts, laps, reverse). */
+    int getNumberOfHighscoreEntries(const std::string &trackName) const
+    {
+        int count = 0;
+        for (const auto& hs : m_all_scores)
+        {
+            if (hs->m_track == trackName)
+                count += hs->getNumberEntries();
+        }
+        return count;
+    }
+    // ------------------------------------------------------------------------
     void sortHighscores(bool reverse)
     {
         (reverse ? std::stable_sort(m_all_scores.rbegin(),
--- a/src/states_screens/tracks_and_gp_screen.cpp
+++ b/src/states_screens/tracks_and_gp_screen.cpp
@@ -29,6 +29,7 @@
 #include "io/file_manager.hpp"
 #include "race/grand_prix_data.hpp"
 #include "race/grand_prix_manager.hpp"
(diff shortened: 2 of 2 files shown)


Alayan-stk-2:
If I get the purpose of this PR right, it's to push tracks that haven't been played much by the player more often when using the random track button?

On the surface this might seem neat, but I see some big issues with it:
- You get different highscores depending on the number of laps and karts, as well as normal race or TT. I haven't compiled the code specifically to test what kind of values get out, but considering the following snippet it seems safe to assume this indeed counts everything.  Which means that imbalances in probability could be very severe and long lasting.
```
        double weight = 1.0 / (1.0 + scores);
        if (weight < 1e-10) weight = 1e-10;
```
- Tracks that have been played across less combinations of lap number, kart number and mode are likely tracks that a given player doesn't like as much. Having the random button push more tracks that the player likes less is not good.
- This goes against player expectations. A random button is expected to be either full random or perhaps more of a shuffle that avoids repeating options too often, not some weighted randomness like this. There are plenty of situations where players use the random track button without wanting to exclude tracks they have played many times.

Instead of the crude highscore method, the game directly tracks the number of time each track has been played for achievements, which would avoid some pitfalls of this implementation.

But some others are just inherent to the very concept, it's just not a good idea to alter the random button that way.

What you seem to want is more of a "random rarely played track" button, which would then make more sense as a separate button (with a nicer name). I'm not sold on dedicating GUI space for that, but I would invite you to post on the forum about your idea and try to make the case for it and see if other people would be interested.

Al2Klimov:
> "random rarely played track" button, which would then make more sense as a separate button

What about a checkbox in the settings instead which configures the behavior of the existing button?

Alayan-stk-2:
> What about a checkbox in the settings instead which configures the behavior of the existing button?

This could be confusing in its own way, but it would also clutter settings.

It's great to give players the ability to adjust the game to best suit their needs and wants, but if you have ever seen those programs that have an option for everything in their menu so much so that even advanced users balk at checking everything that's available, you'll also understand my concern of not wanting to add too many niche options to the GUI options.

I don't intend to have any GUI change related to this concept for 1.5.1.

For Evolution, I could imagine having biased random buttons available for least and most played tracks, track selection and GP selection will be separated in different screens so it may be easier to fit cleanly. But I'm not sure how much interest there is for this idea in general.

Alayan-stk-2:
Due to the reasons discussed in my previous comments, I'm closing this PR.

PR #5666, opened 2026-03-12
Fix mouse wheel scrolling in lists without requiring click first

Al2Klimov:
In CGUIEnvironment::postEventFromUser(), mouse wheel events were only dispatched to the focused element. Focus only changes on left-click, so hovering over a list and scrolling wouldn't work until the list was clicked.

Add a fallback: when a mouse wheel event isn't handled by the focused element, also try sending it to the hovered element.

## Agreement
```

```


Alayan-stk-2:
Code that's clear, simple, and does what it says, what's not to like?

Now that I've finally got around to testing it and confirming it works as intended with no apparent regression, I'm merging.

Thank you for your PR!

PR #5670, opened 2026-03-17
Fix 3574

### LINKED ISSUE #3574: The "show private server" checkbox is confusing
This checkbox is a toggle between showing public servers (no password) and private servers (password).

The previous formulation was "Show private servers only", which accounted for the fact that public servers are not shown when the checkbox is toggled, was changed because it somewhat implied that private servers were shown before checking the box.

But the new formulation is itself partially misleading.

Also, I don't like the (s) in "server(s)".

BenjaminCap2:
## Agreement
```

```
Fixed #3574 by replacing the checkbox with a spinner, giving the user the choice between public and private servers. Also removed the ‘s’ from ‘server(s)’ as requested in the issue.

### DIFF
--- a/src/states_screens/online/server_selection.cpp
+++ b/src/states_screens/online/server_selection.cpp
@@ -111,8 +111,8 @@ void ServerSelection::loadedFromFile()
     assert(m_server_type != NULL);
     m_server_type->m_properties[GUIEngine::PROP_WRAP_AROUND] = "true";
     m_server_type->clearLabels();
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Thank you for the PR!

A spinner widget is an elegant solution to this issue, as it tells the user clearly what is currently shown while implying clearly what is not.

My remark back then was not so much about the "s" but rather about the parentheses surrounding it. Since there is basically always multiple public and private servers, we can just go with the plural form.

Issue #5457, opened 2025-06-13
Soccer Team selection doesn't check for window height


Alayan-stk-2:
A lot of key dimensions are computed based on the screen width only.

So on splitscreen team selection with an ultra-wide game window, the kart preview can completely overflow, especially with 3 karts or more.

Labelling as minor simply because it appears to be extremely rare for players to run into that situation.

![Image](https://github.com/user-attachments/assets/2680ae6d-6ce9-49bf-b83a-e907733e5b8c)

Related to but different from https://github.com/supertuxkart/stk-code/issues/5286

Alayan-stk-2:
Although the overflow issues themselves were fixed with the PR merged earlier, commit https://github.com/supertuxkart/stk-code/commit/d64e55f05fca05a9164e40735f0557de7d8019c2 has further improved soccer display.

The number of columns and rows changes to allow the best display.

<img width="846" height="605" alt="Image" src="https://github.com/user-attachments/assets/8cbac4e5-3e7d-4823-b90f-5057de3d05c9" />
<img width="846" height="605" alt="Image" src="https://github.com/user-attachments/assets/d769b1f4-cdd6-4bd4-b269-876fb1650f1b" />
<img width="846" height="605" alt="Image" src="https://github.com/user-attachments/assets/5f92ffac-5b1a-470e-881d-f811d64bbe0c" />

PR #5671, opened 2026-03-17
Fix soccer setup screen kart scaling based on window height

bricealbiol:
Fix Issue #5457 

This Pull Request fixes the issue where kart models in the soccer setup screen were being cut off vertically on certain window sizes or aspect ratios. 

Previously, the `kart_view_size` was calculated based solely on the available width. This fix introduces a check against the available vertical space (`central_div->m_h`) to ensure the 3D models are scaled down if the window height is too small to display them fully.

## Changes
* **Dynamic Scaling**: Removed `const` from `kart_view_size` to allow its value to be adjusted based on layout constraints.
* **Height Validation**: Added logic to calculate the maximum number of rows among teams and verify if the karts fit within the vertical area.
* **Improved Layout**: The size of the models is now the minimum between the width-based size and the height-based size, preventing any clipping.

## Agreement
```

```

### DIFF
--- a/src/states_screens/soccer_setup_screen.cpp
+++ b/src/states_screens/soccer_setup_screen.cpp
@@ -439,7 +439,8 @@ void SoccerSetupScreen::updateKartViewsLayout()
     // Compute/get some dimensions
     const int nb_columns = 2;   // two karts maximum per column
     const int kart_area_width = (int)((central_div->m_w) / 2 * 0.8f); // size of one half of the screen with some margin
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
This PR fixes the issue reported in the ticket, it's a clear improvement and I will merge it.

However, the karts can get tiny with this setup, the issue being that the number of column is set at a maximum of 2 with no code to try to guess good sizes based on available width, and the margins might also be a tad excessive.

PR #5675, opened 2026-03-27
Spectators should not be able to delay the start of the game

BcfWor:
## Agreement
```

```


Alayan-stk-2:
There are some rare situations where it might be desired to ensure a race or game does not start before select spectators are fully loaded in the game, but those are rare and even rarer are situations where the start countdown doesn't add enough time for them to load in. In the more common case, this prevents pointless waiting.

PR #5677, opened 2026-03-28
Fix incorrect argument count in plural friend-request message.

nutzboi:
## Before
image

```yml
[warn   ] StringUtils: insertValues: Invalid number of arguments in 'You have %d new friend requests!'
```

## After
image

## Agreement
```

```


Alayan-stk-2:
Good catch. Apparently, this code was unchanged for the last 12 years...

I added small comments to standardize with the comments used for the other user of the plural translation functions in the code base, it makes it clearer why we need to pass the argument twice.

PR #5683, opened 2026-03-31
Add a "Player finished the race" message in the race GUI

GustaLOLE:
- This message is shown when a player finishes the race;
- This message appears in blue in the same place of the fastest lap message.

A issue that I've seen here happens when various players finish the race at the same time, and the screen are polluted with various messages.

## Agreement
```

```

### DIFF
--- a/src/modes/linear_world.cpp
+++ b/src/modes/linear_world.cpp
@@ -67,6 +67,7 @@ LinearWorld::LinearWorld() : WorldWithRank()
     m_valid_reference_time = false;
     m_live_time_difference = 0.0f;
     m_fastest_lap_kart_name = "";
+    m_finished_kart_name    = "";
     m_check_structure_compatible = false;
 }   // LinearWorld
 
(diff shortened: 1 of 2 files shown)


qwertychouskie:
Maybe a cleaner option would be to show a flag icon on the player's icon in the top left area for players that have finished?

Gustanator:
The flag can be useful, but ALL racing games I know have this message, and this shows also the username of the player (or the kart name in offline modes) instantly, when the player icons ranking shows this only after few seconds.

Alayan-stk-2:
I'm not sure how useful this is considering what @qwertychouskie pointed out.

It's very common to have 2 or 3 players finishing the race within a second or two, so the message spam you mentioned would be a significant issue. It's possible to have logic to merge messages when additional players finish soon after another ("X and Y have finished the race", "X, Y and Z have finished the race"). But I'm not quite sold on the value this adds.

GustavoBarreira:
Well, I can test the merging, or I can change the timing of the player name appear in the ranking


This can the useful in case of endurance or many players in a race, or even help to estimate the amount of points gained (or lost) in a ranked match (how I made in the ranked server, as I care so much with the ranking)

P.S.: I'm talking by this account because I'm using a alt account in my phone (due login issues), but I'm GustaLOLE anyway.

Gustanator:
bye :)

PR #5687, opened 2026-04-04
reduce memory usage of generating cubic skybox

virophagesp:
## Agreement
```

```
This is a work in progress since my local experimental branch has the changes in a hard to read form


virophagesp:
realized I should put my other changes in a separate pull request because of how different they are

virophagesp:
this commit saves a surprising amount of memory, according to my tests, 20 mib less memory spike during loading a track on my device

virophagesp:
it's ready, pretty sure it doesn't introduce any bugs

virophagesp:
found some similar changes I can make, I'll open a new one when ready

Issue #5560, opened 2025-11-07
Show collected eggs on the minimap


Alayan-stk-2:
As the entire point of egg hunt is exploration, we obviously don't want to give any indication about location before the player has found and collected an egg.

However, it could be nice to indicate on the minimap the location of eggs that the player has already collected. That way, when trying to find the last elusive egg or two, the player will have a clearer picture of where to search.

PR #5690, opened 2026-04-16
Collected eggs show on the minimap

TuxFan53:
Based off of issue [5560](https://github.com/supertuxkart/stk-code/issues/5560)

Fixes the issue by getting the location of the player when collecting an egg and adding it to a vector within the egg hunt class. The vector resets if the player resets the race. 

(I also couldn't find an image of an egg, so I just made a low quality one for now). 

## Agreement
```

```


Alayan-stk-2:
Thank you for the pull request and welcome to SuperTuxKart development!

I'll have to test it but from a cursory look at the code it seems good.

I'm not too sure about merging it in master immediately though, as 1.5.1 had been planned as a bugfix/localization release mostly.

Alayan-stk-2:
The icon is of course not ideal but that was already stated in your PR comment and it's something that could be changed afterwards.

In my testing, your PR appears to work as intended.

However there is one issue I noticed which is related to how the minimap and egg hunt work. This is a challenge to the general idea from the linked ticket.

Egg_minimap

Egg_minimap2

You can see on the right of the minimap in each case there is an egg which is cut. In the minigolf case, it's barely visible.

This is because the minimap is generated based on the racing drivelines. If an egg is some distances away from the driveline, it may therefore be partially cut from the screen or appear in strange places.

Toy Block Raceway is an addon rather than an official track, but it gives a good idea of how bad it can potentially get (although Minigolf is close):

Egg_minimap3

Several eggs are completely out of the minimap area here.

This is the same set of collected eggs but with the "centered" minimap UI option instead of "right-side".

Egg_minimap4

The egg on the right appears but the bottom egg is now missing. I suspect that if I had tested other minimap display option for Minigolf, the bottom egg would have also been out of the screen.

Alayan-stk-2:
Github really didn't want to let me update that master branch to squash cleanly with attribution...

But it's done. I'm merging in the Evolution branch as while the change is cool, these eggs going out of the screen are an issue that needs solving.

PR #5694, opened 2026-04-20
Obfuscate Username on Windows, Linux and Mac in displayed Ghost Path

Pttn:
Often, people use a system username they do not wish to reveal, for example if it is linked to their real names, however it is displayed in the "Replay saved in ..." message when a race is finished with Ghost recording. It is easy to accidentally leak the username in a video if one does not notice the message, and I have seen lots of Speedrunners censoring it manually in submissions.

This Pull Request alleviates the issue by obfuscating the username when it recognizes the usual paths used on Windows (`C:\Users\...`), Linux (`/home/...`) and Mac (`/Users/...`), which already covers a great part of Players.
There does not seem to be a practical way to handle this generically for all systems or esoteric configurations, recognizing and hardcoding common paths seemed to be the best compromise.
More paths for other systems may be added in follow-up Pull Requests.

ObfuscateUsernameGhost

An alternative could be to just omit the first part of the path or even just showing only the Ghost Filename, and maybe have some practical way to open the Ghost Folder directly from the Ui.

## Agreement

By creating a pull request in stk-code, I hereby agree to dual-license my contribution as GNU General Public License version 3 or any later version and Mozilla Public License version 2 or any later version.

This includes my previous contribution(s) under the same name of contributor.

### DIFF
--- a/src/io/file_manager.cpp
+++ b/src/io/file_manager.cpp
@@ -858,6 +858,30 @@ std::string FileManager::getStdoutDir() const
 }   // getStdoutDir
 
 //-----------------------------------------------------------------------------
+/** Obfuscates system username for common paths.
+ */
+std::string FileManager::obfuscateUsername(const std::string &path)
+{
+    std::string obfuscated_path(path), asterisks("***");
+    if (obfuscated_path.size() > 6 && obfuscated_path.substr(0, 6) == "/home/") { // Linux usual Home Dir.
+        size_t next_delimiter_pos(obfuscated_path.find('/', 6));
+        if (next_delimiter_pos != std::string::npos)
+            obfuscated_path.replace(0, next_delimiter_pos, "~");
+    }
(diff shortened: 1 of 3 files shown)


qwertychouskie:
For Linux, ideally `/home/$USER/blah` would be replaced with `~/blah`.

Also, a similar check should be added for macOS (`/Users/<username>/blah`).

Pttn:
Good idea about `~`.
Let's add Mac's `/Users/...`. No equivalent of `~` in Mac I suppose?

ObfuscateUsernameGhost2

HaenschenSTK:
This would be super useful for speedrunners. I always have to manually blur my username in my recordings. I agree that `~` should be used for linux and macOS so that the path remains correct without revealing the username. Im not sure about Windows but I think newer versions support `~` as well(?)

Pttn:
Turns out `~` is well established for Mac too. On Windows `~` does work in PowerShell, though I think that `~` would likely just confuse the average Windows User so it seems better to do the Asterisks obfuscation in this case...
Code was edited to use `~` for Linux and Mac.

PR #5703, opened 2026-04-21
Remember player 1's last kart class filter (light/medium/heavy)

Al2Klimov:
This is especially useful when player 1 plays random karts of a particular class all the time.

## Agreement
```

```

### DIFF
--- a/src/states_screens/kart_selection.cpp
+++ b/src/states_screens/kart_selection.cpp
@@ -311,7 +311,9 @@ void KartSelectionScreen::init()
     const std::vector<std::string> &classes = kart_properties_manager->getAllKartTypes();
     GUIEngine::SpinnerWidget* kart_class = getWidget<GUIEngine::SpinnerWidget>("kart_class");
     assert(kart_class != NULL);
-    kart_class->setValue(classes.size()); // All
+
+    const int last_class = UserConfigParams::m_last_used_kart_class;
+    kart_class->setValue(last_class < 0 ? classes.size() /* All */ : last_class);
 
     // Build kart list (it is built everytime, to account for .g. locking)
     setKartsFromCurrentGroup();
@@ -1242,6 +1244,18 @@ void KartSelectionScreen::allPlayersDone()
         if (n == PLAYER_ID_GAME_MASTER)
         {
             UserConfigParams::m_default_kart = selected_kart;
+
+            if (selected_kart == RANDOM_KART_ID)
+            {
+                SpinnerWidget* kart_class = getWidget<SpinnerWidget>("kart_class");
+                assert(kart_class != NULL);
+
+                UserConfigParams::m_last_used_kart_class = kart_class->getValue();
+            }
+            else
+            {
+                UserConfigParams::m_last_used_kart_class = -1;
+            }
         }
 
         if (selected_kart == RANDOM_KART_ID)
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
I'm not convinced here, because many times the filter can be used to help find faster a specific kart (or to see all the options available in a class) without communicating intent to keep playing that class next time.

I don't have hard data, but I would guess that while this would save some inputs for some players, it would just as well require additional inputs from a greater number of players.

The more extreme version of this suggestion would be remembering what the player typed in the search filter and having it pre-typed the next time the player comes.

Al2Klimov:
STK already remembers the group, so why not also the class/weight?

Alayan-stk-2:
The point of remembering is to gain time for the player by having as a selected default what he is most likely to want to use again.

If a player used a specific category, the game assumes that the player is going to want to pick again from that category next time.

Your PR proposes to do the same for the kart class filter.

Let's look at the situations:
* Player didn't use the class filter, nothing changes
* Player used the class filter.
  * (1) The player consistently want to use karts of this class
    * (a) The player is satisfied with using the same kart again and again (quite common). In this case pre-selection does the job, this PR can help when using the mouse if it's crowded.
    * (b) The player wants to use different karts regularly, but within the same weight class. **Here your PR helps reduce inputs.**
  * (2) The player wanted to find more easily a kart or check which karts belong to a certain class. No specific intent to keep using it. **Here with the PR the player needs to do additional inputs to restore the filter state. Furthermore, the player may not realize that these inputs are needed and may be confused**.

I generally think that case (2) outweighs case (1b) by quite a lot within STK's playerbase, although for some individual players that may not be the case.

There are many GUI settings that return to default once you leave the screen. To name a few: sorting options for lists in basically any screen with list, filtering options in the addons screen.

For technical users, this can be a little annoyance if they have some specific preferences, but default normally does a good job, and for normal users it's actually quite nice for thing to return to default as they can easily reset any change they've done just by leaving the page and going back.

Al2Klimov:
1c) The player (me) always wants a random kart, but only from a particular class. Together with #5692, this PR makes the selection screen zero-click.

Alayan-stk-2:
Your 1c) is more or less a subsest of 1b).

The point is that it's good for some use cases and bad for some other use cases. When doing or merging GUI changes, I have to think about the playerbase at large.

Al2Klimov:
Now the implementation entirely focuses in 1c), without the disadvantage of 2).

src/states_screens/kart_selection.cpp
```diff
@@ -1242,6 +1244,18 @@ void KartSelectionScreen::allPlayersDone()
         if (n == PLAYER_ID_GAME_MASTER)
         {
             UserConfigParams::m_default_kart = selected_kart;
+
+            if (selected_kart == RANDOM_KART_ID)
+            {
+                SpinnerWidget* kart_class = getWidget<SpinnerWidget>("kart_class");
```
Al2Klimov, line 1250: It does still remember your kart class, BUT only if you selected the random kart (case "1c")

src/states_screens/kart_selection.cpp
```diff
@@ -1242,6 +1244,18 @@ void KartSelectionScreen::allPlayersDone()
...
+                assert(kart_class != NULL);
+
+                UserConfigParams::m_last_used_kart_class = kart_class->getValue();
+            }
+            else
+            {
+                UserConfigParams::m_last_used_kart_class = -1;
```
Al2Klimov, line 1257: For any other kart, it forgets your kart class to behave as nothing happened the next time

Al2Klimov (commented):
Precisely speaking:

PR #5704, opened 2026-04-24
Add a security lock in android/generate_assets.sh

GustaLOLE:
The title says all :)

I added this in case of someone type this command without wanting to when he already have the assets converted, and need to wait a long time to convert it again, as I talked in the Matrix chat.

## Agreement
```

```

### DIFF
--- a/android/generate_assets.sh
+++ b/android/generate_assets.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 #
 # (C) 2016-2017 Dawid Gan, under the GPLv3
+# Edited in 2026 by Gustavo Barreira (aka GustaLOLE)
 #
-# A script that generates data files for Android apk
+# A script that generates lighter data files for Android apk
 
 
 # Below you can find some simple configuration variables.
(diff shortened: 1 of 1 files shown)


Gustanator:
I will "undraft" this finally, @Alayan-stk-2

android/generate_assets.sh
```diff
@@ -718,6 +726,6 @@ echo "has_assets" > "$OUTPUT_PATH/has_assets.txt"
 # It will be probably ignored by ant, but create it anyway...
 touch "$OUTPUT_PATH/.nomedia"
 
-
-echo "Done."
+# Just the finish message
+echo "Done ;)"
```
Alayan-stk-2: No need for this smiley

android/generate_assets.sh
```diff
@@ -149,6 +150,13 @@ else
     MAGICK='convert'
 fi
 
+# Ensure that don't exists the assets folder in case of a user error, like wrong typed command, etc...
```
Alayan-stk-2: Better formulation:

Ensure that the assets folder doesn't exist, in case of a user error, like a wrongly typed command, etc...

Alayan-stk-2 (commented):
If I understand correctly, this PR is mainly meant so that  you don't start generating the lighter assets if you already generated them?

GustavoBarreira:
Mainly, this blocks the regeneration if you had compiled all (models, audio, images, etc) and, without wanting to, you type this command (I have made this many times in the terminal when I was hurried)

android/generate_assets.sh
GustavoBarreira: > No need for this smiley

Could be more funny take this here, but I will remove

android/generate_assets.sh
GustavoBarreira: > Better formulation:
> 
> Ensure that the assets folder doesn't exist, in case of a user error, like a wrongly typed command, etc...

Thanks, my English is bad yet and I don't have much time to improve it now (as I am busy now), but thanks

PR #5719, opened 2026-05-03
reduce memory usage of generating the skybox during initalization of a level

virophagesp:
## Agreement
```

```
on my machine, according to massif valgrind, this reduces peak memory usage by a little more than 25 megabytes

### DIFF
--- a/src/graphics/skybox.cpp
+++ b/src/graphics/skybox.cpp
@@ -208,9 +208,10 @@ void Skybox::generateCubeMapFromTextures()
         bool needs_srgb_format = CVS->isDeferredEnabled();
         glBindTexture(GL_TEXTURE_CUBE_MAP, m_cube_map);
 
-        const unsigned tc_flag = squish::kDxt5 | stk_config->m_tc_quality;
         if (CVS->isTextureCompressionEnabled())
         {
+            const unsigned tc_flag = squish::kDxt5 | stk_config->m_tc_quality;
+
             unsigned tex_size = GE::get4x4CompressedTextureSize(size, size);
             uint8_t* compressed = new uint8_t[tex_size];
             squishCompressImage((uint8_t*)rgba, size, size, size * 4,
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Before I review the code, could you give a high-level explanation of the change and how it works?

While a permanent 25MB saving would have its worth, this appears to just be a temporary saving during a short initialization period, so functionally irrelevant in most cases. So between some additional compromises your commit messages indicate ("some potential speed cost", issues if the "size is an odd number"), and the need to review the changes, I'm not quite sold on this yet.

I think it would be useful for validation if you could make another branch where you have both the old and new skybox code and have some code to compare their calculations.

virophagesp:
the potential speed cost was referring to the loop for setting the values of the array called tmp, the array has been removed along with the loop and it's potential speed cost, I suppose I should've mentioned that in the commit where it was removed

virophagesp:
since you mention the odd number texture side length, I'll fix that, then I'll prepare the validation test

virophagesp:
surprisingly simple, now then

virophagesp:
here is an explanation of my changes, my first commit and main one that saves most of the memory ~20 megabytes, rgba was formerly an array of char arrays, however this was very unnecessary since it was only used in a single loop and only used the i-th index, so I changed it to be a single array initialized and deleted within that loop, the next main change from commits after the fourth are for rotation code only used on the top and bottom faces of the skybox texture they save around ~5 megabytes of memory, there are a few other more minor changes

Alayan-stk-2:
**Alright, thank you for your contribution. After reviewing, everything appears to work as intended, so I will be merging this.**

It was useful that you had separate commits for each step of your process to understand exactly what you did.

I added a comment to clarify the working of the latest commit with the modulo as it's not obvious at first glance that it is indeed correct.

More PRs improving code quality, performance, or resource usage would be welcome!

Depending on the area of code that is changed, it may make more sense to target the BalanceSTK2 (Evolution) branch instead of the master branch however, as the [code contribution guide explains](https://supertuxkart.net/How_to_contribute_code#what-to-work-on):

> Changes that improve code quality. This is usually not the focus of new contributors because it feels less exciting, but be assured that the STK team will be much happier seeing the code become simpler, clearer and faster rather than being made more complicated. As is, the game code of STK has a significant amount of technical debt, and there is plenty to do. Check first however if the problem is still relevant in the [STK Evolution branch](https://github.com/supertuxkart/stk-code/tree/BalanceSTK2), as it also includes some code quality improvements.

PR #5742, opened 2026-05-13
Out of view eggs are now displayed in an icon

TuxFan53:
Whenever an egg is outside of the map, it uses a frame to point towards the location of the egg (the same one from soccer game mode). The egg inside of the frame also gets smaller the farther away it is.

(This should hopefully work? I used the main branch for testing and it works well. I wasn't able to test it in the actual BalanceSTK2 branch due to assets being missing, but it was able to compile. So unless a change occurred to how players are rendered on the map it should work the same)

## Agreement
```

```

### DIFF
--- a/src/states_screens/race_gui.cpp
+++ b/src/states_screens/race_gui.cpp
@@ -108,6 +108,7 @@ RaceGUI::RaceGUI()
     m_basket_ball_icon = irr_driver->getTexture(FileManager::GUI_ICON, "rubber_ball-icon.png");
     m_champion = irr_driver->getTexture(FileManager::GUI_ICON, "cup_gold.png");
     m_egg = irr_driver->getTexture(FileManager::GUI_ICON, "egg.png");
+    m_out_of_view = irr_driver->getTexture(FileManager::GUI_ICON, "icons-frame_arrow.png");
 }   // RaceGUI
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


TuxFan53:
Just checked the forums and saw that the Evolution assets were released the same day the pull request was made, oops! 

Anyways I just tested it and it works as expected. Eggs off the minimap display in an icon frame and the farther the egg the smaller it is.

src/states_screens/race_gui.cpp
```diff
@@ -652,14 +653,60 @@ void RaceGUI::drawGlobalMiniMap()
...
-                                     m_map_left+(int)(draw_at.getX()+(m_minimap_player_size/2.5f)),
-                                     lower_y   -(int)(draw_at.getY()-(m_minimap_player_size/2.5f)));
-            draw2DImage(m_egg, position, source, NULL, NULL, true);
-        }
+                float positionX = 0.0f;
+                float positionY = 0.0f;
+                float map_half_width = m_map_width*0.5f;
+                float map_half_height = m_map_height*0.5f;
+                //uses posy to get posx if the slope of the points is higher than the map's
+                if (abs((draw_at.getY() - m_map_height*0.5f)/((draw_at.getX() - m_map_width*0.5f))) >
```
Alayan-stk-2, line 669: The purpose and working (why are the formulas like that?) of this if/else isn't clear at first glance and the comment doesn't do a very good job at clarifying it.

I suppose this is related to the fact that even if say the X value is within bounds, if the Y value is out of bounds, the X value is actually getting changed too when projecting the out-of-map position to a map-border position?

src/states_screens/race_gui.cpp
```diff
@@ -652,14 +653,60 @@ void RaceGUI::drawGlobalMiniMap()
...
+                int distance = (pow((positionX-draw_at.getX()), 2)) + (pow(positionY-draw_at.getY(), 2));
+                float scaler = 1/((0.0001f*distance)+1);
+                source = core::rect<s32>(core::position2di(0, 0), m_egg->getSize());
+                position = core::rect<s32>(m_map_left+(int)(positionX-((scaler)*m_minimap_player_size/2.5f)),
+                                           lower_y   -(int)(positionY+((scaler)*m_minimap_player_size/2.5f)),
+                                           m_map_left+(int)(positionX+((scaler)*m_minimap_player_size/2.5f)),
+                                           lower_y   -(int)(positionY-((scaler)*m_minimap_player_size/2.5f)));
+                draw2DImage(m_egg, position, source, NULL, NULL, true);
+            }
+            else
```
Alayan-stk-2, line 700: In the case the egg is not out of the map, you essentially do the same final drawing operation, it's just that the scaling value is 1.0 and positionX / positionY are different than the draw_at getX/getY. I would take these final operation fully entirely out of the if/else, have the scaling variable and positionX/positionY defined/set before the if, and modified within the if in case the egg is out of the map.

Alayan-stk-2 (changes requested):

Alayan-stk-2:
I tested the PR, and it appears to generally work well.

I don't think it supersedes the concept of having enhanced minimaps for egg hunt that can show more than just the racing mode's drivequads, but it's an overall good change.

There is some noticeable pixellation with the icons but that's STK's fault and not something wrong in this PR.

I think there are a few things that could be improved though. I pointed out two things in the code review. I think there could be some tweaks functionality-wise too, but it's not really worth thinking about those without the enhanced minimap present for context.

Minimap_egg

PR #5747, opened 2026-05-15
astcenc_context_alloc missing arg

matahina:
as seen on
https://github.com/ARM-software/astc-encoder/blob/main/Source/astcenc.h line 69

## Agreement
```

```


Alayan-stk-2:
Considering the limited scope of use of this library and how the game can easily be compiled without, it's not really worth the headache of something more advanced to detect versions.

The dependencies have been updated to have 5.5.0 (the breaking change was introduced in the lead up to 5.4.0)

PR #5748, opened 2026-05-16
Add libastcenc as a bundled library

Alayan-stk-2:
The library maintainer said that 'The library interface has never guaranteed cross-version compatibility across minor versions (arguably it should, but we never have in the past). This isn't a library we expect to be installed as a system library and applications can manage which version they are using, so it's trivial to align versions and manage interface changes'.

Therefore, it makes sense to bundle the library.

This library is only used as an optional part of the Vulkan pipeline, so I've set it as OFF by default.

Add -DUSE_LIBASTCENC=ON to cmake turn on libastcenc usage. By default it will generate a native compile which won't work on other CPUs.
Add -DUSE_SYSTEM_LIBASTCENC=ON to cmake to tell it to use the system library.

Also update the API call in ge_compressor_astc_4x4.cpp in lib/graphics_engine to work with the newer API (as I'm bundling the most recent version of the library)

---

There are several concerns that make me put this in a PR for now:
- I'm not too used to tweaking CMakeLists and while I've done basic testing (compiling with USE_LIBASTCENC ON/OFF), I am not fully sure everything is correct.
- GitHub Actions have not been updated. In particular, we may want to run it with ON for iOS (which uses Vulkan by default), but in that case we also need to guarantee that the generated compile will work for all the target CPUs, which the current CmakeLists don't.
- The dependencies still point to a commit on the external repo. If the lib is bundled it probably shouldn't be kept as part of the dependencies.


Alayan-stk-2:
See #5747 for my reasoning. Bundling the lib is excessive considering the very limited scope of its use within STK, although the policy of breaking API changes upstream may cause trouble again someday.

PR #5749, opened 2026-05-20
Fix a Gradle error in Evolution branch building

GustaLOLE:
- Someone forgot to put as parameter the type variable, so (only in Android building), the compíler fails.
 - So I put this basically.

## Agreement
```

```


Alayan-stk-2:
Good catch! I'm not compiling for mobile during development, so small oversights like this may happen from time to time.

PR #5750, opened 2026-05-21
fix build failure on evolution when building server only

virophagesp:
## Agreement
caught this while preparing to help with #5070 
```

```


Alayan-stk-2:
SERVER_ONLY isn't tested too often, partly because issues are fairly easy to fix once caught and partly because such builds are rarely used at this stage of development, so I'm not surprised that issues sometimes slip in in the Evolution branch.

I'm merging your fix.

About #5070, it's an important task and help would certainly be welcome there!

Issue #5730, opened 2026-05-08
Weird emitter visible in kart selection


Alayan-stk-2:
This happens in the Evolution branch but not in the master branch, so this is presumably some sort of regression.

<img width="587" height="518" alt="Image" src="https://github.com/user-attachments/assets/b32c121f-2e9d-4e52-8482-d3d53e6dfb3c" />

ewingwm:
The regression was caused by the commit f75cefeeac2c2629ca7e74064d024f7fe9b1e574 . It is likely rendering as some kind of attachable library node (just a guess, I have not looked to deep into it yet). I will research this commit further to see if I can mitigate this issue.

Alayan-stk-2:
Something related to the AttachableLibraryNode was indeed at the top of my suspicions, it made the most sense with when the issue was noticed.

It's actually the full attachable library node, just really tiny.

Of note is that when starting a race and going back to kart selection, the issue would disappear so presumably there is a change in context that made it only trigger when the scene manager was still "clean" from the initial game loading.

Anyways, since the solution is so trivial, there is no point looking into it in much more details.

Alayan-stk-2:
Fixed with https://github.com/supertuxkart/stk-code/commit/90792e62fb84cd0e5244c68eeced0ee522f6ef0f

PR #5757, opened 2026-05-28
Hide empty parent node of the attachable library node tool.

ewingwm:
This should hide the emitter visible in the kart selection screen and not break attachable library node functionality. Resolves https://github.com/supertuxkart/stk-code/issues/5730. 

## Agreement
```

```


Alayan-stk-2:
The code is very simple, but you also had to do some testing to confirm the bug's cause and validate the fix, so thank you!

Your comment about what this do is slightly off though, and I think it's probably why the bug slipped in in the first place.

Why would an empty scene node display anything? That doesn't make sense. What you are truly doing by setting the empty parent node to false is also determine the visibility of its children nodes which contain the visible elements. The full attachable library node was displayed, it was just very very small!

PR #5758, opened 2026-05-28
remove a few common guiengine functions from SERVER_ONLY builds

virophagesp:
## Agreement
```

```


Alayan-stk-2:
It's very important for SERVER_ONLY builds to not display graphics (beyond the GUI engine, the graphics rendering logic could be stripped from the SERVER_ONLY builds in the long-term I think), but there are other cases that require NoGraphics too (such as running test races from the command line), so you now have tons of these in your draft code:

```
#ifndef SERVER_ONLY
    if (GUIEngine::isNoGraphics())
```

In the end, the core motivation behind removing GUI code from SERVER_ONLY builds isn't to make the executable a bit leaner (although that's a plus), it's to reduce accidental bugs from forgotten SERVER_ONLY define guards and to reduce occurrence of these define guards within functions. Which we aren't really getting here...

Maybe the isNoGraphics/disableGraphics logic shouldn't be located in GUI Engine in the first place?

virophagesp:
in that case I am going to remove most of my changes except the ones that make it less complicated

PR #5768, opened 2026-06-08
improve SERVER_ONLY build include guards

virophagesp:
## Agreement
```

```
helping out on #5070

### DIFF
--- a/src/graphics/irr_driver.cpp
+++ b/src/graphics/irr_driver.cpp
@@ -418,8 +418,9 @@ void IrrDriver::initDevice()
     core::stringw display_msg;
 
     m_logger_level = irr::ELL_INFORMATION;
-#ifndef __SWITCH__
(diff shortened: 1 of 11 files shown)


virophagesp:
realized I did not and can not test these changes

virophagesp:
checked several times manually, I don't think there are bugs? though still cannot test it

Gustanator:
@virophagesp, I can test it today in a few hours

Gustanator:
I could test now, and the code compiled without errors (just a warning in graphics files, but about an unused variable)

Alayan-stk-2:
Compiling works fine with the patch with both SERVER_ONLY on and off.

I also reviewed all the proposed changes and everything looks good to me.

I haven't run an actual SERVER_ONLY build to test things but I don't think there should be any issue if compilation worked fine.

PR #5771, opened 2026-06-11
Fix some warnings in Gradle compiling (in Evolution branch)

GustaLOLE:
- The warnings related in kart.hpp and movable.hpp (-Winconsistent-missing-override) are fixed now.

More warnings (of STK Evolution) gonna be fixed later by me also in another PRs.
## Agreement
```

```

### DIFF
--- a/src/karts/kart.hpp
+++ b/src/karts/kart.hpp
@@ -465,7 +465,7 @@ public:
     // ------------------------------------------------------------------------
     /** Returns a unique identifier for this kart (name of the directory the
      *  kart was loaded from). */
-    virtual const std::string& getIdent() const;
+    virtual const std::string& getIdent() const OVERRIDE;
     // ------------------------------------------------------------------------
     virtual void   updateGraphics(float dt) OVERRIDE;
     virtual void   createPhysics    ();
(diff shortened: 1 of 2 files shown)


Alayan-stk-2:
**Please analyze the purpose of code you touch before changing it.**

In your second commit here, you are touching code meant to change GCC's warning behaviors (which it did correctly), another compiling tool interpreting that pragma line and complaining because it doesn't recognize GCC's Wmaybe-uninitialized

Now, with GCC 16 the problematic warning is gone so the code is not needed anymore.

But you did not recognize that your change made all these lines completely useless:

```
// Shut up GCC complaining incorrectly about sd.m_country_code
#pragma GCC diagnostic push

#pragma GCC diagnostic pop
```

The correct actions were either not removing anything, or also removing all these lines since they don't serve any purpose anymore. In this specific instance, since it's not needed with a more recent GCC, they can be removed.

In coding, it's important to understand why something is there before removing it. Look up "Chesterton's Fence".

Gustanator:
As a not much experienced programmer, I didn't know how really works the #pragma, so thanks for the tip.

I will search more about this, and obviously remove this lines soon (maybe today)

Gustanator:
I fixed this now

PR #5779, opened 2026-06-22
Fix #5286 and remove a unused variable

### LINKED ISSUE #5286: Edge of some karts are cropped on soccer team selection screen
Some karts are chopped as they rotate during soccer team selection.

Image

GustaLOLE:
- I've fixed the issue by using a inverse relationship in the "scale" variable (used also in kart_selection and kart_color_slider screens)
 - Was removed this warning also:

stk-code/src/states_screens/kart_selection.cpp:292:13: warning: unused variable ‘placeholder’ [-Wunused-variable]
  292 |     Widget* placeholder = getWidget("playerskarts");
      |             ^~~~~~~~~~~

## Agreement
```

```

### DIFF
--- a/src/states_screens/dialogs/kart_color_slider_dialog.cpp
+++ b/src/states_screens/dialogs/kart_color_slider_dialog.cpp
@@ -105,7 +105,17 @@ void KartColorSliderDialog::beforeAddingWidgets()
     core::matrix4 model_location;
 
     // scaling the view basing in the kart length
-    float scale = 40.0f / kart_model.getLength();
+    float scale = 35.0f / kart_model.getLength();
+
+    // if the kart is too short, use a predefined scale
+    if (kart_model.getLength() < 1.45f)
+    {
+        scale = 35.0f;
+    }
+    else if (kart_model.getLength() < 1.75f)
+    {
+        scale = 30.0f;
+    }
 
     model_location.setScale(core::vector3df(scale, scale, scale));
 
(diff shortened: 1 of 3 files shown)


Alayan-stk-2:
This works to solve the cropping, but the issue I see with it is that it will make karts of different sizes appear to be of the same size in kart selection. That's maybe unavoidable with extreme cases like Jetbus, but for karts that are within a "normal range" of sizes, that's undesirable.

Also:
- The screen has free unused space on the sides. Now multiplayer kart selection and multiplayer football team selection may not offer this free space, and scaling can also cause issues with height (Jetbus gets very close to the camera while rotating so it tends to overflow on height too), but for the general case it'd help to increase the size of the showcase section. This would also give more margin before scaling has to be made proportional.
- It'd be kind of nice to have an indicator when zooming out (scaling down) is needed, but that's secondary and more of a different ticket.

Then we also have karts that cause an issue in height, but in practice that's just Pepper with her vertical jump. That could be argued to be a kart animation issue and I'm not expecting any fix for it here, but scaling that tries to use most of the available space based on kart length would upscale karts that are high but short which could lead to issues.

Gustanator:
Solved the scale of the standard karts (the last commit shows how I did it)

Alayan-stk-2:
This reintroduces the issue of karts of different sizes looking as if they are the same size, but I'll just fix things myself.

Gustanator:
> This reintroduces the issue of karts of different sizes looking as if they are the same size, but I'll just fix things myself.

It's better cuz I'm so busy now to fix this issues fast

Alayan-stk-2:
Now the issue of normal karts appearing to be of the same size when they are not is back.

Furthermore the "unused variable" warning removal involves removing a check that's actually useful to warn in case the GUI file is missing a widget it should have. The warning just comes from asserts only doing something in debug builds.

PR #5782, opened 2026-06-23
Only create RaceGUIMultitouch buttons when isn't watching a ghost replay (mobile STK)

GustaLOLE:
- Now, the screen is clearer in this situation, allowing the mobile player to watch a replay with less annoying things on-screen.
 - The player can open ihe pause menu using ESC or pressing the Android back button.

## Agreement
```

```

### DIFF
--- a/src/states_screens/race_gui_multitouch.cpp
+++ b/src/states_screens/race_gui_multitouch.cpp
@@ -34,6 +34,7 @@ using namespace irr;
 #include "karts/abstract_kart.hpp"
 #include "karts/controller/kart_control.hpp"
 #include "network/protocols/client_lobby.hpp"
+#include "race/race_manager.hpp"
 #include "states_screens/race_gui_base.hpp"
 
 #include <IrrlichtDevice.h>
@@ -46,6 +47,7 @@ RaceGUIMultitouch::RaceGUIMultitouch(RaceGUIBase* race_gui)
     m_race_gui = race_gui;
     m_gui_action = false;
     m_is_spectator_mode = false;
+    m_is_watching_replay = false;
     m_height = 0;
     m_steering_wheel_tex = NULL;
     m_steering_wheel_tex_mask_up = NULL;
@@ -164,11 +166,17 @@ void RaceGUIMultitouch::init()
                                         "android/steering_wheel_mask_down.png");
 
     auto cl = LobbyProtocol::get<ClientLobby>();
-    
+    auto rm = RaceManager::get()->isWatchingReplay();
+
     if (cl && cl->isSpectator())
     {
         createSpectatorGUI();
         m_is_spectator_mode = true;
+    }
+    else if (rm)
+    {
+        m_is_watching_replay = true;
+        createRaceGUI();
     }
     else
     {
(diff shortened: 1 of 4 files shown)


src/states_screens/race_gui_multitouch.cpp
```diff
@@ -236,39 +244,44 @@ void RaceGUIMultitouch::createRaceGUI()
     }
 
     m_height = (unsigned int)(2 * col_size + margin / 2);
-    
-    if (UserConfigParams::m_multitouch_controls == MULTITOUCH_CONTROLS_ACCELEROMETER ||
-        UserConfigParams::m_multitouch_controls == MULTITOUCH_CONTROLS_GYROSCOPE)
+    if (!m_is_watching_replay)
```
Alayan-stk-2: You don't need to wrap everything below in a new if block. The simplest thing to do is simply to have an early return if this condition is true.

src/states_screens/race_gui_multitouch.cpp
```diff
@@ -164,12 +166,18 @@ void RaceGUIMultitouch::init()
...
     auto cl = LobbyProtocol::get<ClientLobby>();
-    
+    auto rm = RaceManager::get()->isWatchingReplay();
+
     if (cl && cl->isSpectator())
     {
         createSpectatorGUI();
         m_is_spectator_mode = true;
     }
+    else if (rm)
```
Alayan-stk-2, line 176: I don't see the point of having an extra else if here, you could just set your parameter earlier when you check isWatchingReplay().

src/states_screens/race_gui_multitouch.cpp
```diff
@@ -236,39 +244,44 @@ void RaceGUIMultitouch::createRaceGUI()
...
+        m_device->addButton(BUTTON_SKIDDING,
+                            int(second_column_x), int(h - 1 * col_size),
+                            int(btn_size), int(btn_size));
+        m_device->addButton(BUTTON_FIRE,
+                            int(first_column_x),  int(h - 2 * col_size),
+                            int(btn_size), int(btn_size));
+        m_device->addButton(BUTTON_LOOK_BACKWARDS,
+                            int(first_column_x), int(h - 1 * col_size),
+                            int(btn_size), int(btn_size));
+        m_device->addButton(BUTTON_ESCAPE,
```
Alayan-stk-2: Why do we have two escape buttons now? That seems wrong.

Also, some Android users might not know the other methods available to trigger the pause menu if the button from the touchscreen is not available.

src/states_screens/race_gui_multitouch.cpp
```diff
@@ -236,39 +244,44 @@ void RaceGUIMultitouch::createRaceGUI()
...
-    m_device->addButton(BUTTON_LOOK_BACKWARDS,
-                        int(first_column_x), int(h - 1 * col_size),
-                        int(btn_size), int(btn_size));
+        m_device->addButton(BUTTON_ESCAPE,
+                            int(margin_top), int(margin_small),
+                            int(btn_small_size), int(btn_small_size));
+        m_device->addButton(BUTTON_RESCUE,
+                            int(margin_top + col_small_size), int(margin_small),
+                            int(btn_small_size), int(btn_small_size));
+        m_device->addButton(BUTTON_NITRO,
```
Alayan-stk-2: No nitro button makes sense since it can't be used when watching, however this also means that there is no way to see the nitro level of the kart that is being watched.

Alayan-stk-2 (commented):
The idea of removing clutter when watching a replay on a touchscreen makes sense, but there are multiple issues with this proposed implementation.

Gustanator:
I will start to solve it today

src/states_screens/race_gui_multitouch.cpp
Gustanator: Solved

src/states_screens/race_gui_multitouch.cpp
Gustanator: Used for a different thing now

src/states_screens/race_gui_multitouch.cpp
Gustanator: Lmao

I solved this btw

src/states_screens/race_gui_multitouch.cpp
Gustanator: I put the desktop speedometer

PR #5795, opened 2026-07-10
Fix #5329

### LINKED ISSUE #5329: Selectings some artist debug menu items on main menu crashes SuperTuxKart
Selecting some menu items in the artist debug menu on the main menu crashes SuperTuxKart.
For example right click and choose GUI > First person view (Ctrl + F2).

211ruben:
Fixes #5329 by adding the already existing check `if (!world) return false;` (that was only used for some options) to all previously failing options that only work while playing, not in menus (e.g. Top View, First Person View, etc).

## Agreement
```

```

### DIFF
--- a/src/utils/debug.cpp
+++ b/src/utils/debug.cpp
@@ -738,12 +738,14 @@ bool handleContextMenuAction(s32 cmd_id)
         world->escapePressed();
         break;
     case DEBUG_GUI_CAM_TOP:
+        if (!world) return false;
         CameraDebug::setDebugType(CameraDebug::CM_DEBUG_TOP_OF_KART);
         Camera::changeCamera(0, Camera::CM_TYPE_DEBUG);
         Camera::getActiveCamera()->setKart(World::getWorld()->getKart(kart_num));
         irr_driver->getDevice()->getCursorControl()->setVisible(true);
         break;
     case DEBUG_GUI_CAM_WHEEL:
+        if (!world) return false;
         if (!(World::getWorld()->getKart(kart_num)->isGhostKart()))
         {
             CameraDebug::setDebugType(CameraDebug::CM_DEBUG_GROUND);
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
You missed some entries that would also need it to prevent crashes (for example, attachment entries also mentioned in the original ticket).

However, I think it's not very good either way to have a lot of entries shown in the context menu that cannot be used because they are disabled.

Static action handling is already restricted to when there is no active world, preventing the menu from showing the disallowed options would be better I think.

Alayan-stk-2:
> However, I think it's not very good either way to have a lot of entries shown in the context menu that cannot be used because they are disabled.

I have hidden the majority of irrelevant entries with: https://github.com/supertuxkart/stk-code/commit/9ca1adc36147468d65f226c667e4a492ac6b83bd

There are still some that probably should be hidden too, only a small minority of artist debug mode options appear relevant outside of 3D scenes.

Because DEBUG_ entries can only be accessed through that context menu or through key shortcuts that check for a world's existence to be active, direct checking of world status within the DEBUG_ entries themselves should not be needed (in theory it could work as defense in-depth, but that'd mean a lot of extra checks to add and keep in sync)

211ruben:
Your version is better, I think this PR can be closed.

> There are still some that probably should be hidden too, only a small minority of artist debug mode options appear relevant outside of 3D scenes.

I tested the still available options in the debug menu:
- Graphics
    - SSAO/Shadow/Bounding Boxes viz: works outside scenes, but only when starting a game after that, they actually are relevant
    - Pysics Debug: seems to be irrelevant outside scenes
    - Network Debugging: only relevant in online games
- SP Debug: irrelevant outside scenes
- GE Debug: vulkan only, irrelevant ouside scenes
- Lightning: irrelevant outside 3d scenes

Issue #5796, opened 2026-07-12
Make the shift key speed up the first person camera in artist debug mode.


SvenAndreasBelting:
The first person cam is a great way to quickly look at visual changes made to a track or take screenshots.
However it is quite slow which can be annoying and waste a lot of time.
Holding down the shift key to speed it up would be a good way to increase it's usuability.

I suggest to speed up the cam by about 3-4 times the normal speed when the shift key is pressed.

A way to change the speed value would also be cool but is not very important.

Alayan-stk-2:
Done with https://github.com/supertuxkart/stk-code/commit/f92625f5e8636dbc5918e9f291a8e1b0b89a135d

PR #5798, opened 2026-07-12
allows moving the fps camera in artist debug 3x faster by pressing shift, as suggested in #5796

211ruben:
Allows moving the first person view camera in artist debug mode 3 times as fast by pressing shift, as suggested in #5796 .

## Agreement
```

```


Alayan-stk-2:
Thank you for your pull request!

Everything looks good to me.

PR #5803, opened 2026-07-18
While items are switched, invert the zipper into a brake

Al2Klimov:
Tested on NixOS, btw

## Agreement
```

```


Alayan-stk-2:
Ground zippers are a crucial part of track design, and some of them cover the entire width of a road. Changing their boosting properties could have very bad consequences, such as preventing karts from clearing jumps the ground zipper was supposed to help clear.

Issue #5631, opened 2026-01-30
Nintendo Switch homebrew port crashes on launch


chiro1879:
### Description

FW21 tweaked the kernel and memory mapping a bit, which made Super Tux Kart not bootable and it crashes extremely early when started

### Steps to reproduce

1. Download the latest release version of Super Tux Kart for Nintendo Switch
2. Open the homebrew channel by holding right bumper while opening a game or application
3. Move over to Super Tux Kart
4. Press A to open it
5. Crash

https://github.com/user-attachments/assets/4e784b01-fe5b-4e5e-89ee-7aad459e4a98

### Configuration

SuperTuxKart version 1.5
Installation source: Website
Device: Nintendo Switch 1 - Patched, Modchipped (Picofly)
OS: Horizon 21.2.0 | Atmosphere 1.10.2 | SysCFW
APU: NVIDIA Tegra X1 - aarch64 - Not overclocked
Custom aggressive fan curve

![Image](https://github.com/user-attachments/assets/2f141ae8-935a-4fba-9da6-131a75728c01)
![Image](https://github.com/user-attachments/assets/8d0ea162-8d77-4b83-9e92-db5b5fd0a82f)

### Error message

it crashes the app and not the console extremely early so I can't reasonably get the error without attaching a debugger or something

![Image](https://github.com/user-attachments/assets/6b4a5572-39aa-4633-b21e-32c6cebc41b8)

### Log File (stdout.log)

[Completely empty file, but exists]

### Input Configuration (input.xml)

[File does not exist, crashes too early]

Alayan-stk-2:
Looking things up, it appears that STK would need to be compiled with a newer version of [libnx](https://github.com/switchbrew/libnx). There are multiple references to libnx in switch related code but nothing that seems linked to a specific version, so it's probably a matter of what is pulled at build time. Libnx 4.10.0 came out after STK 1.5.

Does the 1.5 Switch file work if you patch it with [https://github.com/alula/hbpatcher](https://github.com/alula/hbpatcher)?

chiro1879:
I patched it with hbpatcher and it now boots properly and I'm able to sign in to an offline account. I tried playing a Grand Prix Normal Race on the map Antediluvian Abyss with Konqi selected as my character and it didn't crash or throw any errors the entire time. Having to patch the game to run on the current ABI probably isn't the best solution, but it works for now

chiro1879:
[...]

PR #5813, opened 2026-08-01
switch: update devkitpro version

Mstrodl:
required tweaking some freetype stuff to use the target system, because the new devkitpro version came with an updated cmake config for freetype.

## Agreement
```

```

### DIFF
--- a/.github/workflows/switch.yml
+++ b/.github/workflows/switch.yml
@@ -14,7 +14,7 @@ jobs:
     name: Build Switch
     runs-on: ubuntu-latest
     container:
-      image: 'devkitpro/devkita64:20241023'
+      image: 'devkitpro/devkita64'
     steps:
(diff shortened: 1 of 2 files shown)


Mstrodl:
Note I have not been able to test this on real hardware
I tried on a fork of ryujinx, but it crashed with an unimplemented MRS error while compiling the sp_normal_visualizer.frag fragment shader...

Mstrodl:
Hopefully fixes #5631

Alayan-stk-2:
The fails for apple/windows come from some issues in compiling updated dependencies that I hadn't time/motivation to finish fixing yet.

The important part here is the switch build, which works.

I'll merge the PR so that the build gets added to the preview releases, and hopefully we will get confirmation by someone with real hardware that it actually works.

PR #5817, opened 2026-08-04
Reject unsafe paths while extracting add-on ZIPs

acts-1631:
Add-on ZIP files are extracted recursively by extract_zip(). Archive member names were appended to the extraction directory without rejecting traversal components, allowing a ZIP entry with .. components to write outside the add-on tree.

Validate recursive member names before creating their parent directory or opening the output file. The check rejects absolute, drive-qualified, empty, dot, and dot-dot path components while preserving normal nested add-on files.

### DIFF
--- a/src/addons/zip.cpp
+++ b/src/addons/zip.cpp
@@ -15,6 +15,7 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
+#include <cctype>
 #include <string.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Thank you for your PR!

After testing, it appears that STK would indeed blindly follow relative path indications when extracting zips, which could lead to malicious zips potentially writing in directories they had no business to.

The PR's code fixes the issue and the code is clean, so that's great.

The resulting error message could be nicer, because we get both a new warn and the unspecific "Failed to unzip" error, which directly comes from the file being ignored because unsafe, but that's a very minor point.

PR #5818, opened 2026-08-04
Validate special ping packets before decoding

acts-1631:
The client receive loop handles unencrypted ping packets before normal event parsing. It accepted any packet with the five-byte ping prefix, then read a timestamp and a counted peer list without first checking that those bytes were present. A malformed packet from a connected server could terminate the listening thread.

Check the fixed ping header before decoding it, then verify that the remaining packet contains every advertised peer record. Invalid ping packets are dropped before they reach the decoder.

### DIFF
--- a/src/network/stk_host.cpp
+++ b/src/network/stk_host.cpp
@@ -1127,12 +1127,27 @@ void STKHost::mainLoop(ProcessType pt)
                 {
                     if (!is_server)
                     {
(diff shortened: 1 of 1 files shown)


acts-1631:
A friendly ping if you have time, for this and https://github.com/supertuxkart/stk-code/pull/5819

PR #5826, opened 2026-08-10
Fix libjpeg in android compilation

Gustanator:
- With the update of deps, libjpeg reorganized some source files to libjpeg/src, but the compiler doesn't recognize it.
 - So, the irrlicht build returns a fatal error: 'something.h' file not found in the source file that handles jpeg files.
 - To fix it, I put more one flag in irrlicht instructions (of Android.mk) linking this new folder to irr compiler.

The compiler now compiles sucessfully the game.

## Agreement
```

```


Alayan-stk-2:
The reorganization of libjpeg's structure had also required some changes in the compilation of dependencies: https://github.com/supertuxkart/dependencies/commit/033e6cb79fa294b1fc2a309cf9d172083e88fc4e

The change looks good to me, so I'll take your word that it fixes Android compilation.

PR #5830, opened 2026-08-12
Detect Linux touchscreens for multitouch auto mode

dixonSolutions:

## Summary
- `multitouch_active` 1 already meant "if available", but `SDL_GetNumTouchDevices()` is often 0 on Linux until the first finger.
- Scan `/proc/bus/input/devices` (and DMI chassis / Ubuntu Touch) so Auto works on Linux tablets.
- Add `multitouch_touch_only` (default **false**, so current "any touchscreen" behaviour is unchanged) and Off / Auto / Always in the touch settings dialog.
- One helper, `IrrDriver::isMultitouchEnabled()`, replaces the repeated 3-line check.

We ship a tablet build at [SuperTuxKart-Touch](https://github.com/dixonSolutions/SuperTuxKart-Touch) that turns the touch-only filter on by default. This PR is only the detection layer, not that product overlay.

## Test plan
- [ ] Android/iOS: Auto still enables the race GUI (touch-only is true on those platforms).
- [ ] Linux desktop, no touchscreen: Auto stays off.
- [ ] Linux tablet / Ubuntu Touch: Auto enables with `multitouch_touch_only` on.
- [ ] Laptop with a keyboard + touchscreen: Auto + touch-only off still matches old "if available"; touch-only on leaves mouse/keyboard.
- [ ] Touch settings: Off / Auto / Always and the touch-only checkbox persist after Apply.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/data/gui/dialogs/android/multitouch_settings.stkgui
+++ b/data/gui/dialogs/android/multitouch_settings.stkgui
@@ -3,7 +3,23 @@
     <div x="2%" y="2%" width="96%" height="96%" layout="vertical-row" >
         <header id="title" width="100%" height="fit" text_align="center" word_wrap="true" text="Touch Device Settings" />
 
-        <spacer height="35" width="10" />
+        <spacer height="20" width="10" />
+
+        <div width="90%" layout="horizontal-row" proportion="1">
+            <label proportion="1" align="center" text_align="right" I18N="In the multitouch settings screen" text="Touch controls"/>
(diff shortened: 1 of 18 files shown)


Alayan-stk-2:
https://supertuxkart.net/How_to_contribute_code#supertuxkarts-policy-on-ai-generated-code

The default pull request messages warned about this already.

PR #5837, opened 2026-08-19
Allow GhostController to move backwards, make ghost(s) visible on rewind

### LINKED ISSUE #5834: GhostController::update() can only advance its replay index forward
### Description

GhostController::update() ([ghost_controller.cpp:44-47](https://github.com/supertuxkart/stk-code/blob/master/src/karts/controller/ghost_controller.cpp#L44-L47)) reads the race clock into m_current_time, then walks m_current_index forward until it catches up. There is no corresponding backward walk, so if the clock ever moves backwards the index stays where it is and the ghost freezes in place: it keeps interpolating against a pair of recorded transforms that the clock has already passed.

This is currently unreachable in stock SuperTuxKart, since nothing ever drives World::getTime() backwards. That's presumably why it hasn't been noticed. It becomes reachable the moment replay seeking exists, which is what [#5197](https://github.com/supertuxkart/stk-code/issues/5197) asks for, so I'm filing it separately as the thing that blocks that feature.

There's a related latch in GhostKart::update() ([ghost_kart.cpp:119-121](https://github.com/supertuxkart/stk-code/blob/master/src/karts/ghost_kart.cpp#L119-L121)): reaching the end of a replay calls m_node->setVisible(false), and in watch-replay mode nothing restores it. GhostKart::reset() sets it back to visible, so a fresh viewing is fine, but seeking backwards from the finish within a single viewing would leave an invisible kart. The engine sound and GFX recover on their own since they're driven every frame; only the scene node is latched.

If you want evidence, I lifted the index logic out into a standalone harness and drove it against every replay I have locally (113, including the ones that ship with the game), comparing against an independently computed reference. On ordinary forward playback the current logic is correct on every file. Across 20,000 random seeks it returns the wrong index 18,181 times. A backward walk that mirrors the existing forward one fixes all of them, and provably cannot execute during forward playback: after the forward search m_all_times[m_current_index] <= m_current_time, and the next frame's time is larger still.

I found this while reading the ghost replay code for an external tool.

### Steps to reproduce

There is no gameplay sequence that triggers this, because stock STK never moves the race clock backwards. It is observable by inspection, and becomes live as soon as seeking exists.

1. Open src/karts/controller/ghost_controller.cpp and look at update() (lines 40-49).

2. Note that m_current_index is only ever incremented, and only while m_current_time is ahead of the next recorded frame.

3. Observe that no path decrements it, and reset() is the only thing that returns it to 0.

4. Any caller that sets the race clock to a time earlier than m_all_times[m_current_index] therefore leaves the index stale, and the ghost stops           tracking the clock.

### Configuration

SuperTuxKart version: 1.5 (defect also present in master, c0f97e7)
Installation source: Arch Linux official repository (supertuxkart 1.5-1); master additionally built from source
OS: Arch Linux, kernel 7.1.5-arch1-2
CPU: Intel Core i5-10300H @ 2.50GHz
GPU: Intel UHD Graphics (CometLake-H GT2) / NVIDIA GeForce GTX 1650 Mobile
Input device: Logitech G502 Hero mouse, Logitech G413 keyboard

### Error message

_No response_

### Log File (stdout.log)

_No response_

schrowd:
As of the current version, GhostController::update() only has the ability to advance it's frame index. Consequently, any backwards movement made by the world clock leaves ghosts at a specific pair of transforms without advancing any further, effectively rendering them frozen. This adds the functionality for a backwards walk which restores the invariant that is documented in the header ("the last index in m_all_times that is smaller than the current world time"). Regular forward playback remains untouched as the condition for a backward ghost movement:

m_all_times[m_current_index] > m_current_time

never fires under normal playback as there is (as of writing) no feature in the game that can move the clock back. In other words, this change is inert during normal playback.

A latch in GhostKart::update() has also been fixed. When the end of a replay is reached, the node is hidden and nothing shows it again. If you were to scrub back from the finish, the kart would be invisible. GFX and the sound engine recover by themselves from the replay data every frame and so only the scene node needs restoration.

These fixes do nothing under normal gameplay as there is no feature that drives the clock backwards. However, these fixes are required by any feature that would be implemented in the future that *would* drive the clock backwards, such as replay seeking.

Fixes #5834

## Agreement
```

```

### DIFF
--- a/src/karts/ghost_kart.cpp
+++ b/src/karts/ghost_kart.cpp
@@ -138,11 +138,11 @@ void GhostKart::update(int ticks)
             // Start showing the ghost when it start racing
             m_node->setVisible(true);
         }
-	else
+        else
         {
             // When a replay ends, the node is hidden and nothing in
-	    // watch-replay mode restores it. GFX and sound recover
-	    // by themselves so there's no need to restore them.
+            // watch-replay mode restores it. GFX and sound recover
+            // by themselves so there's no need to restore them.
             m_node->setVisible(true);
         }
     }
--- a/src/karts/ghost_kart.cpp
+++ b/src/karts/ghost_kart.cpp
@@ -138,6 +138,13 @@ void GhostKart::update(int ticks)
             // Start showing the ghost when it start racing
             m_node->setVisible(true);
         }
+	else
+        {
+            // When a replay ends, the node is hidden and nothing in
+	    // watch-replay mode restores it. GFX and sound recover
+	    // by themselves so there's no need to restore them.
+            m_node->setVisible(true);
+        }
     }
 
     const float rd         = gc->getReplayDelta();
--- a/src/karts/controller/ghost_controller.cpp
+++ b/src/karts/controller/ghost_controller.cpp
@@ -38,6 +38,19 @@ void GhostController::reset()
 void GhostController::update(int ticks)
 {
     m_current_time = World::getWorld()->getTime();
+
+    // This has been done to make the class keep the promise that was
+    // documented in the ghost_controller header.
+    // Consequently, this allows for replay seeking to be implemented.
+    // It is safe because it only triggers when a backwards clock
+    // movement occurs, which is impossible in regular play as forward
+    // playback only ever increases the clock.
+    while (m_current_index > 0 &&
+           m_all_times[m_current_index] > m_current_time)
+    {
+        m_current_index--;
+    }
+
     // Find (if necessary) the next index to use
     if (m_current_time != 0.0f)
     {


src/karts/ghost_kart.cpp
```diff
@@ -138,6 +138,13 @@ void GhostKart::update(int ticks)
             // Start showing the ghost when it start racing
             m_node->setVisible(true);
         }
+	else
+        {
+            // When a replay ends, the node is hidden and nothing in
+	    // watch-replay mode restores it. GFX and sound recover
```
Alayan-stk-2: Inconsistent indentation

schrowd: My bad, fixed that plus some tab indentation.

schrowd:
Already merged in BalanceSTK2 branch

PR #5854, opened 2026-09-01
Drop the repeated track operand in track object presentation guards

darkdi:
Both guards in `track_object_presentation.cpp` test `track` twice, so the middle operand adds nothing. Both call sites pass `m_node` into the call underneath, and at the first one `m_node` comes straight out of `addMesh` and can be null, so if the second operand was meant to be a check on that rather than a repeat, say so and I will change it to `m_node` instead of dropping it.

### DIFF
--- a/src/tracks/track_object_presentation.cpp
+++ b/src/tracks/track_object_presentation.cpp
@@ -546,7 +546,7 @@ void TrackObjectPresentationMesh::init(const XMLNode* xml_node,
             enabled = false;
             m_force_always_hidden = true;
             Track *track = Track::getCurrentTrack();
-            if (track && track && xml_node)
+            if (track && xml_node)
                 track->addPhysicsOnlyNode(m_node);
         }
     }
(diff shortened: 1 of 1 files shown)


Alayan-stk-2:
Hello, and thank you for the PR!

What you point out is sound, however it has already been fixed in the Evolution branch (in this commit: https://github.com/supertuxkart/stk-code/commit/a864e94412ff3bc8cc3d6da85a96c51fd82ba8a6).

Since this code issue has no consequence on the compiled executable and since merging this PR would add a merge conflict later, I will close this PR.

Code quality fixes are very welcome, so don't take this PR being closed as a signal other PRs wouldn't be merged. It's just better to check the Evolution branch to make sure it's still applicable, [as explained in the contribution guide](https://supertuxkart.net/How_to_contribute_code#what-to-work-on).

darkdi:
Fair enough, I checked master and not Evolution. Will look there first next time.

Issue #5197, opened 2024-10-27
Watching ghost replay lacks control buttons


kimden:
## Description
This is a suggestion.

Watching ghost replays is useful for speedrunners, but looking at specific replay moments in detail is not convenient ingame (and it's only possible by recording the video of the replay). I think buttons doing the following functions could be useful:

- Pausing the replay
- Rewinding replay to a certain moment forwards / backwards
- Going one physical frame forwards / backwards
- Changing the speed of playback

I am not aware how hard all of those are. Feel free to correct me if some of those are hard / not possible to implement.

Alayan-stk-2:
Pause, rewind and speed change are all things I would like to have as well. This won't be in 1.5 and maybe not in 2.0, but it's definitely on the roadmap.

schrowd:
I've implemented the engine side of this and would like to offer it upstream if there's interest.

**What it does:** A small ReplayControl class that owns the replay clock, with seek(), setPlaying(), setRate() and advance(dt); a hook in WorldStatus::updateTime(); and a reset in World::reset() so a second viewing doesn't inherit the first one's paused state. It's behind a --replay-control flag, and when that's absent the code path is the original statement, byte for byte, so the default build is unchanged.

**An important design decision:** It drives m_time directly rather than changing the main loop's tick rate. In watch-replay mode every kart is a GhostKart whose position is a pure function of the race clock, so moving that one clock moves the ghosts, the on-screen timer and the lap counters together. Rate-limiting the main loop instead would leave the replay UI running on a different clock from the karts, which is the desync worth avoiding.

**The prerequisite:** GhostController::update() currently only ever walks m_current_index forward, so moving the clock backwards leaves the ghost frozen. It needs a matching backward walk (about 8 lines). It's written so the forward path provably never enters the new loop: after the existing forward search m_all_times[m_current_index] <= m_current_time, and the next frame's time is larger still, so ordinary playback is untouched. There's a second latch in GhostKart::update(), where reaching the end of a replay calls setVisible(false) and in watch-replay mode nothing ever shows it again, so scrubbing back from the finish would otherwise leave an invisible kart.

**What I have NOT implemented is the UI.** This is the mechanism, not the buttons. There is no in-game control surface yet. I currently drive it from an external tool over a loopback socket, which I am deliberately not proposing here; that part is specific to my own use case and doesn't belong in the game.

[...]

PR #5859, opened 2026-09-08
Add functionality to change rate of and pause replays in watch replay mode

schrowd:
This feature introduces a new ReplayControl class that, when watch replay mode is enabled,
allows the user to pause and change the rate of replays. It is forward-only for now; rewinding
and potentially seeking would be added in a subsequent PR. That PR would also likely be when
frame-stepping is implemented.

Partially addresses #5197 

## Agreement
```

```


Alayan-stk-2:
All the issues I brought up for forward speed control mode in my previous testing have been fixed, so that's great.

There are some remaining oddities and issues:
- If you pause, and then restart the race, it will still be paused. With replay control info not being displayed (because the race hasn't started yet), this can be slightly confusing. Since restarting the race already resets the replay rate back to 1, it would make sense to also reset pause.
- If you pause, look back doesn't work anymore. It seems that the game wait for the next time step or something similar to switch the camera view.

But we can address those in subsequent commits/PRs.

PR #5871, opened 2026-09-15
Update OpenGL ES version from 2.0 to 3.0

### LINKED ISSUE #5869: Support for OGLES 3.x
Currently, Android builds just use an OGLES 2 context everywhere, so various OpenGL ES features that would work and are available on most of the phones players are actually using end up left behind. This ends up costing picture quality or performance.

There are two broad approaches:
- Drop support entirely for OGLES 2 devices. OGLES 3 was finalized in 2012, and while it took some time for adoption, the old phones that don't support it will get an awful performance anyway on modern STK maps.
- Add feature detection and compatibility paths. Some of it may be required to distinguish between 3.0/3.1/3.2 devices.

In both cases, all the code that currently checks for USE_GLES2 needs to be reviewed to see if the branch should be different for GLES3+ devices. There is also code in lib/irrlicht that also appears to refer to GLES2 that needs to be checked and updated.

This is not something I have much time or experience to contribute to.

The SoC vendors have moved on to promote Vulkan as the new thing, but changing the existing code to move from GLES2 to GLES3+ support requires considerably less work than finishing the Vulkan renderer and would provide immediate benefits to most Android users.

Gustanator:
Fixes #5869 and adds ~3 FPS to the Android build (in the Black Forest benchmark).

**Tests (OGLES2):**

WhatsApp Image 2026-09-15 at 14 53 45
WhatsApp Image 2026-09-15 at 14 53 45 (1)
WhatsApp Image 2026-09-15 at 15 03 51

**Tests (OGLES3):**

WhatsApp Image 2026-09-15 at 14 53 46
WhatsApp Image 2026-09-15 at 14 53 46 (1)
WhatsApp Image 2026-09-15 at 15 05 23

## Agreement
```

```


Gustanator:
I noted some #define's are messed in gles2-ext.h and I didn't edited the files name to OGLES3-related things

I will fix them later

Gustanator:
By the way, here is the PR apk: https://github.com/Gustanator/stk-code/releases/download/ogles3/android-bata-with-ogles3.apk

Alayan-stk-2:
That was a quick turnaround.

However there are some things to clarify first:
- While this addresses lib/irrlicht code ; it doesn't change anything in STK's main code. In android/android_mk we still have  `                     -DUSE_GLES2      \` ; and the main code is peppered with USE_GLES2 checks.
- We still have utils/progress_bar_android.cpp:148:    params.DriverType    = video::EDT_OGLES2;
- We still have IRR_COMPILE_WITH_OGLES2
- Most of the line changes come from removing "OES" from the name of variables. That's perhaps required but could you link some relevant documentation explaining the difference and explain the purpose?
- All the main STK code if USE_GLES2 still make the same assumptions as before. Now, it may be too much work to do all at once, but we should probably (1) comment that GLES2 assumptions are still used for every #ifdef check that hasn't been analyzed for potential benefits of a GLES3 baseline (with USE_GLES2 being renamed to USE_GLES3, it's important to document in the code), (2) try to find at least some cases where the change allows to go down a different execution path (for higher graphical quality or perf).

Gustanator:
I just edited the GLES files without change the #ifdefs for a fast change and test, n
but I will edit later

Gustanator:
> - Most of the line changes come from removing "OES" from the name of variables. That's perhaps required but could you link some relevant documentation explaining the difference and explain the purpose?

In a Raspberry Pi forum: https://forums.raspberrypi.com/viewtopic.php?t=302164

A screenshot that I take of Gemini:

Gustanator:
> I just edited the GLES files without change the #ifdefs for a fast change and test, n
> but I will edit later

It's done ;)

lib/irrlicht/include/EDriverTypes.h
```diff
@@ -54,7 +54,7 @@ namespace video
         
 		//! OpenGL-ES 2.x driver, for embedded and mobile systems
```
Alayan-stk-2, line 55: The comment needs to be updated

lib/irrlicht/source/Irrlicht/CContextEGL.cpp
```diff
@@ -419,7 +419,7 @@ bool ContextManagerEGL::createContext()
 
             std::vector<EGLint> context_attribs;
             context_attribs.push_back(EGL_CONTEXT_CLIENT_VERSION);
-            context_attribs.push_back(2);
+            context_attribs.push_back(3);
```
Alayan-stk-2, line 422: I checked in context within ContextManagerEGL::createContext() ;  and the update doesn't make sense.

The general structure of the code is "if not forced to create a legacy renderer by the setting, create a GLES3 context ; if this fails set legacy device to true and create a GLES2 context".

With an explicit GLES3 floor, the policy should be to create GLES3. On failure, the program should probably error out in some way that communicates the issue.

We could also request GLES3.2 as the default (98% of handeld Android devices support it) and GLES 3.0 as the fallback, but without the legacy device param in play, as it has some other side effects that aren't desirable here iirc.

lib/irrlicht/source/Irrlicht/CContextEGL.cpp
```diff
@@ -488,7 +488,7 @@ bool ContextManagerEGL::createContext()
 
             std::vector<EGLint> context_attribs;
             context_attribs.push_back(EGL_CONTEXT_MAJOR_VERSION);
-            context_attribs.push_back(2);
+            context_attribs.push_back(3);
```
Alayan-stk-2, line 491: This change sits within

    else if (m_creation_params.opengl_api == CEGL_API_OPENGL)

That's regular OpenGL, which should be left untouched by any OGLES change done in this PR.

Alayan-stk-2: This kind of stuff is why I bother to actually review even when it looks good on the surface, and why reviewing isn't always quick.

lib/irrlicht/source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -631,14 +631,14 @@ void CIrrDeviceSDL::tryCreateOpenGLContext(u32 flags)
 		Window = NULL;
 	}
 
-#ifdef _IRR_COMPILE_WITH_OGLES2_
+#ifdef _IRR_COMPILE_WITH_OGLES_
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
```
Alayan-stk-2, line 635: This spot is reached when legacy device is true with OGLES. It tells SDL that we are using OGLES 2.0.

Obviously, we aren't using OGLES 2.0, because we only create 3.0 (or newer perhaps) context, so we shouldn't need all the OGLES cases after the
`legacy:`
goto tag.

However, this code path could still be reached if the parameter itself is turned on. The legacy device parameter should probably be ifdef'd out with OGLES on to guarantee the path is never taken when it shouldn't be.

data/shaders/irrlicht/COGLESRenderer2D.fsh
```diff
@@ -4,20 +4,22 @@
...
 void main(void)
 {
 	vec4 Color = vVertexColor;
 
 	if(uUseTexture)
-		Color *= texture2D(uTextureUnit, vTexCoord);
+		Color *= texture(uTextureUnit, vTexCoord);
 	
-	gl_FragColor = Color;
+	out vec4 Color
```
Alayan-stk-2, line 24: I'm not familiar enough with ES3.0 shader, but I think that the out is supposed to be declared before?

I'm not sure how we would go to test these fragment shaders or what's their precise purpose, but we don't want to silently break them.

Alayan-stk-2 (changes requested):
The majority of the changes are fairly safe renames to remove the misleading GLES2 naming. I had a look over just in case, but that part is fine.

I uncovered a couple of issues, mentioned in my review comments. I have not analyzed the shader code changes to confirm correctness, it will need verification.

In the process of reviewing this, I had a look at code that indicates that CIrrDeviceSDL tries to get a 3.0 context normally, but has a 2.0 fallback. So that confirms that OGLES2 could in theory sometimes happen, but in practice it's already vanishingly rare as very few devices don't support OGLES3 (and even less of those have the required minimum Android version).

There are also potential issues if the legacy device settings is on while using OGLES.

My biggest concern, which hasn't a direct review comment, is with the gles2-ext.h file. That's supposed to be an explicit list of GLES2 extension and their defines (although I'm not too familiar with the specifics), and I think that going in that file to replace various things with GLES3 core equivalent misses the point. If something is already in the main GLES3 headers, why do we need to try to pull the enum values from gles2-ext?

I haven't fully considered the question, but I'm rather confident that's not the right approach.

qwertychouskie:
If the goal is to require GLES 3.0, the old Irrlicht GLES renderer should probably be removed entirely.  No point in carrying around an entire render engine that can never be activated (due to STK's own renderer being used).  If stripping the old renderer is the path chosen, this PR should look very different.

That said, is there any particular benefit to stripping the old renderer?

> but in practice it's already vanishingly rare as very few devices don't support OGLES3 (and even less of those have the required minimum Android version).

On Android, perhaps, but e.g. the Raspberry Pi 1-3 use GLES 2, and it's not hard to imagine a decent number of potential users there (especially since STK is included in distro repos).

Gustanator:
> If stripping the old renderer is the path chosen, this PR should look very different.

I don't know much about the Antarctica Engine (I always thought that the irrlicht forder contains this engine), so I did this basing in afaik the game structure. 

If someone help me to understand the Antarctica engine, I will be very happy :3

Alayan-stk-2:
> If the goal is to require GLES 3.0, the old Irrlicht GLES renderer should probably be removed entirely. No point in carrying around an entire render engine that can never be activated (due to STK's own renderer being used). If stripping the old renderer is the path chosen, this PR should look very different.

Fair point! I hadn't particularly thought about it, but we ought to check in which cases that path is actually taken.

If the COGLES titled files are only used for the legacy GLES2 renderer, then there is no real point in modifying them, they should be stripped after confirming they only ever get used in the legacy path. The flow of instructions leading to them would need to be checked.

However a file like lib/irrlicht/source/Irrlicht/CContextEGL.cpp contains code about both the modern and legacy path so I would assume it is still relevant.

> That said, is there any particular benefit to stripping the old renderer?

Currently, the GLES defines lead to some code paths that assume GLES 2.0 instead of being able to take advantage of core GLES 3.0 features. Now, in theory there are other ways this could be handled, like replacing/supplementing the defines with legacy_device checks, but if we don't have to handle the GLES2 case we get less work and meaningful code simplifications.

Simpler code is also easier to reason about in the future.

> On Android, perhaps, but e.g. the Raspberry Pi 1-3 use GLES 2, and it's not hard to imagine a decent number of potential users there (especially since STK is included in distro repos).

I checked, there are indeed some people reporting playing STK on a Pi 3. I incidentally found an old thread on the Raspberry Pi forums where you participated yourself!

But:
- SBCs aren't really meant to be a primary computing device and most SBC owners will have a more appropriate device to run STK on.
- Performance sucks. We are talking about 30FPS using the very poor-looking legacy renderer. In the one video I found, the speedometer and nitro gauge didn't even work.  It's the very edge of playable, and presumably the 30FPS figure is for the average track, not for the more demanding ones (the video showcase Nessie's Pond which is notoriously . Something like Black Forest would fall further down. And that's on the Pi3, Pi1 and Pi2 don't event enter the argument.
- These factors also mean that while the theoretical number of _devices_ that run that path might be higher than anticipated, the practical number of affected _players_ is extremely low.
- The average future standard STK tracks will be more demanding to run, not less (although we don't want to increase the peak requirements too much, the average requirements will increase significantly).
- Previous STK versions would always remain available. That's one of the beauties of an open project like SuperTuxKart. I would assume that if a user is willing to put up with various things that make for a poor experience (low FPS, crappy resolution and graphics), using the latest and greatest is not their primary concern.

Now in fairness, I suppose that users relying on distro package might find installing an old version more difficult than it is with our official packages, although the Linux arm packages that are available should work on Pi devices, I think.

data/shaders/irrlicht/COGLESRenderer2D.fsh
Gustanator: This is the new syntax for ES 3.0.

The unique error is forgot the ";" :P

Gustanator: fixed

lib/irrlicht/include/EDriverTypes.h
Gustanator: fixed

lib/irrlicht/source/Irrlicht/CContextEGL.cpp
Gustanator: fixed

lib/irrlicht/source/Irrlicht/CIrrDeviceSDL.cpp
Gustanator: Fixed (I guess)

PR #5877, opened 2026-09-17
Speed up SP frustum culling with a p-vertex AABB/plane test

Aminkbi:
## What this changes

`SP::addObject()` and `SP::handleDynamicDrawCall()` decide whether a mesh buffer's bounding box can be culled against the six frustum planes of up to five shadow cascades. That test evaluated all eight corners of the box through a switch-statement `getCorner()`, and stored the per-cascade result in a `std::vector<bool>` that heap-allocated on every call — per mesh buffer and per dynamic draw call, every frame.

**One file changed, `src/graphics/sp/sp_base.cpp`: 17 insertions, 36 deletions.**

1. The corner loop is replaced by the p-vertex (positive extrema vertex) test: one plane evaluation instead of up to eight.
2. `std::vector<bool> discard` becomes `std::array<bool, 5> discard`, removing the allocation. The pass count is pulled into a local so the loop condition is written once.

## Why it is equivalent

For a plane `A*x + B*y + C*z + D = 0`, the maximum of the left-hand side over an axis-aligned box is attained at

```
x = (A >= 0) ? Max.X : Min.X
y = (B >= 0) ? Max.Y : Min.Y
z = (C >= 0) ? Max.Z : Min.Z
```

If that maximum is `< 0`, every point of the box is outside — which is exactly what the old loop decided by requiring all eight corners to be `< 0`. The predicate is unchanged; only the way it is evaluated is.

The flag array is unchanged in meaning too: `{false}` zero-initialises all five entries, and only the first `num_passes` of them are read, matching the previous `resize(handle_shadow ? 5 : 1, false)`.

## Measurements

Method: the registered benchmark replay via `--benchmark`, 5 repetitions per side run in alternating order so drift affects both sides equally, vsync disabled, release build (`-march=native -O3`), 5,400-7,100 frames per run. Host: Ryzen 9 5900HS / Radeon Vega.

| Metric | Before | After | Delta | % |
| :--- | ---: | ---: | ---: | ---: |
| Allocator calls / frame | 11,342.3 | 7,451.5 | **-3,890.8** | **-34.3%** |
| `malloc` calls / frame | 5,620.5 | 3,658.8 | **-1,961.7** | **-34.9%** |
| Instructions / frame | 11.17 M | 10.69 M | **-0.47 M** | **-4.2%** |
| Cycles / frame | 9.06 M | 8.84 M | -0.23 M | -2.5% |
| Branches (per process) | 13.80 G | 13.28 G | -523.8 M | -3.8% |
| `"- culling"` phase mean | 837.9 us | 799.5 us | -38.5 us | -4.6% |
| `"- culling"` phase median | 770.0 us | 727.0 us | -43.0 us | -5.6% |
| Whole-frame mean | 5,405.5 us | 5,444.4 us | +39.0 us | +0.7% |

**Reliable:** the allocation reduction is a mechanical consequence of removing the `std::vector<bool>`, and the instruction reduction does not depend on CPU clock (instructions retired is frequency-independent, unlike cycles).

**Indicative only:** the `- culling` phase moved in the right direction consistently across mean, median and share-of-frame, but the delta sits inside the run-to-run noise band on this host — it ran a `powersave` CPU governor, where drift (about 3%) is larger than the effect being measured. The `cycles` figure should not be read as a 2.5% win for the same reason.

**Unchanged:** whole-frame time, within noise.

## How the equivalence was checked

Both predicates were run against each other differentially before submitting: 1,000,000 random box/plane pairs plus 734,083 probes with a box face placed exactly on a plane and nudged by up to 3 ULP, with 0 disagreements. That check is not part of this commit.

Worth knowing when reproducing it: exact bit-equality is not guaranteed under FMA contraction (GCC's default at `-O3`). One build revision showed 4,523 of 1,734,083 cases disagree, all of them in the deliberately constructed on-plane probes and none among the random ones; `-ffp-contract=off` showed 0. They differ only for a box sitting within a few ULP of a plane, which no real object is.

## Risks

* The p-vertex test does more arithmetic per plane than the old early-out (which stopped at the first non-front corner), but skips eight corner transforms entirely. The instruction counts above are the arbiter, not the operation count.
* Measured on one host and one replay. A quieter setup (fixed `performance` governor) would resolve the culling-phase delta more cleanly.
* `getCorner()` is still used by the bounding-box visualization path, so it is not dead code.

## Reviewer checklist

- [ ] Check the p-vertex selection against `isAABBOutsidePlane()` for a plane with a zero component (`>= 0` picks `MaxEdge`, whose product is zero anyway)
- [ ] Confirm `std::array<bool, 5> discard = {false}` zero-initialises all five entries
- [ ] Confirm the non-shadow path still only consults `discard[0]`


Alayan-stk-2:
The agreement message that fills the default PR description box is required.

Gustanator:
This description seems that was make with AI

PR #5878, opened 2026-09-19
Fix Italian translation

lolgab:
'Set' is translated to 'Partenza' (as it should) in the game

## Agreement
```

```


Alayan-stk-2:
Hello,

Thank you for wanting to help improving SuperTuxKart, however GitHub is not the place to submit translation changes.

The proper procedure is explained at: https://supertuxkart.net/Translating_STK

Directly merging language changes would mess with the proper translation pipeline, and the changes would be overwritten or have to be manually put back after the updated translations are obtained from Transifex.

If you encounter issues with the normal translation procedure or have question related to translating STK, you can also open a thread on the relevant forum: https://forum.supertuxkart.net/forum-42.html

lolgab:
I'm sorry for the noise. I will follow the procedure.
Thank you!

